PR #3, opened 2022-11-09
Add pr template

pducolin:
Add a PR template, copy pasted from dd-go

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,26 +1,11 @@
 What does this PR do?
 ---------------------
 
-Who will it impact?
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,26 @@
+What does this PR do?
```
vboulineau, line 1: This seems a bit overkill for this repo.
The `checklist` does not really match what we do here.
I think What, Why, Impacted Scenarios is enough

pducolin: Done !

vboulineau (approved):

PR #6, opened 2022-11-14
Add SSH Agent support

vboulineau:
What does this PR do?
---------------------

Add SSH Agent support (using `command` 0.6)

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/utils/remote.go
+++ b/common/utils/remote.go
@@ -1,0 +1,29 @@
+package utils
+
+import (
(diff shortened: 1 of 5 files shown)


ogaca-dd (approved):

common/utils/remote.go
```diff
@@ -0,0 +1,29 @@
+package utils
```
pducolin, line 1: **❓ question**
What about calling this file `ssh` ?

PR #13, opened 2022-12-07
add EC2Instance

ogaca-dd:
What does this PR do?
---------------------

This PR adds the ability to create a EC2 instance on different platforms and architectures, and to install the Agent on it.

See https://github.com/DataDog/datadog-agent/pull/14631 for example of usage.

- We have to define which scenarios should be callable from `test-infra-definition` (There are some `string` that should be updated to `pulumi.String`).
- Windows architecture is not yet supported as there is no command runner for Windows.
- For MACOS, there is an issue with the capacity limit.
- `WithAgentConfig` doesn't work on existing instance as the new configuration is first installed and the old one is removed later. `pulumi.DeleteBeforeReplace(true)` doesn't seem to help. 
  
-------------------

Motivation
----------

Additional Notes
----------------


aws/scenarios/vm/ec2instance/params.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func newParams(env aws.Environment, options ...func(*Params) error) (*Params, error) {
+	params := &Params{
+		keyPair: "agent-ci-sandbox",
+		env:     env,
+	}
+
+	// By default use Ubuntu
+	return common.ApplyOption(params, WithOS(os.UbuntuOS, os.AMD64Arch), options)
```
vboulineau, line 30: I don't think defaulting is required at this level. Using this means that scenario has been created entirely through code, not from a CLI-reachable scenario.
If some defaulting is required, it should be at a higher level.

aws/scenarios/vm/os/unix.go
```diff
@@ -0,0 +1,22 @@
...
+func (unix) GetDefaultInstanceType(arch Architecture) string {
+	return getDefaultInstanceType(arch)
+}
+func (unix) GetTenancy() string    { return "default" }
+func (unix) GetConfigPath() string { return "/etc/datadog-agent/datadog.yaml" }
+
+func getDefaultInstanceType(arch Architecture) string {
+	switch arch {
+	case AMD64Arch:
+		return "t3.large"
```
vboulineau, line 16: Variables are available for that already

common/utils/file.go
```diff
@@ -18,7 +18,7 @@ func ReadSecretFile(filePath string) (pulumi.StringPtrOutput, error) {
 }
 
 func WriteStringCommand(content pulumi.StringInput, filePath string) pulumi.StringInput {
-	return pulumi.Sprintf(`cat <<EOF > %s
+	return pulumi.Sprintf(`sh -c 'cat <<EOF > %s
```
vboulineau, line 21: Why do we need this `sh` wrapper? The commands are already run through a shell.

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,74 @@
...
+			Create: pulumi.Sprintf(cmd, params.apiKey),
+		})
+	if err != nil {
+		return err
+	}
+
+	agentConfig := ""
+	if params.agentConfig != "" {
+		fileManager := command.NewFileManager(runner)
+		agentConfig = fmt.Sprintf(params.agentConfig, params.apiKey)
```
pducolin, line 29: **❓ question**
This means we expect `params.agentConfig` to be in the format 

```yaml
...
DD_API_KEY: %s
...
```

Have you considered receiving an agent config already formatted ? Or is there a way to be explicit about expecting agent config to be a template ?

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,74 @@
...
+type agentCommand struct {
+	version version
+	cmd     string
+}
+
+func (a *agentCommand) VisitUnix()    { a.cmd = getInstallFormalString("install_script.sh", a.version) }
+func (a *agentCommand) VisitMacOS()   { a.cmd = getInstallFormalString("install_mac_os.sh", a.version) }
+func (a *agentCommand) VisitWindows() { panic("Not implemented") }
+
+func getInstallFormalString(scriptName string, version version) string {
```
pducolin, line 59: **👏 praise**
Cool !

aws/scenarios/vm/ec2instance/ec2instance.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/DataDog/test-infra-definitions/aws/scenarios/vm/agentinstall"
+	"github.com/DataDog/test-infra-definitions/aws/scenarios/vm/os"
+	"github.com/DataDog/test-infra-definitions/command"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/ec2"
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type Ec2Instance struct {
```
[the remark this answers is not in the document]
pducolin: Or move VM to `interface` as recommended in the first place ?

aws/scenarios/vm/os/os.go
```diff
@@ -0,0 +1,50 @@
...
+
+const (
+	AMD64Arch = Architecture("x86_64")
+	ARM64Arch = Architecture("arm64")
+)
+
+type OSType int
+
+const (
+	WindowsOS OSType = iota
```
pducolin, line 17: TIL `iota` to define enums in golang

pducolin (approved):
Lot of cool improvements ! Let's align on the code structure

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,74 @@
+package agentinstall
```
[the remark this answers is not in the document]
ogaca-dd: Indeed, this code should not be here. The file organisation was not 100% clear to me and so I put everything here. I will start discussions about the file organisation in this repo (AETT-36) and move this code.

aws/scenarios/vm/agentinstall/params.go
```diff
@@ -0,0 +1,68 @@
...
+	return func(p *Params) error {
+		p.version.major = "7"
+		p.version.betaChannel = false
+		return nil
+	}
+}
+
+// WithVersion use a specific version of the Agent. For example: `6.39.0` or `7.41.0~rc.7-1`
+func WithVersion(version string) func(*Params) error {
+	return func(p *Params) error {
```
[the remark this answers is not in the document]
ogaca-dd: This is syntax of the Agent version used by the Agent installer. I can support `7.41.0-rc.7` if you think it is better.

common/utils/file.go
ogaca-dd: When running `sudo` with `cat <<EOF > %s`, sudo applies only to `cat` and not to the redirection. This is for example required to update `datadog.yaml`.

aws/scenarios/vm/agentinstall/install.go
ogaca-dd: If this function accepts a whole Agent config, then the user has to manually adds the API key to this config which is not very convenient. 
Another option would be that `DD_API_KEY` is automatically inserting as a first line of the config. @pducolin what do you think?

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/DataDog/test-infra-definitions/aws"
+	"github.com/DataDog/test-infra-definitions/aws/scenarios/vm/os"
+	"github.com/DataDog/test-infra-definitions/command"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func Install(runner *command.Runner, env aws.Environment, params *Params, os os.OS) error {
+	agentCommand := &agentCommand{version: params.version}
+	os.Visit(agentCommand)
```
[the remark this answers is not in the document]
ogaca-dd: @pducolin : Visitor pattern https://en.wikipedia.org/wiki/Visitor_pattern.

I have spent time to think about using a visitor or not (the other option would be to have a `GetOSType()` function and switch on the `GetOSType()`). 

The reasons, I decided to use a Visitor are:
- I expect to have more OS types (for example AWS linux or Red hat) and more features that depend on the OS type.
- Visitor is a nice way to not have everything in `os` package (It has a nice composability properties).
- If a new `OSType` is added then the user cannot forget to handle the new OS type.

Indeed, Visitor pattern is an advanced pattern and if you don't know it, it can be hard to undertand it. 
I will remove it for now and revisit this decision if there are many `switch GetOStype()` added in the code.

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,74 @@
...
+	cmd := agentCommand.cmd
+	lastCommand, err := runner.Command(
+		env.CommonNamer.ResourceName("agent-install", utils.StrHash(cmd)),
```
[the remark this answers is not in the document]
ogaca-dd: Do you have an idea to update the config file when the api key changed?

aws/scenarios/vm/ec2instance/params.go
```diff
@@ -0,0 +1,83 @@
...
+	userData                   string
+	os                         os.OS
+	env                        aws.Environment
+	arch                       os.Architecture
+	optionalAgentInstallParams *agentinstall.Params
+}
+
+func newParams(env aws.Environment, options ...func(*Params) error) (*Params, error) {
+	params := &Params{
+		keyPair: "agent-ci-sandbox",
```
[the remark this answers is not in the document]
ogaca-dd: What do you mean? Do you suggest using an env var instead or having a way to config the `keyPair`?

aws/scenarios/vm/ec2instance/params.go
ogaca-dd: I think it is useful to be able to create a VM with simply `ec2instance.NewEc2Instance(ctx)` from the `datadog-agent` repository. What do you have in mind with higher level as this API is already high level?

aws/scenarios/vm/agentinstall/install.go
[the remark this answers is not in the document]
ogaca-dd: Code updated.

aws/scenarios/vm/ec2instance/params.go
```diff
@@ -0,0 +1,83 @@
...
+// WithUserData set the userdata for the EC2 instance. User data contains commands that are run at the startup of the instance.
+func WithUserData(userData string) func(*Params) error {
+	return func(p *Params) error {
+		p.userData = userData
+		return nil
+	}
+}
+
+// WithHostAgent installs an Agent on this EC2 instance. By default use with agentinstall.WithLatest().
+func WithHostAgent(apiKey string, options ...func(*agentinstall.Params) error) func(*Params) error {
```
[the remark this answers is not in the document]
ogaca-dd: Maybe, let's wait to have the proper abstraction for VM to see which API is the best. See AETT-37.

aws/scenarios/vm/os/unix.go
ogaca-dd: What do you mean?

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,74 @@
...
+		agentConfig = fmt.Sprintf(params.agentConfig, params.apiKey)
+		remotePath := os.GetConfigPath()
+		lastCommand, err = fileManager.CopyInlineFile("agent-config", agentConfig, remotePath, true, pulumi.DependsOn([]pulumi.Resource{lastCommand}))
+		if err != nil {
+			return err
+		}
+
+	}
+
+	// When the file content has changed, make sure the Agent is restarted.
```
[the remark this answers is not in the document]
ogaca-dd: After properly restarting the Agent, the test https://github.com/DataDog/datadog-agent/pull/14631/commits/e6007b6acdf40a9ff4c0bd5dbe4bcaf67a101810#diff-d4adadf0ea379b565c3595a99ca3bcabf5a14281fa714c5d8f7cd302b9093332R67 is OK. Can you describe an use case where it doesn't work?

aws/scenarios/vm/agentinstall/install.go
pducolin: As a potential user, if I need to pass the agent config, I would probably pre-fill it correctly with the API key, so running the `Sprintf` from the caller side.

I find it counter intuitive to pass both the templated config and the api key, plus with `agentConfig` being required to respect a certain template. I would rather:

- pass only the api key -> a default template would be filled with the api key
- pass the whole config -> caller will take care of filling their template

that would translate in 

```golang
if params.agentConfig == "" && params.apiKey == "" {
  panic("one of agent config or apiKey must be valid")
}

agentConfig := params.agentConfig
if agentConfig == "" {
  // use default agent config template
  agentConfig = fmt.Sprintf(defaultAgentConfigTemplate, params.apiKey)
}
```

Does it make sense ?

aws/scenarios/vm/agentinstall/install.go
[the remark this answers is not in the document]
vboulineau: It goes with the previous command, if you put your inputs in `Triggers` it will run command anytime one of the input is modified.

aws/scenarios/vm/agentinstall/params.go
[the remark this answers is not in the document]
vboulineau: We need to have a way that is consistent across the different ways we install the Agent, I think the semver parsing is well-suited for that.

aws/scenarios/vm/ec2instance/params.go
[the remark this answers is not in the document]
vboulineau: Sorry I wasn't precise enough. We already have `aws/defaultKeyPairName` in the `ConfigMap` for that

aws/scenarios/vm/ec2instance/params.go
vboulineau: What I mean is something like:
```
// Function with params in input
NewEC2Instance(ctx, ...options)

// Shortcuts
NewEC2Ubuntu(ctx)
NewEC2Windows(ctx)
```

aws/scenarios/vm/os/unix.go
vboulineau: Same as for `keyPair` we have a `ConfigMap` parameter for that:
```
	ddinfraDefaultInstanceTypeParamName    = "aws/defaultInstanceType"
	ddinfraDefaultARMInstanceTypeParamName = "aws/defaultARMInstanceType"
```

ogaca-dd: Code updated. Note: I have to export some constants defined in `environment.go` in order to use them in `datadog-agent` repo to set the config map.

aws/scenarios/vm/ec2instance/params.go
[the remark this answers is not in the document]
ogaca-dd: Code updated.

aws/scenarios/vm/agentinstall/install.go
ogaca-dd: > pass only the api key -> a default template would be filled with the api key
You don't need to pass the API key if you use the default configuration so this case never happen.

I will update the code to pass the whole config.

aws/scenarios/vm/agentinstall/install.go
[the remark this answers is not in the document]
ogaca-dd: Api key was removed from the config: https://github.com/DataDog/test-infra-definitions/pull/13#discussion_r1044393843

aws/scenarios/vm/ec2instance/params.go
ogaca-dd: I am keeping it for now as it doesn't hurt.

aws/scenarios/vm/agentinstall/install.go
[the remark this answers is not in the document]
vboulineau: I still see the APIKey right below in `Create`? The proper way would be no hash, and use Triggers on `params.apiKey and cmd`

aws/scenarios/vm/agentinstall/install.go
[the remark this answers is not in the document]
vboulineau: It will work until `agentConfig` needs to be a `pulumi.StringInput`, the clean way is to use `Triggers` here as well.

aws/scenarios/vm/ec2VM/ec2VM.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type Ec2VM struct {
+	runner *command.Runner
+}
+
+// NewEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
+func NewEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (*Ec2VM, error) {
```
vboulineau, line 20: What if I need two VMs? a Name parameter would appropriate and would avoid hardcoding names.

aws/scenarios/vm/ec2VM/ec2VM.go
```diff
@@ -0,0 +1,84 @@
...
+	return &Ec2VM{runner: runner}, nil
+}
+
+func createRunner(ctx *pulumi.Context, env aws.Environment, instance *ec2.Instance, os os.OS) (remote.ConnectionOutput, *command.Runner, error) {
+	connection, err := createConnection(instance, os.GetSSHUser(), env)
+	if err != nil {
+		return remote.ConnectionOutput{}, nil, err
+	}
+
+	runner, err := command.NewRunner(*env.CommonEnvironment, ctx.Stack()+"-conn", connection, func(r *command.Runner) (*remote.Command, error) {
```
vboulineau, line 65: Should use Namer for the name.

common/config/environment.go
```diff
@@ -14,7 +14,7 @@ import (
 
 const (
 	namerNamespace         = "common"
-	ddInfraConfigNamespace = "ddinfra"
+	DDInfraConfigNamespace = "ddinfra"
 	ddAgentConfigNamespace = "ddagent"
```
vboulineau, line 18: Should be aligned and public

common/utils/file.go
vboulineau: I think it would be cleaner to have a `sudo` parameter and use `tee` to write to a file:
```
	return pulumi.Sprintf(`cat - | %s tee %s > /dev/null`, sudo, filePath)
```

aws/scenarios/vm/agentinstall/install.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+
+	"github.com/DataDog/test-infra-definitions/aws"
+	"github.com/DataDog/test-infra-definitions/aws/scenarios/vm/os"
+	"github.com/DataDog/test-infra-definitions/command"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func Install(runner *command.Runner, env aws.Environment, params *Params, os os.OS) error {
```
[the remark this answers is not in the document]
ogaca-dd: This function could take  `config.CommonEnvironment` or even better `common.Namer`. Anyway, as this function is used only in AWS context for now, it is OK. I have a followup PR that change this function to take `common.Namer`.

common/config/environment.go
ogaca-dd: Are you suggesting to make `ddAgentConfigNamespace` also public? It will be probably needed to move `ddAgentConfigNamespace` to public but I prefer to keep private until it is needed.

aws/scenarios/vm/agentinstall/install.go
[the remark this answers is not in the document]
ogaca-dd: Indeed, but when changing `agentConfig` to `pulumi.StringInput` the code will not compile (and so we cannot forgot) and then we can update the code to use `Triggers`.

aws/scenarios/vm/agentinstall/params.go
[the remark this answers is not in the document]
ogaca-dd: For custom branches, Agent versions are not necessary semver for example `7.40.0-devel.git.126.3d581ab-1-x86_64`. Tracked by AETT-40.

aws/scenarios/vm/ec2VM/ec2VM.go
ogaca-dd: Code updated.

common/utils/file.go
ogaca-dd: Code updated.

vboulineau (approved):

aws/environment.go
```diff
@@ -187,3 +189,18 @@ func (e *Environment) EKSBottlerocketNodeGroup() bool {
...
+}
+
+func DefaultARMInstanceTypeKey() string {
+	return getInfraKey(ddinfraDefaultARMInstanceTypeParamName)
+}
+func DefaultKeyPairNameKey() string {
+	return getInfraKey(ddinfraDefaultKeyPairParamName)
+}
+
+func getInfraKey(keyName string) string {
```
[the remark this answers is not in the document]
ogaca-dd: Code updated.

PR #14, opened 2022-12-08
Add support Linux/Windows Azure VM

vboulineau:
What does this PR do?
---------------------

Add support for Linux/Windows Azure VMs

-------------------

None, new scenarios

Motivation
----------

Additional Notes
----------------

The interface is low level at the moment and does not integrate with the `VM` type infrastructure due to on-going changes on this code in #13


azure/compute/vm.go
```diff
@@ -0,0 +1,153 @@
...
+		AdminUsername: pulumi.String(adminUsername),
+		AdminPassword: windowsAdminPassword.Result,
+		CustomData:    userData,
+	}
+
+	if firstLogonCommand != nil {
+		windowsOsProfile.WindowsConfiguration = compute.WindowsConfigurationArgs{
+			AdditionalUnattendContent: compute.AdditionalUnattendContentArray{
+				compute.AdditionalUnattendContentArgs{
+					ComponentName: compute.ComponentNames_Microsoft_Windows_Shell_Setup,
```
pducolin, line 66: **❓ question**
Is this to allow ssh connections?

azure/compute/vm.go
```diff
@@ -0,0 +1,153 @@
...
+	})
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	vm, err := compute.NewVirtualMachine(e.Ctx, e.Namer.ResourceName(name), &compute.VirtualMachineArgs{
+		ResourceGroupName: pulumi.String(e.DefaultResourceGroup()),
+		VmName:            e.Namer.DisplayName(pulumi.String(name)),
+		HardwareProfile: compute.HardwareProfileArgs{
+			VmSize: pulumi.StringPtr(instanceType),
```
pducolin, line 124: **💭 thought**
If `instanceType` is used to set the `VmSize`, should it be called `instanceSize` instead ?

azure/scenarios/vm/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: vm
+runtime: go
+description: Agent VM
```
pducolin, line 3: **🥜 nitpick**
```suggestion
description: Agent Windows VM
```

azure/scenarios/vm/vm/run.go
```diff
@@ -0,0 +1,24 @@
...
+		return err
+	}
+
+	_, publicIP, nw, adminPassword, err := compute.NewWindowsInstance(env, "vm", compute.WindowsLatestURN(), env.DefaultInstanceType(), nil, nil)
+	if err != nil {
+		return err
+	}
+
+	ctx.Export("private-instance-ip", nw.IpConfigurations.Index(pulumi.Int(0)).PrivateIPAddress())
+	ctx.Export("public-instance-ip", publicIP.IpAddress)
```
pducolin, line 21: **❓ question**
Why do we need private and public address ?

azure/compute/vm.go
```diff
@@ -0,0 +1,153 @@
...
+	linuxOsProfile := compute.OSProfileArgs{
+		ComputerName:  pulumi.String(name),
+		AdminUsername: pulumi.String(adminUsername),
+		LinuxConfiguration: compute.LinuxConfigurationArgs{
+			DisablePasswordAuthentication: pulumi.BoolPtr(true),
+			Ssh: compute.SshConfigurationArgs{
+				PublicKeys: compute.SshPublicKeyTypeArray{
+					compute.SshPublicKeyTypeArgs{
+						KeyData: sshPublicKey,
+						Path:    pulumi.StringPtr(fmt.Sprintf("/home/%s/.ssh/authorized_keys", adminUsername)),
```
pducolin, line 35: **👏 praise**
Super nice !

pducolin (approved):

azure/compute/vm.go
vboulineau: So Azure calls it `VmSize` but it really is the `instanceType`, it's really the equivalent of the AWS `instanceType` and not just the # of CPUs.

azure/scenarios/vm/vm/run.go
vboulineau: I have a bastion available in this network as well. Only SSH/RDP is allowed from external. If you need any other kind of interaction you will need to go through the bastion (and private ip)

azure/compute/vm.go
vboulineau: No, it's a way to run script on first-login. I **guess** it could be used to setup SSH, but I haven't verified that yet.

PR #15, opened 2022-12-09
Use stdin to copy inline content

vboulineau:
What does this PR do?
---------------------

Use `stdin` instead of args to copy inline content to avoid size limitation.

-------------------

All

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/utils/file.go
+++ b/common/utils/file.go
@@ -17,8 +17,6 @@ func ReadSecretFile(filePath string) (pulumi.StringPtrOutput, error) {
 	return s, nil
 }
 
-func WriteStringCommand(content pulumi.StringInput, filePath string) pulumi.StringInput {
-	return pulumi.Sprintf(`cat <<EOF > %s
-%s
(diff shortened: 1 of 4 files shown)


common/utils/file.go
```diff
@@ -17,8 +17,6 @@ func ReadSecretFile(filePath string) (pulumi.StringPtrOutput, error) {
 	return s, nil
 }
 
-func WriteStringCommand(content pulumi.StringInput, filePath string) pulumi.StringInput {
-	return pulumi.Sprintf(`cat <<EOF > %s
-%s
-EOF`, filePath, content)
+func WriteStringCommand(filePath string) pulumi.StringInput {
+	return pulumi.Sprintf(`cat - > %s`, filePath)
```
pducolin, line 21: **👏 praise**
TIL ! Wow did not know we could open a file as a stream, super cool.

pducolin (approved):

PR #16, opened 2022-12-09
Implement Kind installation

vboulineau:
What does this PR do?
---------------------

Implement Kind installation

-------------------

None

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/utils/pulumi.go
+++ b/common/utils/pulumi.go
@@ -1,0 +1,9 @@
+package utils
+
+import (
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func PulumiDependsOn(resources ...pulumi.Resource) pulumi.ResourceOption {
+	return pulumi.DependsOn(resources)
+}
--- a/kubernetes/kind.go
+++ b/kubernetes/kind.go
@@ -1,0 +1,81 @@
+package kubernetes
+
+import (
+	_ "embed"
+	"fmt"
+
+	"github.com/DataDog/test-infra-definitions/aws/ec2/ec2"
+	"github.com/DataDog/test-infra-definitions/command"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+const (
+	kindVersion       = "v0.17.0"
+	kindReadinessWait = "60s"
+)
+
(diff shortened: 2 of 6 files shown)


common/utils/pulumi.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+import (
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func PulumiDependsOn(resources ...pulumi.Resource) pulumi.ResourceOption {
```
pducolin, line 7: **❓ question**
What is the advantage of this ? Is this to potentially hide `pulumi` to callers ? So importing `test-infra-definitions/utils` rather than `pulumi` ?

pducolin (approved):

kubernetes/kind.go
```diff
@@ -0,0 +1,81 @@
...
+	kindVersion       = "v0.17.0"
+	kindReadinessWait = "60s"
+)
+
+//go:embed kind-cluster.yaml
+var kindClusterConfig string
+
+// Install Kind on a Linux virtual machine
+// Currently using ec2.VM waiting for correct abstraction
+func NewKindCluster(vm *ec2.VM, clusterName, arch string) (*remote.Command, error) {
```
pducolin, line 25: **💭 thought**
I wonder if we should rather have installer helpers in the VM, so having for example a `KubernetesManager` like we do have a `FileManager`. 

Installing Kind would look like 

```golang
vm.KubernetesManager.CreateKindCluster(clusterName, arch string)
```

This way we could prevent having a public `Runner` in the VM

pducolin: Same for `DockerManager`, it could `vm.DockerManager.InstallAgent(version, image, extraCompose string)`

What do you think ?

pducolin (approved):
Not sure about having the `Runner` public in the VM. But can be addressed separately

common/utils/pulumi.go
vboulineau: Just because it's more practical than typing `[]pulumi.Array{}` every single time.

kubernetes/kind.go
pducolin: What I probably find complex in `Command` is the requirement for a `delete` command. There are many parameters that might be counter intuitives. 

Do we expect test code to use the `Runner` ? 

I see your point, agree having all managers would create lot of coupling on the VM.

vboulineau: It's not a requirement at all, it can be `nil` if it makes sense.

PR #17, opened 2022-12-12
Set storage of root EBS volume

usamasaqib:
What does this PR do?
---------------------
This PR allows the user to specify the storage size of an ec2 instance.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/aws/ec2/ec2/ebs.go
+++ b/aws/ec2/ec2/ebs.go
@@ -1,0 +1,88 @@
+package ec2
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/ec2"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type BlockDeviceArgs struct {
+	// Whether the volume should be destroyed on instance termination. Defaults to `true`.
+	DeleteOnTermination pulumi.BoolPtrInput `mapstructure:"delete_on_termination"`
+	// Name of the device to mount.
+	DeviceName pulumi.StringInput `mapstructure:"device_name"`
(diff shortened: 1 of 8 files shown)


pducolin (approved):
Looks good ! To be double checked if `SearchAmi` change impacts other PRs currently open

aws/ec2/ec2/ebs.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+	"reflect"
+	"strconv"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/ec2"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type BlockDeviceArgs struct {
```
vboulineau, line 13: Why going through all that for a dozen of fields?
Why not just having a simple function like:
```
func convertEbsToBlockDeviceArgs(ebs map[string]string) (*ec2.InstanceEbsBlockDeviceArgs, error) {
  args.Foo = ebs["bar"]
}
```

aws/ec2/ec2/vm.go
```diff
@@ -89,6 +83,25 @@ func NewEC2Instance(e aws.Environment, name, ami, arch, instanceType, keyPair, u
...
+	if ami == "" {
+		ami, err = LatestUbuntuAMI(e, arch)
+		if err != nil {
+			return nil, err
+		}
+		device, err := LatestUbuntuAMIRootDevice(e, arch)
+		if err != nil {
+			return nil, err
+		}
+		rootBlockDevice, err := createNewRootDeviceMapping(device, e.DefaultInstanceStorageSize())
```
vboulineau, line 97: What's the goal of setting this while the AMI is already `Ubuntu`?
Are you trying to simply change the root disk size? In which case you can simply do:
```
        RootBlockDevice: ec2.InstanceRootBlockDeviceArgs{
            VolumeSize: pulumi.IntPtr(200),
        },
```

vboulineau (changes requested):

aws/ec2/ec2/vm.go
usamasaqib: That is exactly what I needed! Thanks

PR #19, opened 2022-12-13
Add support for simple AKS Cluster

vboulineau:
What does this PR do?
---------------------

Add Azure AKS scenario

-------------------

None

Motivation
----------

Additional Notes
----------------


azure/aks/cluster.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	creds := containerservice.ListManagedClusterUserCredentialsOutput(e.Ctx,
+		containerservice.ListManagedClusterUserCredentialsOutputArgs{
+			ResourceGroupName: pulumi.String(e.DefaultResourceGroup()),
+			ResourceName:      cluster.Name,
+		}, pulumi.Provider(e.Provider),
+	)
+
+	kubeconfig := creds.Kubeconfigs().Index(pulumi.Int(0)).Value().
```
ogaca-dd, line 67: Can `creds.Kubeconfigs()` be empty? Is there an explicit error message?

ogaca-dd (approved):

azure/aks/cluster.go
vboulineau: It should not be but `ListManagedClusterUserCredentialsOutput` does not return any error. I believe there's an error check inside the `ApplyT` that should make it fail.

Normally it should not be empty.

PR #21, opened 2022-12-15
MicroVMs scenario

usamasaqib:
What does this PR do?
---------------------
This PR includes a scenario for provisioning libvirt based micro-vms on an ec2 metal instance.

-------------------

Motivation
----------
This scenario will be used to provision the environment for the kernel-version matrix testing framework being built by the ebpf-platform team.

Additional Notes
----------------
Run this scenario with the following command:
`aws-vault exec sandbox-account-admin -- pulumi up -c ddinfra:aws/defaultKeyPairName=<SSH-KEY-NAME> -c ddinfra:aws/defaultPrivateKeyPath=<SSH-KEY-PATH> -c ddinfra:aws/defaultPrivateKeyPassword=<SSH-KEY-PASS> -c ddinfra:env=aws/sandbox -c ddinfra:aws/defaultInstanceType=z1d.metal -c ddinfra:aws/defaultInstanceStorageSize=100 -c ddMicroVM:microVMConfigFile="<PATH-TO-VMCONFIG-FILE>" -C ./aws/scenarios/microVMs -s <STACK-NAME>`


command/runner.go
```diff
@@ -84,3 +85,29 @@ func (r *Runner) Command(name string, args *CommandArgs, opts ...pulumi.Resource
 
 	return remote.NewCommand(r.e.Ctx, r.namer.ResourceName("cmd", name), args.toRemoteCommandArgs(r.connection), opts...)
 }
+
+type LocalRunner struct {
```
usamasaqib, line 89: I would like to change this to implement a `Runner` interface, used by both the remote runner and the local runner. This would allow us to use the same code path in this PR for provisioning local development environments as well.

Would this be suitable?

aws/scenarios/microVMs/microVMs/libvirt.go
```diff
@@ -0,0 +1,185 @@
...
+	baseVolumeId := generatePoolPath(vmset.Name) + basefsName
+
+	provider, err := libvirt.NewProvider(ctx, "provider", &libvirt.ProviderArgs{
+		Uri: libvirtUri,
+	}, pulumi.DependsOn(waitForList))
+	if err != nil {
+		return err
+	}
+
+	domainXls, err := os.ReadFile("resources/domain.xls")
```
pducolin, line 68: **💬 suggestion**
You can embed directive to embed this file in the binary with `//go:embed` https://pkg.go.dev/embed
It raises compile time errors if the file is missing, no need to read the file content. Wrote a tiny comment about it at https://github.com/pducolin/advent-of-code#embed

aws/scenarios/microVMs/vmconfig/load.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+)
+
+func LoadFile(filename string) (*Config, error) {
```
pducolin, line 11: **💭 thought**
Since this file is supposed to be a configuration file, would it make sense to call this function `LoadConfigFile` or `ParseFromFile` ?

Note you could use `embed` here too and `ParseFromString`

command/runner.go
pducolin: What is preventing from using `Runner` ? Would you need to add a `isLocal` or `flavour` property ?

pducolin: You could also use a common interface and implement common functions in `LocalRunner` and `RemoteRunner`

pducolin (approved):
Great work ! 

Makes sense to have `microVMs` in `scenarios`, but we might change it as we are discussing filesystem organisation within the squad cc @vboulineau

command/runner.go
usamasaqib: From what I understand `Runner` runs commands on the remote instance provisioned by pulumi.

In this case I am trying to run commands on the machine where pulumi itself is running. I also, need these commands to be tracked by pulumi to gracefully handle update scenarios.

The main issue I am running into with a common interface is that the `Command` object for the local runner and the remote runner are both different. I am not sure how I can implement a scenario given this.

aws/scenarios/microVMs/microVMs/libvirt_fs.go
```diff
@@ -0,0 +1,246 @@
...
+				imageSource:   img.ImageSourceURI,
+				volumeKey:     volKey,
+				volumeXML:     fmt.Sprintf(volumeXMLTemplate, basefsName, volKey, volKey),
+				volumeXMLPath: fmt.Sprintf("/tmp/volume-%s.xml", img.ImageName),
+			},
+		},
+	}
+}
+
+func downloadRootfs(fs *libvirtFilesystem, runner *command.Runner, depends []pulumi.Resource) ([]pulumi.Resource, error) {
```
vboulineau, line 69: Considering the number of commands in this file, would be it clearer to simply copy (once) and execute a bash script?

aws/scenarios/microVMs/microVMs/libvirt_fs.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	extractRootfsDone, err := extractRootfs(fs, runner, downloadRootfsDone)
+	if err != nil {
+		return []pulumi.Resource{}, err
+	}
+
+	poolXmlWrittenArgs := command.CommandArgs{
+		Create: pulumi.String(
+			fmt.Sprintf("echo \"%s\" > /tmp/pool.xml", fs.poolXML),
```
vboulineau, line 205: We have a dedicated `FileManager` for that

command/runner.go
vboulineau: Implementation looks legit to me. We could introduce an interface to wrap both, but IMHO may not be worth it ATM.

vboulineau (commented):
LGTM overall. I think it will benefit from the `VM` interface we're introducing. Maybe some simplification can be achieved by using remote scripts instead of many commands.

aws/scenarios/microVMs/microVMs/libvirt_fs.go
usamasaqib: We plan to use this scenario for local testing as well. Using a single bash script could be problematic if the user wants to add and remove different VMs from the configuration, since we will have to rebuild the volumes for each every time the configuration file is edited.
On the other hand by having pulumi track each command, only the relevant commands are rerun on each modification of the microVM configuration file.

vboulineau: Maybe you can generate the script on-the-fly from a template or something.
I think it makes sense if commands are slowish (few seconds) or not idempotent, otherwise I think it's okay to just call the script all the time

aws/scenarios/microVMs/microVMs/libvirt.go
```diff
@@ -0,0 +1,288 @@
...
+	ipv4 := ip.To4()
+	ipv4 = ipv4.Mask(subnetGroupMask)
+	ipv4[2]++
+
+	return ipv4
+}
+
+func generateNewUnicastMac() (string, error) {
+	buf := make([]byte, 6)
+	_, err := rand.Read(buf)
```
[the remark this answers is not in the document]
usamasaqib: Does the pulumi `random` plugin generate the same values for all executions of the stack?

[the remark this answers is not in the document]
vboulineau: Yes, basically you can use `random` and get the same random value regardless of the operations you perform (many `pulumi up` for instance)

aws/ec2/ec2/vm.go
```diff
@@ -54,8 +54,8 @@ func NewVM(ctx *pulumi.Context) (*VM, error) {
 	return vm, nil
 }
 
-func NewDefaultEC2Instance(e aws.Environment, name, instanceType string) (*ec2.Instance, remote.ConnectionOutput, error) {
-	awsInstance, err := NewEC2Instance(e, name, "", AMD64Arch, instanceType, e.DefaultKeyPairName(), "", "default")
+func NewDefaultEC2Instance(e aws.Environment, name, arch, instanceType string) (*ec2.Instance, remote.ConnectionOutput, error) {
```
usamasaqib, line 57: @pducolin @vboulineau 
Made a small change to allow users to specify arch of EC2 instance. Any issues with this?

pducolin: Why not using `NewEC2Instance` to define the architecture instead of `Default` Wold keep `default` with as low params as possible

usamasaqib: Makes sense!

aws/scenarios/microVMs/microVMs/libvirt_fs.go
usamasaqib: We are planning to generate a custom ami for our purposes, and much of this logic will be moved there.
I can refactor this code in a followup PR once we have the relevant parts ready.

PR #24, opened 2022-12-20
Add Azure VM API

ogaca-dd:
What does this PR do?
---------------------
This PR adds an API to create Azure VMs.

Note: Windows VM doesn't work yet.


aws/scenarios/vm/os/macos.go
```diff
@@ -3,46 +3,46 @@ package os
 import (
 	"github.com/DataDog/test-infra-definitions/aws"
 	"github.com/DataDog/test-infra-definitions/aws/ec2/ec2"
+	"github.com/DataDog/test-infra-definitions/common/os"
 )
 
 type macOS struct {
+	*os.MacOS
```
pducolin, line 10: TIL: field names are not mandatory in struct, by default go uses field type as a name

aws/scenarios/vm/os/os.go
```diff
@@ -2,42 +2,24 @@ package os
...
-
-const (
-	WindowsOS OSType = iota
-	UbuntuOS         = iota
-	MacOS            = iota
+	"github.com/DataDog/test-infra-definitions/common/os"
 )
 
 type OS interface {
+	os.OS
```
pducolin, line 9: **❓ question**
Is this a field part of an interface ? Didn't know we could have fields in interfaces

aws/scenarios/vm/os/os.go
```diff
@@ -2,42 +2,24 @@ package os
...
-	ARM64Arch = Architecture("arm64")
-)
-
-type OSType int
-
-const (
-	WindowsOS OSType = iota
-	UbuntuOS         = iota
-	MacOS            = iota
+	"github.com/DataDog/test-infra-definitions/common/os"
```
pducolin, line 5: **💭 thought**
Could help renaming this to `commonos` to avoid confusion with the current package name

common/os/macos.go
```diff
@@ -0,0 +1,15 @@
...
+
+func NewMacOS() *MacOS {
+	return &MacOS{}
+}
+
+func (*MacOS) GetServiceManager() *serviceManager {
+	return &serviceManager{restartCmd: []string{"launchctl stop com.datadoghq.agent", "launchctl start com.datadoghq.agent"}}
+}
+
+func (*MacOS) GetConfigPath() string { return "~/.datadog-agent/datadog.yaml" }
```
pducolin, line 13: **🥜 nitpick**
Wonder if this should be `GetAgentConfigPath`. Should we also have a `GetAgentConfdPath` for integration config paths ?

common/common.go
```diff
@@ -0,0 +1,10 @@
+package common
+
+func ApplyOption[T any](instance *T, options []func(*T) error) (*T, error) {
```
pducolin, line 3: **👏 praise**
Very elegant !

common/agentinstall/params.go
```diff
@@ -17,7 +17,8 @@ func NewParams(apiKey string, options ...func(*Params) error) (*Params, error) {
 	params := &Params{
 		apiKey: apiKey,
 	}
-	return common.ApplyOption(params, WithLatest(), options)
+	options = append([]func(*Params) error{WithLatest()}, options...)
```
pducolin, line 20: **❓ question**
Is this different from 
```
options = append(options, []func(*Params) error{WithLatest()})
```
?

Is it because you want first to apply `WithLatest` ?

common/vm/params.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+// SetInstanceType set the instance type
+func (p *Params[OS]) SetInstanceType(instanceType string) error {
+	p.InstanceType = instanceType
+	return nil
+}
+
+// SetUserData set the userdata for the EC2 instance. User data contains commands that are run at the startup of the instance.
```
pducolin: **🥜 nitpick**
Should we remove the `EC2` mention here ?

aws/scenarios/vm/os/os.go
```diff
@@ -12,15 +12,6 @@ type OS interface {
...
-	case os.WindowsOS:
-		return newWindows(env)
-	case os.UbuntuOS:
-		return newUbuntu(env)
-	case os.MacosOS:
-		return newMacOS(env)
-	default:
-		panic("OS not supported")
-	}
+func GetOSes(env aws.Environment) []OS {
```
pducolin, line 14: **🥜 nitpick**
Could this be named `GetSupportedOSes` ?

common/vm/vm.go
```diff
@@ -0,0 +1,69 @@
...
+	instanceIP pulumi.StringInput,
+	os os.OS,
+	optionalAgentInstallParams *agentinstall.Params,
+) (*command.Runner, error) {
+	connection, runner, err := createRunner(ctx, env, instanceIP, os.GetSSHUser())
+	if err != nil {
+		return nil, err
+	}
+
+	if optionalAgentInstallParams != nil {
```
pducolin, line 26: **❓ question**
Does this mean that if `optionalAgentInstallParams` is `nil` we don't install the agent ?

azure/compute/vm/azurevm.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	params, err := newParams(env, options...)
+	if err != nil {
+		return nil, err
+	}
+
+	if params.common.OS.GetOSType() != os.UbuntuOS {
```
pducolin, line 31: **💭 thought**
Should we instead comment out `windows` in `azure/compute/os/os.go` in ` GetOSes` while we don' support it ?

pducolin (approved):
Super cool and elegant, learning a lot about generics in Golang. 

Couple of questions and some suggestions, no blocker !

aws/scenarios/vm/os/macos.go
ogaca-dd: `*os.MacOS` is an [embedded struct](https://eli.thegreenplace.net/2020/embedding-in-go-part-1-structs-in-structs/). The usage of an embedded struct and a regular field is not the same. 
Embedded struct should be used only when all the methods from the embedded struct (`*os.MacOS`) should be accessible from parent type (`macOS`).

aws/scenarios/vm/os/os.go
ogaca-dd: `os.OS` is an embedded interface. To implement `OS`, a type must implement the method from `OS` and the methods from `os.OS`.

common/agentinstall/params.go
ogaca-dd: The order matters. `WithLatest` should be applied first and maybe override by an option.

common/os/macos.go
ogaca-dd: I didn't change the name of `GetConfigPath` in this PR but indeed `GetAgentConfigPath` is more relevant.

aws/scenarios/vm/os/os.go
ogaca-dd: I didn't change the name of this function in this PR but `GetSupportedOSes` is more relevant.

common/vm/vm.go
ogaca-dd: Yes

azure/compute/vm/azurevm.go
ogaca-dd: I am working on fixing the issue (AETT-43).

aws/scenarios/vm/os/os.go
ogaca-dd: Code updated.

aws/scenarios/vm/os/os.go
ogaca-dd: Code updated.

pducolin (approved):
Thank you for addressing the comments !

aws/scenarios/vm/ec2VM/ec2VM.go
```diff
@@ -18,75 +14,42 @@ type Ec2VM struct {
...
-	}
+	os := params.common.OS
 	instance, err := awsEc2.NewEC2Instance(
-		e,
-		resourceName,
-		params.ami,
-		params.os.GetAMIArch(params.arch),
-		params.instanceType,
+		env,
+		env.CommonNamer.ResourceName(params.common.GetInstanceNameOrDefault("ec2-instance")),
```
vboulineau, line 30: Parameter should be mandatory, otherwise we cannot have 2 VMs.

common/config/environment.go
```diff
@@ -136,3 +136,11 @@ func (e *CommonEnvironment) GetObjectWithDefault(config *sdkconfig.Config, param
 
 	return defaultValue
 }
+
+type Environment interface {
```
vboulineau, line 140: I'm not sure this belongs here.
I think what you want is a `Environment` interface with the methods you're interested in, probably in `vm`.

common/vm/vm.go
```diff
@@ -0,0 +1,69 @@
...
+) (remote.ConnectionOutput, *command.Runner, error) {
+	connection, err := createConnection(instanceIP, sshUser, env)
+	if err != nil {
+		return remote.ConnectionOutput{}, nil, err
+	}
+
+	commonEnv := env.GetCommonEnvironment()
+	runner, err := command.NewRunner(
+		*commonEnv,
+		commonEnv.CommonNamer.ResourceName(ctx.Stack()+"-conn"),
```
vboulineau, line 48: Namer should be used

common/vm/vm.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/DataDog/test-infra-definitions/common/os"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func InitVM(
+	ctx *pulumi.Context,
```
vboulineau, line 15: Redundant with Environment

common/vm/vm.go
ogaca-dd: > The parameter is unclear

Can you elaborate what is not clear? `optionalAgentInstallParams` is an optional field and when it is not null contains the information for installing the Agent. This code seems straightforward to understand. Did I miss something?

> we also have a ConfigMap parameter for that.

I have updated the code in https://github.com/DataDog/test-infra-definitions/pull/25 (https://github.com/DataDog/test-infra-definitions/pull/25/commits/1d1628008215700cac50ebac0bd0653261be39bc) to limit the code conflict.
Note: This code was introduced in a previous PR.

aws/scenarios/vm/ec2VM/ec2VM.go
ogaca-dd: I assume the common case is to create a single VM and creating more than one VM will not be common. In this case, I think it is better to provide a default name and raises an explicit error message when creating several unnamed VM with a clear explanation about what to do. 
Unfortunately, this has other implication as for example the `AWSEnvironment` has to be uniq  (https://github.com/DataDog/test-infra-definitions/pull/25/commits/1d1628008215700cac50ebac0bd0653261be39bc). I will handle this change in another PR.
Tracked by AETT-55.

common/config/environment.go
ogaca-dd: I put this interface here because of dependencies.

`Environment` is used by both `common/vm` and `common/os` and the first depends on the later. `Environment` cannot be in `common/vm`.  `Environment` also depends on `CommonEnvironment` and so cannot be in the root package `common`.
`Environment` can be in `config/config` or in a dedicated package.
If you prefer, I can rename `common/config/environment.go` to `common/config/commonEnvironment.go` and move the code of `Environment` inside `common/config/environment.go`.

common/vm/vm.go
ogaca-dd: Code updated.

common/vm/vm.go
ogaca-dd: Namer is used. Are you suggesting using something like `commonEnv.CommonNamer.ResourceName("connection")?

vboulineau: You don't need to prefix with `ctx.Stack()` when using namer indeed.

common/vm/vm.go
vboulineau: > Does this mean that if optionalAgentInstallParams is nil we don't install the agent ?

The fact that giving no install params deactivate Agent installation is not clear. The name of parameter suggests that indeed it's optional and you **can** give install params but not that not giving any **will** deactivate Agent installation.

I believe the choice to install/uninstall should still be governed with the ConfigMap parameter.

common/config/environment.go
vboulineau: Given the links between `vm` and `os`, it may mean that it should be a single package.
Maybe another option is to have something like `common/vm/environment.go` and then `common/vm/builder/*` with existing code.

ogaca-dd: As I have several local PRs that are based on this file hierarchy, I will do the change in a later PR. Tracked by `AETT-57`.

common/vm/vm.go
ogaca-dd: Code updated

PR #25, opened 2022-12-27
Remove API key from agentinstall.Params

ogaca-dd:
What does this PR do?
---------------------

This PR provides small improvements:
 - Don't use SemVer for the Agent version as custom build versions are not SemVer versions.
- Remove API key from `agentinstall.Params` and use config map instead.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/agentinstall/install.go
+++ b/common/agentinstall/install.go
@@ -2,20 +2,22 @@ package agentinstall
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/DataDog/test-infra-definitions/command"
-	"github.com/DataDog/test-infra-definitions/common/namer"
+	"github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/os"
 	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-func Install(runner *command.Runner, commonNamer namer.Namer, params *Params, os os.OS) error {
+func Install(runner *command.Runner, env *config.CommonEnvironment, params *Params, os os.OS) error {
 	cmd := getInstallFormatString(os.GetOSType(), params.version)
+	commonNamer := env.CommonNamer
 	lastCommand, err := runner.Command(
 		commonNamer.ResourceName("agent-install", utils.StrHash(cmd)),
 		&command.CommandArgs{
-			Create: pulumi.Sprintf(cmd, params.apiKey),
+			Create: pulumi.Sprintf(cmd, env.AgentAPIKey()),
 		})
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


common/agentinstall/params.go
```diff
@@ -46,16 +43,12 @@ func WithVersion(version string) func(*Params) error {
...
-			// Update from `7.41.0-rc.7` to `7.41.0~rc.7-1
-			p.version.minor = strings.ReplaceAll(p.version.minor, "-", "~") + "-1"
-		}
+		p.version.betaChannel = strings.Contains(version, "~")
 		return nil
 	}
 }
 
-// WithAgentConfig sets the configuration of the Agent.
+// WithAgentConfig sets the configuration of the Agent. `{{API_KEY}}` can be used as a placeholder for the API key.
```
pducolin, line 51: **❓ question**
Have we considered using `DD_API_KEY` environment variable ? Though it could be available only for Docker install

common/agentinstall/install.go
```diff
@@ -24,7 +26,11 @@ func Install(runner *command.Runner, commonNamer namer.Namer, params *Params, os
 	if params.agentConfig != "" {
 		fileManager := command.NewFileManager(runner)
 		remotePath := os.GetConfigPath()
-		lastCommand, err = fileManager.CopyInlineFile("agent-config", pulumi.String(params.agentConfig), remotePath, true, pulumi.DependsOn([]pulumi.Resource{lastCommand}))
+		agentConfig := env.AgentAPIKey().ApplyT(func(apiKey string) pulumi.String {
```
pducolin, line 29: **❓ question**
Does `ApplyT` block and wait for the resolution of the string (yes I could read the docs)

common/agentinstall/params.go
```diff
@@ -46,16 +43,12 @@ func WithVersion(version string) func(*Params) error {
 		}
 
 		p.version.minor = strings.TrimPrefix(version, prefix)
-		p.version.betaChannel = strings.Contains(version, "-")
-		if p.version.betaChannel {
-			// Update from `7.41.0-rc.7` to `7.41.0~rc.7-1
-			p.version.minor = strings.ReplaceAll(p.version.minor, "-", "~") + "-1"
-		}
+		p.version.betaChannel = strings.Contains(version, "~")
```
pducolin, line 46: 👍

common/agentinstall/params.go
```diff
@@ -30,7 +27,7 @@ func WithLatest() func(*Params) error {
 	}
 }
 
-// WithVersion use a specific version of the Agent. For example: `6.39.0` or `7.41.0-rc.7`
+// WithVersion use a specific version of the Agent. For example: `6.39.0` or `7.41.0~rc.7-1
```
pducolin, line 30: **❓ question**
We still support rc candidate versions right ? Not used to the `~` format, I only used `-` rc candidates

pducolin (approved):
Nice !

common/agentinstall/install.go
pducolin: Read the doc, it's more like a `.then`, returning a new ~Promise~ Output

common/agentinstall/params.go
ogaca-dd: Nothing prevent from using `DD_API_KEY` as `{{API_KEY}}` is a placeholder

common/agentinstall/params.go
ogaca-dd: A RC candidate version has a format similar to `7.41.0~rc.7-1`. Note: the Agent image docker version uses `-` format.

pducolin:
All good !

PR #26, opened 2022-12-28
Add function to copy folder in FileManager

ogaca-dd:
What does this PR do?
---------------------

Add function to copy folder in FileManager

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/command/filemanager.go
+++ b/command/filemanager.go
@@ -147,7 +147,7 @@ func (fm *FileManager) CopyFSFolder(
 	return fileResources, nil
 }
 
-// When copying foo/bar to /tmp to result folder is /tmp/bar
+// When copying foo/bar to /tmp the result folder is /tmp/bar
 // This function remove the root prefix from the path (`foo` in this case)
 func getDestinationPath(folder string, rootFolder string) (string, error) {
 	baseFolder := filepath.Base(rootFolder)
--- a/command/filemanager.go
+++ b/command/filemanager.go
@@ -67,6 +67,9 @@ func (fm *FileManager) CopyInlineFile(name string, fileContent pulumi.StringInpu
 // then the full path of the folder will be ~/foo/bar“.
 // This function returns the resources that can be used with `pulumi.DependsOn`.
 func (fm *FileManager) CopyRelativeFolder(relativeFolder string, remoteFolder string, opts ...pulumi.ResourceOption) ([]pulumi.Resource, error) {
+	// `./` cannot be used with os.DirFS
+	relativeFolder = strings.TrimPrefix(relativeFolder, "."+string(filepath.Separator))
+
 	fullPath, rootFolder, err := getFullPath(relativeFolder, 2)
 	if err != nil {
 		return nil, err
@@ -103,7 +106,7 @@ func (fm *FileManager) CopyFSFolder(
 
 	files, folders, err := readFilesAndFolders(folderFS, folderPath)
 	if err != nil {
-		return nil, fmt.Errorf("cannot read files and folders for %v", folderPath)
+		return nil, fmt.Errorf("cannot read files and folders for %v. Error: %v", folderPath, err)
 	}
 
 	var folderResources []pulumi.Resource
@@ -128,15 +131,9 @@ func (fm *FileManager) CopyFSFolder(
 			return nil, err
 		}
 
-		fileContent, err := fs.ReadFile(folderFS, file)
-		if err != nil {
-			return nil, err
-		}
-		fileCommand, err := fm.CopyInlineFile(
-			resourceName+"-"+file,
-			pulumi.String(fileContent),
+		fileCommand, err := fm.CopyFile(
+			file,
 			path.Join(remoteFolder, destFile),
-			useSudo,
 			pulumi.DependsOn(folderResources))
 		if err != nil {
 			return nil, err
--- a/command/filemanager.go
+++ b/command/filemanager.go
@@ -1,7 +1,14 @@
 package command
 
 import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
 	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
 
 	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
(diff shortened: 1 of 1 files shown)


command/filemanager.go
```diff
@@ -53,3 +60,133 @@ func (fm *FileManager) CopyInlineFile(name string, fileContent pulumi.StringInpu
...
+	rootWithoutBase := rootFolder[:len(rootFolder)-len(baseFolder)]
+
+	if !strings.HasPrefix(folder, rootWithoutBase) {
+		return "", fmt.Errorf("%v doesn't have the prefix %v", folder, rootWithoutBase)
+	}
+
+	return folder[len(rootWithoutBase):], nil
+}
+
+func getFullPath(relativeFolder string, skip int) (string, string, error) {
```
pducolin, line 163: **❓ question**
What does `skip` do ?

command/filemanager.go
```diff
@@ -53,3 +60,133 @@ func (fm *FileManager) CopyInlineFile(name string, fileContent pulumi.StringInpu
...
+		if err != nil {
+			return nil, err
+		}
+		fileResources = append(fileResources, fileCommand)
+	}
+
+	return fileResources, nil
+}
+
+// When copying foo/bar to /tmp to result folder is /tmp/bar
```
pducolin, line 150: **🥜 nitpick**
```suggestion
// When copying foo/bar to /tmp the result folder is /tmp/bar
```

command/filemanager.go
```diff
@@ -53,3 +60,133 @@ func (fm *FileManager) CopyInlineFile(name string, fileContent pulumi.StringInpu
 			Sudo:   useSudo,
 		}, opts...)
 }
+
+// CopyRelativeFolder copies recursively a relative folder to a remote folder.
+// The path of the folder is relative to the caller of this function.
+// For example, if this function is called from ~/foo/test.go with remoteFolder="bar"
+// then the full path of the folder will be ~/foo/bar“.
```
pducolin, line 67: **❓ question**
What is the result if we call from `~/foo/test.go` with relative folder `bar/toto` and remoteFolder `tutu` ?
Isn't copy always happening in `tmp` remote folder ?

pducolin (approved):
Couple of questions !

command/filemanager.go
ogaca-dd: It means you are trying to copy files from `~/foo/bar/toto` to `tutu` which means files in `~/foo/bar/toto` will be copied into `tutu/toto` (I did my best to make this function behaves the same as `cp -r`).

command/filemanager.go
ogaca-dd: See https://pkg.go.dev/runtime#Caller.

command/filemanager.go
```diff
@@ -53,3 +60,133 @@ func (fm *FileManager) CopyInlineFile(name string, fileContent pulumi.StringInpu
...
+
+	if !strings.HasPrefix(folder, rootWithoutBase) {
+		return "", fmt.Errorf("%v doesn't have the prefix %v", folder, rootWithoutBase)
+	}
+
+	return folder[len(rootWithoutBase):], nil
+}
+
+func getFullPath(relativeFolder string, skip int) (string, string, error) {
+	_, file, _, ok := runtime.Caller(skip)
```
vboulineau, line 164: If I'm not mistaken `runtime.Caller` returns the build path, which will be different depending on many things (if it's built in `test-infra` or `datadog-agent`, another machine, CI, etc.).

By design, the equivalent of `__file__` doest not exist in `Go`, so I'm not sure this code is entirely reliable. That's the reason why the existing API takes full path or embed strings.

command/filemanager.go
```diff
@@ -53,3 +60,133 @@ func (fm *FileManager) CopyInlineFile(name string, fileContent pulumi.StringInpu
...
+
+		if err != nil {
+			return nil, err
+		}
+
+		fileContent, err := fs.ReadFile(folderFS, file)
+		if err != nil {
+			return nil, err
+		}
+		fileCommand, err := fm.CopyInlineFile(
```
vboulineau, line 135: It's probably better to use the file-system copy file instead of inline (stdin) as it should be faster.

vboulineau (changes requested):

command/filemanager.go
ogaca-dd: Code updated.

command/filemanager.go
ogaca-dd: `runtime.Caller` returns the **full path**. It makes no difference, if it's built in `test-infra` or `datadog-agent` (it works the same in both cases). If the binary is moved to another machine then copying the data makes a lot of sense.

vboulineau: It makes a lot of sense in case of remote execution, artifact copy in CI, docker mounts, etc. IMHO if the caller wants to do that, it can call `runtime.Caller()` itself. It would make this API reliable in any case and defer the decision to the one that really knows where the file is.

command/filemanager.go
vboulineau: Push probably missing as code did not change

command/filemanager.go
ogaca-dd: I understand the use cases you described but I think they doesn't apply for what we are doing:
- For running E2E tests, `go test` requires the source code
- For creating an environment, `pulumi up` requires also the source code
The only case where it can theoretically happen is when creating an environment from `datadog-agent` repository and the creation of the environment is done using a binary that is moved to another host. It is unlikely to happen and the user can still use embedded file in this very particular case.
I am a little hesitant to remove a feature that make life easier for the users because there is a theoretical case where it doesn't work. @vboulineau  What do you think?

ogaca-dd: In which case `CopyFolder(fullPath string...)` can be used as we don't know the full path (We don't know where the code of `datadog-agent` is)

vboulineau: I don't think there's a single use case of copying things from `datadog-agent` repo (even though, we have several ways of getting the repo path (through `git` for instance), it's especially easy in the CI), we may want to copy generated content (from `/tmp`) or downloaded content.

command/filemanager.go
ogaca-dd: Code reverted to have the same code for embeded and absolute / relative paths.

command/filemanager.go
ogaca-dd: Function `CopyAbsoluteFolder` added.

vboulineau (approved):

PR #27, opened 2023-01-02
Open RDP port for Windows VM on azure

ogaca-dd:
What does this PR do?
---------------------

Open the RDP port for Windows VM on azure.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/azure/compute/vm.go
+++ b/azure/compute/vm.go
@@ -40,7 +40,7 @@ func NewLinuxInstance(e azure.Environment, name, imageUrn, instanceType string, 
 		CustomData: userData,
 	}
 
-	return newVMInstance(e, name, imageUrn, instanceType, linuxOsProfile)
+	return newVMInstance(e, name, imageUrn, instanceType, linuxOsProfile, nil)
 }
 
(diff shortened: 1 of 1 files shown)


pducolin (approved):

azure/compute/vm.go
```diff
@@ -72,15 +72,39 @@ func NewWindowsInstance(e azure.Environment, name, imageUrn, instanceType string
 		}
 	}
 
-	vm, publicIP, nw, err := newVMInstance(e, name, imageUrn, instanceType, windowsOsProfile)
+	securityGroup, err := network.NewNetworkSecurityGroup(e.Ctx, e.Namer.ResourceName(name, "security-group-rdp"), &network.NetworkSecurityGroupArgs{
```
vboulineau, line 75: It should not be required while we already have the network bits in place. We need to understand what happens, rather than creating a NSG per VM.

vboulineau (changes requested):

azure/compute/vm.go
ogaca-dd: Can you elaborate how it should work? The network settings are the same for Windows and Linux which is strange as we want to expose RDP port for Windows only.

vboulineau: The network setup is already created (vnet, subnet, nsg). The issue was that for some reason the subnet lost the default nsg setting. Fixed it and RDP/SSH works as expected now.

ogaca-dd:
Close as not needed anymore.

PR #37, opened 2023-02-08
Extract Agent installation from VM

ogaca-dd:
What does this PR do?
---------------------
This PR Extract Agent installation from VM and move Agent installer from common to datadog/agent

Review commit by commit is recommended.

-------------------

Motivation
----------

Additional Notes
----------------


pducolin (approved):
Ok for me, what happens of the configuration option `OptionalAgentInstallParams` ? Should we delete it ? Seems to have no effect

vboulineau (approved):

PR #38, opened 2023-02-09
Add package managers for Nix OSes.

ogaca-dd:
What does this PR do?
---------------------
Add the package managers for Nix OSes.
This PR also introduces `NixVM` which is a Linux / Unix / MacOS VM. `NixVM` has extra features compared to a regular VM:
 - FileManager (Not yet supported for Windows)
 - PackageManager (Doesn't make a lot of sense for Windows Server)
 - DockerManager (Not yet supported for Windows)

For tests that should be run for all OSes including Windows, the user should use `CreateXXVM` otherwise the user should use `CreateNixXXXVM`.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/aws/scenarios/vm/ec2VM/ec2VM.go
+++ b/aws/scenarios/vm/ec2VM/ec2VM.go
@@ -16,14 +16,14 @@ func NewEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (commonvm.VM,
 	return newVM(ctx, options...)
 }
 
-// NewNixEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
+// NewUnixLikeEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
 // The returned vm provides additional methods compared to NewEc2VM
-func NewNixEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (*commonvm.NixVM, error) {
+func NewUnixLikeEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (*commonvm.UnixLikeVM, error) {
 	vm, err := newVM(ctx, options...)
 	if err != nil {
 		return nil, err
 	}
-	return commonvm.NewNixVM(vm)
+	return commonvm.NewUnixLikeVM(vm)
 }
 
 // newVM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
--- a/aws/scenarios/vm/ec2VM/ec2VM.go
+++ b/aws/scenarios/vm/ec2VM/ec2VM.go
@@ -7,12 +7,27 @@ import (
 	"github.com/DataDog/test-infra-definitions/aws"
 	awsEc2 "github.com/DataDog/test-infra-definitions/aws/ec2/ec2"
 	commonos "github.com/DataDog/test-infra-definitions/common/os"
-	"github.com/DataDog/test-infra-definitions/common/vm"
+	commonvm "github.com/DataDog/test-infra-definitions/common/vm"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
 // NewEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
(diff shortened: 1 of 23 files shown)


common/os/packagemanager.go
```diff
@@ -0,0 +1,16 @@
+package os
+
+import (
+	"github.com/DataDog/test-infra-definitions/command"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func NewAptManager(runner *command.Runner) command.PackageManager {
```
pducolin, line 8: **❓ question**
Why some managers are here and some in `aws/scenarios/vm/os/packagemanager.go` ? Are these available in both aws and azure ?

pducolin:
**❓ question**
Are NixOS https://en.wikipedia.org/wiki/NixOS ? Why do we limit docker support to Nix VMs ?

common/os/packagemanager.go
ogaca-dd: `common/os` defines `Ubuntu` and `Windows` OSes as they will be available for all datacenters. I kept the package managers in the same go package. For OSes only supported by AWS, the package manager are defined in aws package.

ogaca-dd:
> ❓ question Are NixOS https://en.wikipedia.org/wiki/NixOS ? Why do we limit docker support to Nix VMs ?

I meant `*NIX` let me rename to something more explicit.

pducolin (approved):

aws/scenarios/vm/ec2VM/ec2VM.go
```diff
@@ -7,12 +7,27 @@ import (
...
 
 // NewEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
-func NewEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (vm.VM, error) {
+func NewEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (commonvm.VM, error) {
+	return newVM(ctx, options...)
+}
+
+// NewUnixLikeEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
+// The returned vm provides additional methods compared to NewEc2VM
+func NewUnixLikeEc2VM(ctx *pulumi.Context, options ...func(*Params) error) (*commonvm.UnixLikeVM, error) {
```
vboulineau, line 21: I'm not sure `Unix` like really make. Except Linux (~same outside of package managers), we only support MacOS, which, in terms of Agent setup and capabilities, is not really similar to Linux.

ogaca-dd: The extra fields of `UnixLikeVM` compared to `VM` are:
- `FileManager`: It will be moved to `VM` once `FileManager` supports Windows.
- `DockerManager`: It is not clear, if we want to support it on Windows or rely only on ECS.
- `PackageManager`: It doesn't make sense for Windows especially for Windows server.

MacOS has a package manager and `FileManager` should work out of the box so it may make sense to have  common features with Unix. I don't have a strong opinion on it and we can exclude MacOS. What do you think?

pducolin: No strong opinion, in terms of abstraction my rule of thumbs is to skip it up to 3 similar implementation, here we have 2. But I see the advantage of sharing concepts such us those above. 

What if `VM` was a generic accepting a os type ? Something like `ec2.VM[MacOS]` or `ec2.VM[Linux]` returning all the available members ?

vboulineau: It's the design I suggested through interface + composition with `InfraVM` and `OSVM` in previous PR.
Generally speaking, I think `Linux` and `MacOS` are different enough to treat them differently.
Technically speaking the `FileManager` is tied to SSH more than being Unix-like, `DockerManager` won't be available on Mac and `PackageManager` is entirely OS-dependant.

pducolin: But `PackageManager` has an implementation on `macOS` though with `brew` with https://github.com/aws/homebrew-aws. Not sure we need it, tests based on `macOS` are mostly for the install feature. 

From a user perspective it could be confusing to have `ec2VM` and `UnixLikeVM`: if I create an `ec2VM` with linux os, I would probably expect the same result as `UnixLikeVM`.

vboulineau (approved):

PR #44, opened 2023-02-17
Add runner level user

usamasaqib:
What does this PR do?
---------------------
This PR provides support for adding a runner level option to set the user for all commands. It also extends `Args` to specify a specific `User` for individual command.
The priority level for deciding the user is: `sudo` -> `args.User` -> `runner.User`

-------------------
This impacts all scenarios.

Motivation
----------
The microvms scenario is simplified by being able to set a user for running all the commands, rather than handling permissions on a case-by-case basis.

Additional Notes
----------------

### DIFF
--- a/command/runner.go
+++ b/command/runner.go
@@ -12,8 +12,6 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-const DefaultUser = ""
-
 type Args struct {
 	Create      pulumi.StringInput
 	Update      pulumi.StringInput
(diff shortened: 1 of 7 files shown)


command/runner.go
```diff
@@ -18,20 +22,30 @@ type Args struct {
 	Stdin       pulumi.StringPtrInput
 	Environment pulumi.StringMap
 	Sudo        bool
+	User        string
```
pducolin, line 25: **💭 thought**
Wonder if there is a way to wrap a command inside a `sudo` command, with `sudo` options, instead of passing `sudo` and `user` to args. `User` makes sense only in `sudo` context, it is not ideal to have two options changing the same behaviour - it can be confusing, what if I pass `sudo = false` and `user: "something` ?

command/runner.go
```diff
@@ -45,26 +59,33 @@ func (args *Args) buildCommandInput(command pulumi.StringInput, env pulumi.Strin
...
 type Runner struct {
 	e           config.CommonEnvironment
 	namer       namer.Namer
-	connection  remote.ConnectionInput
 	waitCommand *remote.Command
+	config      runnerConfiguration
 }
 
-func NewRunner(e config.CommonEnvironment, connName string, conn remote.ConnectionInput, readyFunc func(*Runner) (*remote.Command, error)) (*Runner, error) {
+func NewRunner(e config.CommonEnvironment, connName, asUser string, conn remote.ConnectionInput, readyFunc func(*Runner) (*remote.Command, error)) (*Runner, error) {
```
pducolin, line 81: **💬 suggestion**
What about using the functional options pattern ? Or taking a config struct instead ? I don't like that we need to pass `asUser` even when we want a runner using current user, it can be confusing.

Alternatively, you could create a NewRunnerAsUser with the extra param in the signature

usamasaqib: Great suggestions! I did not know about functional options pattern. I will look into it.

command/runner.go
usamasaqib: One option is to remove `Sudo` and keep `User`. In order to run commands as root the `User` must be set to `root`. The prefix will then be `sudo -u root`. I am not sure if this is intuitive though...?

usamasaqib: Alternatively, `Sudo` can be renamed to `Root` or `AsRoot`, and the behavior can be to override `User` and run the command as root.

pducolin (approved):

command/runner.go
usamasaqib: I will revert to the old way for now, and address this in a different PR.

PR #46, opened 2023-02-22
Add deserializer for client in datadog-agent

ogaca-dd:
What does this PR do?
---------------------

This PR adds the data needed by the client in `datadog-agent` to interact with a `VM` and with an `Agent`.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/datadog/agent/install.go
+++ b/datadog/agent/install.go
@@ -12,11 +12,13 @@ import (
 	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/DataDog/test-infra-definitions/common/vm"
 	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/auto"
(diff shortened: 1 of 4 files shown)


datadog/agent/install.go
```diff
@@ -142,3 +144,20 @@ func installIntegrations(
 
 	return lastCommand, utils.StrHash(parts...), nil
 }
+
+type ClientData struct {
+	Connection utils.Connection
+}
+
+func (installer *Installer) GetClientDataDeserializer() func(auto.UpResult) (*ClientData, error) {
+	vmDataDeserializer := installer.vm.GetClientDataDeserializer()
```
pducolin, line 153: **❓ question**
Why not returning `installer.vm.GetClientDataDeserializer` ?

pducolin (approved):

datadog/agent/install.go
ogaca-dd: The goal is to have different types between `Installer` and `installer.vm.GetClientDataDeserializer` to enforce strong typing.

PR #47, opened 2023-02-22
Comment `MacosOS` vm type until it is really supported.

ogaca-dd:
What does this PR do?
---------------------

Comment `MacosOS` vm type until it is really supported.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/aws/scenarios/vm/os/os.go
+++ b/aws/scenarios/vm/os/os.go
@@ -16,14 +16,14 @@ type OS interface {
 type Type int
 
 const (
-	WindowsOS     Type = iota
-	UbuntuOS           = iota
-	MacosOS            = iota
-	AmazonLinuxOS      = iota
-	DebianOS           = iota
-	RedHatOS           = iota
-	SuseOS             = iota
-	FedoraOS           = iota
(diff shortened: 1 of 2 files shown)


aws/scenarios/vm/os/os.go
```diff
@@ -32,8 +32,8 @@ func GetOS(env aws.Environment, osType Type) (OS, error) {
 		return newWindows(env), nil
 	case UbuntuOS:
 		return newUbuntu(env), nil
-	case MacosOS:
-		return newMacOS(env), nil
+	//case MacosOS:
+	//return newMacOS(env), nil // Not yet supported
```
pducolin, line 36: **💭 thought**
What about returning a specific error instead ?
```suggestion
	case MacosOS:
	    return nil, errors.New("MacosOS is not supported yet")
```

pducolin (approved):

aws/scenarios/vm/os/os.go
ogaca-dd: I want to avoid someone writing code for MacOS and realise later it is not supported.

PR #53, opened 2023-02-27
Better handling of AWS providers

vboulineau:
What does this PR do?
---------------------

-------------------

Motivation
----------

Additional Notes
----------------


aws/environment.go
```diff
@@ -66,24 +66,36 @@ func NewEnvironment(ctx *pulumi.Context) (Environment, error) {
...
 	}
 
 	return env, nil
 }
 
+func (e *Environment) InvokeProviderOption() pulumi.InvokeOption {
+	return pulumi.Provider(e.awsProvider)
+}
+
+func (e *Environment) ResourceProvidersOption() pulumi.ResourceOption {
```
pducolin, line 95: **❓ question**
What is the difference between a `pulumi.ResourceOption` and a `pulumi.InvokeOption` ? I see `pulumi.Provider(e.Provider)` was changed to one of the two, can you give more details on why you used one and not the other ?

common/config/helpers.go
```diff
@@ -1,10 +1,21 @@
 package config
 
 import (
+	"strings"
+
 	"github.com/Masterminds/semver"
 	"github.com/pulumi/pulumi/sdk/v3/go/auto"
 )
 
+func FindEnvironmentName(environments []string, prefix string) string {
```
pducolin, line 10: **💬 suggestion**
This name could be more explicit: this function returns the first element in a list that is prefixed by a given prefix, it could become `FindEnvironmentNameByPrefix`

pducolin (approved):
Looks good, curious at what is the difference between `pulumi.ResourceOption` and `pulumi.InvokeOption`

common/config/helpers.go
vboulineau: To me, the parameters are part of the function signature (even though `Go` does not support overloading), so `FindEnvironmentNameByPrefix(prefix)` is redundant.

aws/environment.go
vboulineau: It's some Pulumi arcane, it's not documented so I'm not 100% sure, but it's the different types required by the functions:
```
NewXXX creates Resources, hence require `ResourceOption`
Lookup/GetXXX gets some value, hence require `InvokeOption`
```

The `pulumi.Provider(e.Provider)` returns an object that implements both interfaces, and from what I can see in the implementation, `InvokeOption` only tolerate a single provider (unlike `ResourceOption`).

pducolin: Thank you !

PR #57, opened 2023-03-08
[ecs] agent: network mode should be awsvpc for correct instantiation

truthbk:
What does this PR do?
---------------------
This PR addresses an issue where the ECS deployments failed due to an incorrect network mode being defined for EC2-based ECS cluster tasks. 

-------------------
ECS EC2(classic) based deployments

Motivation
----------
`bridge` mode was resulting in the following cluster/task instantiation error:
```
 * creating ECS Service (<id>-ec2-linux-dd-agent): InvalidParameterException: Network Configuration is not valid for the given networkMode of this task definition.
```

Changing the mode to `awsvpc` seems to address the problem.

source: https://stackoverflow.com/questions/59678757/awsvpc-network-configuration-is-not-valid-for-the-given-networkmode-of-this-tas

Additional Notes
----------------


pducolin (approved):

pducolin:
Thank you for addressing this !

PR #58, opened 2023-03-09
Add an invoke task to create and destroy a vm

ogaca-dd:
What does this PR do?
---------------------

Add an invoke task to create and destroy a vm.
Note: More options are going to be added to configure a vm in a later PR like for configuring the operating system and the architecture.

### DIFF
--- a/tasks/stack.py
+++ b/tasks/stack.py
@@ -1,0 +1,33 @@
+import getpass
+import invoke
+
+stack_vm = "vm"
+stacks = {stack_vm: ("./aws/scenarios/vm/", "aws-vm")}
+
+
+def get_stack_name(scenario_name):
+    stack_name = stacks[scenario_name][1]
+    return getpass.getuser() + "-" + stack_name
+
+
+def get_scenario_folder(scenario_name):
+    return stacks[scenario_name][0]
+
+
+def get_scenario_folder_from_stack_name(full_stack_name):
+    scenario_folder = find_scenario_folder(full_stack_name)
+    if scenario_folder is not None:
+        return scenario_folder
+    raise invoke.Exit(
+        f"Cannot find the associated scenario for the stack '{full_stack_name}'. Was this stack created outside invoke tasks?")
+
+
+def find_scenario_folder(full_stack_name):
+    for scenario_folder, stack_name in stacks.values():
+        if full_stack_name.endswith(stack_name):
+            return scenario_folder
+    return None
+
+
+def is_known_stack(full_stack_name):
+    return find_scenario_folder(full_stack_name) is not None
--- a/tasks/vm.py
+++ b/tasks/vm.py
@@ -1,0 +1,51 @@
+from invoke import task
+from .config import get_config
+import subprocess
+import os
+import invoke
+from .stack import stack_vm, get_scenario_folder, get_stack_name
+
+
+@task(help={
+    'install_agent': "Install the Agent (default True).",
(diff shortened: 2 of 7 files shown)


pducolin (approved):
I like how this will help discover environments and options !

Have you considered using `golang` task managers ? I can see how agent teams are used to `invoke` so ok for using it, I remember investigating about go solutions in the past and there are good alternatives, just in case we want to keep this 100% golang

tasks/vm.py
```diff
@@ -0,0 +1,51 @@
...
+    aws_deploy(
+        config,
+        stack_vm,
+        install_agent,
+        f"ddagent:deploy={install_agent}",
+        f"ddagent:version={agent_version}",
+    )
+
+
+def aws_deploy(config, scenario_name, install_agent, *args):
```
vboulineau, line 30: I think this belongs to another file

tasks/vm.py
```diff
@@ -0,0 +1,51 @@
...
+        if api_key is None or len(api_key) != 32:
+            raise invoke.Exit("Invalid API KEY. You must define the environment variable 'DD_API_KEY'. " +
+                              "If you don't want an agent installation add '--no-install-agent'.")
+
+    scenario_folder = get_scenario_folder(scenario_name)
+    stack_name = get_stack_name(scenario_name)
+    cmd_args = [
+        "aws-vault", "exec", "sandbox-account-admin", "--",
+        "pulumi", "up",
+        "-c", "ddinfra:aws/defaultKeyPairName=" + config["aws_key_pair"],
```
vboulineau, line 43: The way I would see it is a `ParameterBuilder` that generates option based on `Configuration` dataclass.

tasks/vm.py
```diff
@@ -0,0 +1,51 @@
...
+    if install_agent:
+        api_key = os.getenv("DD_API_KEY")
+        if api_key is None or len(api_key) != 32:
+            raise invoke.Exit("Invalid API KEY. You must define the environment variable 'DD_API_KEY'. " +
+                              "If you don't want an agent installation add '--no-install-agent'.")
+
+    scenario_folder = get_scenario_folder(scenario_name)
+    stack_name = get_stack_name(scenario_name)
+    cmd_args = [
+        "aws-vault", "exec", "sandbox-account-admin", "--",
```
vboulineau, line 41: I don't think we can really enforce that. It does not on all setups (for instance, I use a `ServiceAccount`), it also does not work in CI if we implement `stack as a service`.

I think we need some run profiles  to handle that. The wrapping logic (i.e. running with `aws-vault`), should be separated from the construction of the parameters and Pulumi commands.

vboulineau (commented):
Also requires a `requirements.txt` file.

tasks/stack.py
```diff
@@ -0,0 +1,33 @@
...
+def get_stack_name(scenario_name):
+    stack_name = stacks[scenario_name][1]
+    return getpass.getuser() + "-" + stack_name
+
+
+def get_scenario_folder(scenario_name):
+    return stacks[scenario_name][0]
+
+
+def get_scenario_folder_from_stack_name(full_stack_name):
```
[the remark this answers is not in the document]
ogaca-dd: >. This methods is perhaps not entirely reliable.
Why?

tasks/vm.py
ogaca-dd: I going to move it to another file when adding new command in a later PR.

ogaca-dd:
Closed in favor of https://github.com/DataDog/test-infra-definitions/pull/68

PR #59, opened 2023-03-10
Wait OpenSSH is ready for Windows host

ogaca-dd:
What does this PR do?
---------------------

Wait OpenSSH to be ready for Windows host


common/vm/vm.go
```diff
@@ -33,9 +33,24 @@ func NewGenericVM(
 
 	readyFunc := func(r *command.Runner) (*remote.Command, error) { return command.WaitForCloudInit(ctx, r) }
 	if os.GetType() == commonos.WindowsType {
-		// On Windows, there is no equivalent of cloud init, but the code wait until ssh connection is ready
-		// so it is OK to not have a ready function.
-		readyFunc = nil
+		cmd := `for ($i = 0; $i -le 120; $i++) {
```
pducolin, line 36: Are you using `120` to wait for 2 minutes ?

pducolin (approved):

common/vm/vm.go
ogaca-dd: Yes

vboulineau, line 36: As you note below, the script is not useless as running it means that the connection was successful. It's simpler and clearer to just inline `Create: <dummy command>` (like `Get-Date` or `echo "connected").

However, given the time it takes on Windows, I would use the new `DialErrorLimit` field introduced in newer `remote` version:
https://github.com/pulumi/pulumi-command/pull/180

ogaca-dd: It is still relevant to wait ssh is running to consider the VM ready even if most of the waiting time is spend to try running the command. I will update the comment.

ogaca-dd: For `DialErrorLimit`, I think running a command on Windows is part of the regular workflow and I guess the default value for `DialErrorLimit` should be enough. It is still good to know about `DialErrorLimit` especially when dealing with reboot.

vboulineau: I am not sure this code is going to be understood clearly by the next person working on this code. You're using SSH to check if SSH is running and yet looping 120 times to check if SSH is running, avoiding the only option that can make this non flaky (e.g. `DialErrorLimit`) as we have no control over when the `userData` script installing OpenSSH will be run, nor how long it will take to run.

The default value of `DialErrorLimit` is `10` with an exponential backoff starting a `100ms` and with a `1.5` multiplier, leading to a maximum of retrying for ~11s (considering the connection attempt fails ~immediately, which, I believe, should be the case if firewalls are set to allow connections, even though we may hit a `connect()` timeout on the first attempt as the IP may not be reachable, artificially extending our wait time by ~2-3 minutes).

Any extra attempt is going to wait 5s, so I believe that setting `DialErrorLimit` to `20` to always wait at least 1 minute would be safer.

ogaca-dd: > You're using SSH to check if SSH is running

Does `remote.Command` require OpenSSH to work? Please note that this code wait for OpenSSH to be ready which is installed by [here](https://github.com/DataDog/test-infra-definitions/blob/9db59f289fe6e74be3298bd131f20cba103cc0ff/aws/scenarios/vm/ec2VM/ec2VM.go#L71-L90).

vboulineau: Yes, `remote.Command` is exclusively supporting SSH.

ogaca-dd: I didn't know this requirement.
About `DialErrorLimit` I don't think it is needed as typical wait duration is ~120 seconds whereas the default timeout is more than 500s.

ogaca-dd: As discussed privately, I will merge this PR.

PR #62, opened 2023-03-20
Remove registration for scenarios replace by hardcoded list in registry

vboulineau:
What does this PR do?
---------------------

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/registry/scenarios.go
+++ b/registry/scenarios.go
@@ -3,23 +3,34 @@ package registry
 import (
 	"strings"
 
+	dockervm "github.com/DataDog/test-infra-definitions/aws/scenarios/dockerVM"
+	"github.com/DataDog/test-infra-definitions/aws/scenarios/ecs"
+	"github.com/DataDog/test-infra-definitions/aws/scenarios/eks"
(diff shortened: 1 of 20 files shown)


registry/scenarios.go
```diff
@@ -3,23 +3,34 @@ package registry
...
 
-type scenarioRegistry map[string]pulumi.RunFunc
-
-var Scenarios = make(scenarioRegistry)
+type ScenarioRegistry map[string]pulumi.RunFunc
 
-// Register lower case the given name
-func (s scenarioRegistry) Register(name string, rf pulumi.RunFunc) {
-	s[strings.ToLower(name)] = rf
+func Scenarios() ScenarioRegistry {
```
pducolin, line 18: **❓ question**
@ogaca-dd what is the advantage of registering scenarios here, rather than have each scenario registering itself independently ? 

I see the cons of this implementation as leading to possibly forgetting scenarios, as adding one requires editing 2 files instead of just one - the scenario `run`. 
Another cons of this implementation is that we would edit the same file when adding scenarios, leading to possible git conflicts.

PR #63, opened 2023-03-21
Add fakeintake service on ECS

pducolin:
What does this PR do?
---------------------

Add helper to deploy fakeintake to ECS Fargate cluster dedicated to fakeintake

-------------------

None

Motivation
----------

This will help deploying fakeintake by default from e2e test 

Additional Notes
----------------

I used a temporary scenario for testing this that I did not push, you can create one temporary by pasting the following code and registering the scenario

```golang
package fakeintaketest

import (
	"github.com/DataDog/test-infra-definitions/aws"
	"github.com/DataDog/test-infra-definitions/datadog/fakeintake"

)

const (
	fakeintageFargateArn = "arn:aws:ecs:us-east-1:601427279990:cluster/fakeintake-ecs"
)

func Run(ctx *pulumi.Context) error {
	env, err := aws.NewEnvironment(ctx)
	if err != nil {
		return err
	}

	ipAddress, err := fakeintake.ECSFargateService(env, "fakeintake-001", fakeintageFargateArn)
	if err != nil {
		return err
	}

	ctx.Export("ecs-fargate-fakeintake-ip-address", ipAddress)

	return err
}
```


datadog/fakeintake/ecsFargate.go
```diff
@@ -0,0 +1,89 @@
...
+		TaskRole: &awsx.DefaultRoleWithPolicyArgs{
+			RoleArn: pulumi.StringPtr(e.ECSTaskRole()),
+		},
+		Family: e.CommonNamer.DisplayName(pulumi.String("fakeintake-ecs")),
+	}, e.ResourceProvidersOption())
+}
+
+func fargateLinuxContainerDefinition() *ecs.TaskDefinitionContainerDefinitionArgs {
+	return &ecs.TaskDefinitionContainerDefinitionArgs{
+		Cpu:         pulumi.IntPtr(100),
```
vboulineau, line 73: Misleading, container-level limit are not applied unless there's CPU contention. As this is the only container in the Task, it will use everything available (up to task cpu limit).

pducolin: So I should omit them ?

vboulineau: You can omit Memory, I think you need to explicitly set CPU to `0` as the Pulumi provider mandates it.

pducolin: I see they're required in the `TaskDefinition`, not int the `TaskDefinitionContainerDefinition`, removing this

vboulineau (approved):

aws/ecs/client.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func NewECSClient(ctx context.Context, region string) (*Client, error) {
+	cfg, err := awsConfig.LoadDefaultConfig(ctx,
+		awsConfig.WithRegion(region),
+	)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{Client: *awsECS.NewFromConfig(cfg), ctx: ctx}, nil
```
ogaca-dd, line 25: If `NewFromConfig` returns `*awsECS.Client` then it would make sense  that `Client` has the same type ( `*awsECS.Client` vs  `awsECS.Client`).

aws/ecs/client.go
```diff
@@ -0,0 +1,68 @@
...
+		Cluster:     &clusterArn,
+		ServiceName: &serviceName,
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(taskList.TaskArns) < 1 {
+		return "", errors.New("no task arn found")
+	}
+	return taskList.TaskArns[0], nil
```
ogaca-dd, line 47: Can `c.ListTasks` returns more than one value?

aws/ecs/client.go
```diff
@@ -0,0 +1,68 @@
...
+	if len(taskOutput.Tasks) < 1 {
+		return "", fmt.Errorf("no task found on cluster %s with arn %s", clusterArn, taskArn)
+	}
+	if len(taskOutput.Tasks[0].Containers) < 1 {
+		return "", fmt.Errorf("no container found on cluster %s with arn %s", clusterArn, taskArn)
+	}
+	if len(taskOutput.Tasks[0].Containers[0].NetworkInterfaces) < 1 {
+		return "", fmt.Errorf("no network interface found on cluster %s with arn %s", clusterArn, taskArn)
+	}
+	return *taskOutput.Tasks[0].Containers[0].NetworkInterfaces[0].PrivateIpv4Address, nil
```
ogaca-dd, line 67: What do you think about introducing local variables to avoid having 4 `.`?

datadog/fakeintake/ecsFargate.go
```diff
@@ -0,0 +1,89 @@
...
+		return pulumi.StringOutput{}, err
+	}
+	fargateService, err := tidECS.FargateService(e, e.Namer.ResourceName(name), pulumi.String(clusterArn), taskDef.TaskDefinition.Arn())
+
+	// Hack passing taskDef.TaskDefinition.Arn() to execute apply function
+	// when taskDef has an ARN, thus it is defined on AWS side
+	ipAddress = pulumi.All(taskDef.TaskDefinition.Arn(), fargateService.Service.Name()).ApplyT(func(args []any) (string, error) {
+		var ipAddress string
+		err := backoff.Retry(func() error {
+			fmt.Println("waiting for task private ip")
```
ogaca-dd, line 35: I think this is not displayed when calling e2e automation API

aws/ecs/client.go
pducolin: It should not as we deploy one single task to one single service, but the API response returns a list of tasks. I can return an error if there is more than one task as it would be unexpected.

datadog/fakeintake/ecsFargate.go
pducolin: Moving to pulumi.logs.

aws/ecs/client.go
pducolin: Thank you, I thought struct embedding required values vs reference

ogaca-dd (approved):

PR #65, opened 2023-03-28
Add Windows support for Filemanager

ogaca-dd:
What does this PR do?
---------------------
This PR adds Windows support to methods of Filemanager.
It also adds the support of running commands on Windows correctly.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/command/filemanager.go
+++ b/command/filemanager.go
@@ -77,7 +77,8 @@ func (fm *FileManager) CopyRelativeFolder(relativeFolder string, remoteFolder st
 func (fm *FileManager) CopyAbsoluteFolder(absoluteFolder string, remoteFolder string, opts ...pulumi.ResourceOption) ([]pulumi.Resource, error) {
 	baseFolder := filepath.Base(absoluteFolder)
 	rootWithoutBase := absoluteFolder[:len(absoluteFolder)-len(baseFolder)]
-	return fm.CopyFSFolder(absoluteFolder, os.DirFS(rootWithoutBase), baseFolder, remoteFolder, opts...)
+	// Use remoteFolder as `absoluteFolder` may be a random file path that is different for each run.
+	return fm.CopyFSFolder(remoteFolder, os.DirFS(rootWithoutBase), baseFolder, remoteFolder, opts...)
 }
 
 // CopyRelativeFile copies relative path to a remote path.
@@ -138,7 +139,7 @@ func (fm *FileManager) CopyFSFolder(
 			return nil, err
 		}
 		fileCommand, err := fm.CopyInlineFile(
-			resourceName+"-"+file,
+			resourceName+"-"+destFile, // Use destfile as `file` may be a random file path that is different for each run
 			pulumi.String(fileContent),
 			path.Join(remoteFolder, destFile),
 			useSudo,
--- a/command/filemanager.go
+++ b/command/filemanager.go
@@ -10,70 +10,49 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-const (
-	linuxTempDir = "/tmp"
-)
-
 type FileManager struct {
-	runner *Runner
+	runner  *Runner
+	command osCommand
 }
 
 func NewFileManager(runner *Runner) *FileManager {
 	return &FileManager{
-		runner: runner,
+		runner:  runner,
+		command: runner.osCommand,
 	}
 }
 
 func (fm *FileManager) CreateDirectory(name string, remotePath pulumi.StringInput, useSudo bool, opts ...pulumi.ResourceOption) (*remote.Command, error) {
-	return fm.runner.Command(name,
-		&Args{
-			Create: pulumi.Sprintf("mkdir -p %s", remotePath),
-			Delete: pulumi.Sprintf("rm -rf %s", remotePath),
-			Sudo:   useSudo,
-		}, opts...)
+	return fm.command.CreateDirectory(fm.runner, name, remotePath, useSudo, opts...)
 }
 
-func (fm *FileManager) TempDirectory(name string, opts ...pulumi.ResourceOption) (*remote.Command, string, error) {
-	tempDir := path.Join(linuxTempDir, name)
-	folderCmd, err := fm.CreateDirectory("tmpdir-"+name, pulumi.String(tempDir), false, opts...)
-	return folderCmd, tempDir, err
+func (fm *FileManager) TempDirectory(resourceName string, opts ...pulumi.ResourceOption) (*remote.Command, string, error) {
+	return fm.command.CreateTemporaryFolder(fm.runner, resourceName, opts...)
 }
 
 func (fm *FileManager) CopyFile(localPath, remotePath string, opts ...pulumi.ResourceOption) (*remote.CopyFile, error) {
-	return remote.NewCopyFile(fm.runner.e.Ctx, fm.runner.namer.ResourceName("copy", utils.StrHash(localPath, remotePath)), &remote.CopyFileArgs{
+	// If the file was previously created, make sure to delete it before creating it.
+	opts = append(opts, pulumi.DeleteBeforeReplace(true))
+
+	return remote.NewCopyFile(fm.runner.e.Ctx, fm.runner.namer.ResourceName("copy", remotePath), &remote.CopyFileArgs{
 		Connection: fm.runner.config.connection,
 		LocalPath:  pulumi.String(localPath),
 		RemotePath: pulumi.String(remotePath),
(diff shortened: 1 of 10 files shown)


usamasaqib (approved):

command/windowsOSCommand.go
```diff
@@ -0,0 +1,74 @@
...
+	remotePath string,
+	useSudo bool,
+	append bool,
+	opts ...pulumi.ResourceOption) (*remote.Command, error) {
+	flags := ""
+	if append {
+		flags = "-Append"
+	}
+
+	createCmd := pulumi.Sprintf(`[System.Console]::In.ReadToEnd() | Out-File %v -FilePath %v`, flags, remotePath)
```
ogaca-dd, line 48: I have to update this line to handle binary file.

command/unixOSCommand.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/alessio/shellescape"
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+const (
+	linuxTempDir = "/tmp"
+)
+
+var _ osCommand = (*unixOSCommand)(nil)
```
pducolin, line 17: **❓ question**
What is this used for ?

command/unixOSCommand.go
```diff
@@ -0,0 +1,85 @@
...
+	runner *Runner,
+	name string,
+	fileContent pulumi.StringInput,
+	remotePath string,
+	useSudo bool,
+	append bool,
+	opts ...pulumi.ResourceOption) (*remote.Command, error) {
+	var createCmd pulumi.StringInput
+	if append {
+		createCmd = utils.AppendStringCommand(remotePath, useSudo)
```
pducolin, line 51: **❓ question**
Are append and write string command shared across windows and unix ? I see they are in `utils`

command/windowsOSCommand.go
pducolin: `ReadToEnd()` is only for text files ?

pducolin (approved):
Huge work !! 🤯 

Some comments and questions, no blocker

command/runner.go
```diff
@@ -124,10 +107,11 @@ func WithLocalUser(user string) func(*LocalRunner) {
 	}
 }
 
-func NewLocalRunner(e config.CommonEnvironment, options ...func(*LocalRunner)) *LocalRunner {
+func NewLocalRunner(e config.CommonEnvironment, isWindows bool, options ...func(*LocalRunner)) *LocalRunner {
```
vboulineau, line 110: `LocalRunner` is, well, local, so you can rely on `GOOS`.

command/unixOSCommand.go
ogaca-dd: This is the Go way to check a struct implement an interface.

command/unixOSCommand.go
ogaca-dd: No there are not shared. Please note that `AppendStringCommand` and `WriteStringCommand` are removed in https://github.com/DataDog/test-infra-definitions/pull/67

command/filemanager.go
```diff
@@ -10,70 +10,49 @@ import (
...
-	opts ...pulumi.ResourceOption) (*remote.Command, error) {
-	return fm.runner.Command(name,
-		&Args{
-			Create:   createCmd,
-			Stdin:    fileContent,
-			Sudo:     useSudo,
-			Triggers: pulumi.Array{pulumi.String(remotePath), fileContent, pulumi.BoolPtr(useSudo)},
-		}, opts...)
+func (fm *FileManager) AppendInlineFile(name string, fileContent pulumi.StringInput, remotePath string, useSudo bool, opts ...pulumi.ResourceOption) (*remote.Command, error) {
+	return fm.command.CopyInlineFile(fm.runner, name, fileContent, remotePath, useSudo, true, opts...)
```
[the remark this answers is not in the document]
ogaca-dd: This function is removed in https://github.com/DataDog/test-infra-definitions/pull/67. 
Also `CopyInlineFile` has less parameters in https://github.com/DataDog/test-infra-definitions/pull/67 (`name` is removed) but indeed I am wondering if `useSudo` should be an optional parameter?

command/windowsOSCommand.go
ogaca-dd: `ReadToEnd()` handles only string.

command/filemanager.go
```diff
@@ -10,70 +10,49 @@ import (
...
-	folderCmd, err := fm.CreateDirectory("tmpdir-"+name, pulumi.String(tempDir), false, opts...)
-	return folderCmd, tempDir, err
+func (fm *FileManager) TempDirectory(resourceName string, opts ...pulumi.ResourceOption) (*remote.Command, string, error) {
+	return fm.command.CreateTemporaryFolder(fm.runner, resourceName, opts...)
 }
 
 func (fm *FileManager) CopyFile(localPath, remotePath string, opts ...pulumi.ResourceOption) (*remote.CopyFile, error) {
-	return remote.NewCopyFile(fm.runner.e.Ctx, fm.runner.namer.ResourceName("copy", utils.StrHash(localPath, remotePath)), &remote.CopyFileArgs{
+	// If the file was previously created, make sure to delete it before creating it.
+	opts = append(opts, pulumi.DeleteBeforeReplace(true))
```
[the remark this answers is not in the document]
ogaca-dd: Indeed this is not needed for `CopyFile`.

command/filemanager.go
```diff
@@ -10,70 +10,49 @@ import (
...
-	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-const (
-	linuxTempDir = "/tmp"
-)
-
 type FileManager struct {
```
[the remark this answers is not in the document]
ogaca-dd: Indeed but I already spent too many time fixing `FileManager`. Feel free to update the code in a later PR.

command/runner.go
```diff
@@ -80,13 +48,20 @@ func WithUser(user string) func(*Runner) {
 	}
 }
 
-func NewRunner(e config.CommonEnvironment, connName string, conn remote.ConnectionInput, readyFunc func(*Runner) (*remote.Command, error), options ...func(*Runner)) (*Runner, error) {
+func NewRunner(
+	e config.CommonEnvironment,
+	connName string,
+	conn remote.ConnectionInput,
+	readyFunc func(*Runner) (*remote.Command, error),
+	isWindows bool,
```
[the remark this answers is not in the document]
ogaca-dd: The issue is that `command.NewRunner` is called by [microvms](https://github.com/DataDog/test-infra-definitions/blob/07c1ed49352ff31eb4d8e5bb55b6dd4ddb0f523d/aws/scenarios/microVMs/microvms/run.go#L157) which would require two calls of `newUnixOSCommand()` . After a second though, I will remove `isWindows`.

command/runner.go
ogaca-dd: I have injected the osCommand.

command/unixOSCommand.go
pducolin: Nice, thank you !

pducolin (approved):

command/filemanager.go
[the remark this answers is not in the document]
ogaca-dd: > I'm not sure why FileManager is not an interface

I think it is better that `FileManager` is a struct and takes an interface like `osCommand` which is specific to the operating system. The reason is that you may want to not expose directly the low level function of `osCommand`
```go
func (fm *FileManager) MyFunction() {
        // Do something independent of the OS
       fm.command.MyLowLevelFunction()
       // Do something else independent of the OS
}
```

Note: I have realised that `CreateTemporaryDirectory` shouldn't be in `osCommand` and `osCommand` should only provide a `TemporaryDirectory`.

In addition, in Go, it is not straightforward to implement the pattern for `MyFunction` with a `WindowsFileManager` and `UnixFileManager`. Indeed, in order to factorize the common code, `UnixFileManager` would need to have an embedded struct that provide `MyFunction`. The embedded struct would require an interface to access `MyLowLevelFunction` which would be implemented by `UnixFileManager`. There is a loop in the initialisation of `UnixFileManager` and the embedded struct.

> I'm not sure that putting a mix of everything in *OSCommand makes sense

If you follow the [Interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle) from  [SOLID](https://en.wikipedia.org/wiki/SOLID) principle,  `FileManager` should only require the methods it uses, ie, everything except `BuildCommandString`. To do so, you could have a `osCommand` that provides a `Filesystem` interface and a `Process` interface for example. 

I think this is not needed for now because:
- There are not many functions and don't expect to grow that much 
- It would be easy to change the design later
- Changing  the design now or later will cost nearly the same
- If there are no new methods require, the current code is simple [KISS](https://en.wikipedia.org/wiki/KISS_principle).

If you have a strong opinion that `OSCommand` should be split let me know.

PR #66, opened 2023-04-04
Sudo docker

ogaca-dd:
What does this PR do?
---------------------

Use sudo for Docker commands.


ogaca-dd:
It seems this is not the right way to fix the issue. Should we keep it until a right fix is available?

pducolin:
> It seems this is not the right way to fix the issue. Should we keep it until a right fix is available?

Why this is not the right way to fix it ? Should docker run without sudo ?

pducolin:
@ogaca-dd  I'm investigating on this, ssh-ed into the VM and found the user is not part of `docker` group.

PR #67, opened 2023-04-04
Implement Pulumi delete action for CopyInlineFile

ogaca-dd:
What does this PR do?
---------------------

Implement Pulumi delete action for `CopyInlineFile`. This is needed to undo `WithAgentConfig`

### DIFF
--- a/command/unixOSCommand.go
+++ b/command/unixOSCommand.go
@@ -39,7 +39,6 @@ func (unixOSCommand) CreateDirectory(
 
 func (unixOSCommand) CopyInlineFile(
 	runner *Runner,
-	name string,
 	fileContent pulumi.StringInput,
 	remotePath string,
 	useSudo bool,
@@ -53,7 +52,7 @@ func (unixOSCommand) CopyInlineFile(
 	backupCmd := fmt.Sprintf("if [ -f '%v' ]; then %v mv -f '%v' '%v'; fi", remotePath, sudo, remotePath, backupPath)
 	createCmd := fmt.Sprintf("(%v) && cat - | %s tee %s > /dev/null", backupCmd, sudo, remotePath)
 	deleteCmd := fmt.Sprintf("if [ -f '%v' ]; then %v mv -f '%v' '%v'; else %v rm -f '%v'; fi", backupPath, sudo, backupPath, remotePath, sudo, remotePath)
-	return copyInlineFile(name, runner, fileContent, useSudo, createCmd, deleteCmd, opts...)
+	return copyInlineFile(remotePath, runner, fileContent, useSudo, createCmd, deleteCmd, opts...)
 }
 
 func (fs unixOSCommand) CreateTemporaryFolder(runner *Runner, resourceName string, opts ...pulumi.ResourceOption) (*remote.Command, string, error) {
--- a/command/unixOSCommand.go
+++ b/command/unixOSCommand.go
@@ -44,14 +44,9 @@ func (unixOSCommand) CopyInlineFile(
 	fileContent pulumi.StringInput,
 	remotePath string,
 	useSudo bool,
-	append bool,
 	opts ...pulumi.ResourceOption) (*remote.Command, error) {
-	var createCmd pulumi.StringInput
-	if append {
-		createCmd = utils.AppendStringCommand(remotePath, useSudo)
-	} else {
-		createCmd = utils.WriteStringCommand(remotePath, useSudo)
-	}
+
+	createCmd := utils.WriteStringCommand(remotePath, useSudo)
 	return copyInlineFile(name, runner, fileContent, useSudo, createCmd, opts...)
 }
 
--- a/command/unixOSCommand.go
+++ b/command/unixOSCommand.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"path"
 
-	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/alessio/shellescape"
 	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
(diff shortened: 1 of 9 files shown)


command/unixOSCommand.go
```diff
@@ -40,19 +39,20 @@ func (unixOSCommand) CreateDirectory(
...
-	} else {
-		createCmd = utils.WriteStringCommand(remotePath, useSudo)
+
+	sudo := ""
+	if useSudo {
+		sudo = "sudo"
 	}
-	return copyInlineFile(name, runner, fileContent, useSudo, createCmd, opts...)
+	backupPath := remotePath + "." + backupExtension
+	backupCmd := fmt.Sprintf("if [ -f '%v' ]; then %v mv -f '%v' '%v'; fi", remotePath, sudo, remotePath, backupPath)
```
vboulineau, line 52: Adding this backup adds complexity but I don't see the added value given that we use automation.

datadog/agent/install.go
```diff
@@ -89,11 +87,13 @@ func updateAgentConfig(
...
+	for _, extraConfig := range extraAgentConfig {
+		agentConfig = agentConfig + "\n" + extraConfig
+	}
+
 	if agentConfig != "" {
-		agentConfigWithAPIKEY := env.AgentAPIKey().ApplyT(func(apiKey string) pulumi.String {
-			config := strings.ReplaceAll(agentConfig, "{{API_KEY}}", apiKey)
-			return pulumi.String(config)
-		}).(pulumi.StringInput)
+		agentConfigWithAPIKEY := pulumi.Sprintf("api_key: %v\n%v", env.AgentAPIKey(), agentConfig)
```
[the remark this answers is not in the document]
pducolin: **❓ question**
Is `env.AgentAPIKey()` safely checked and mandatory ? If empty, it would break the agent config load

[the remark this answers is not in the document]
pducolin: What happens if `api_key` is already there ?

command/unixOSCommand.go
pducolin: Isn't `deleteCmd` called even by automation, on stack update ?

command/unixOSCommand.go
```diff
@@ -46,8 +45,15 @@ func (unixOSCommand) CopyInlineFile(
...
-	createCmd := utils.WriteStringCommand(remotePath, useSudo)
-	return copyInlineFile(name, runner, fileContent, useSudo, createCmd, opts...)
+	sudo := ""
+	if useSudo {
+		sudo = "sudo"
+	}
+	backupPath := remotePath + "." + backupExtension
+	backupCmd := fmt.Sprintf("if [ -f '%v' ]; then %v mv -f '%v' '%v'; fi", remotePath, sudo, remotePath, backupPath)
+	createCmd := fmt.Sprintf("(%v) && cat - | %s tee %s > /dev/null", backupCmd, sudo, remotePath)
+	deleteCmd := fmt.Sprintf("if [ -f '%v' ]; then %v mv -f '%v' '%v'; else %v rm -f '%v'; fi", backupPath, sudo, backupPath, remotePath, sudo, remotePath)
```
pducolin, line 54: **❓ question**
Should you also remove the backup ? Would it collide with a new backup creation, or would it be overwritten ?

command/unixOSCommand.go
ogaca-dd: If you call `CopyInlineFile(content, remotePath1, false)` in an environment creation and then update the environment but without calling `CopyInlineFile(content, remotePath1, false)` then you want `remotePath1` to be removed.
For example, if you set the configuration of the Agent in one test but not in another one, then the change must be reverted.

datadog/agent/install.go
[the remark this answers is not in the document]
ogaca-dd: `env.AgentAPIKey()` is [mandatory](https://github.com/DataDog/test-infra-definitions/blob/67d9b1bee16a4e26ae2ee0d786e8047e85e0962f/common/config/environment.go#L97)

> What happens if api_key is already there

The Agent takes the second api key (the one provided by the user) and displays a warning.

Note: From datadog-agent repository, you will not have an easy way to have the API KEY.

command/unixOSCommand.go
ogaca-dd: because of `mv -f` old backup would be overwritten.

pducolin (approved):

command/unixOSCommand.go
vboulineau: Yes but you don't a backup for that, you just need for `deleteCmd` to delete the file.

I think the thing you want to support is a file that was existing before any test (or at least before a `Stack/TestSuite` execution), is changed by `test1` then not changed by `test2` (so you'd like to revert to original file).
However I think this logic does not really work: if there's a change by `test1`, then a change by `test2`, then "not changed" by `test3`, you will rollback to `test2` version instead of rolling back to original version.

Also for `WithAgentConfig`, this is non-issue, the configuration must be given, in full, all the time. So not giving it implies empty `datadog.yaml` (which is default)

ogaca-dd: > if there's a change by test1, then a change by test2, then "not changed" by test3, you will rollback to test2 version instead of rolling back to original version.

Consider the case with a file `foo` that contains `42`.
 - `test1` updates `foo` with `10` and create `foo.backup` with `42`
 - `test2` will update `foo`. The delete action of `test1` restore `42` in `foo`.  The `create` action create `foo.backup` with `42` and update `foo`
 - test3 call delete action of `test2` and restores `foo` with `42`.

vboulineau: Yes okay with `DeleteBeforeReplace` it should work, however we're still dependent of whether the `command` provider considers that it should update the resource in-place or not. I'm guessing that an alteration of `createCommand` should trigger a new resource, but to be safe we should also set `ReplaceOnChanges`.

The code also needs to be commented because the Pulumi options are not in the same place as the code that requires it (which we should probably change as well).

ogaca-dd: Code updated.

vboulineau (approved):

ogaca-dd:
Closed in favor of https://github.com/DataDog/test-infra-definitions/pull/77

PR #68, opened 2023-04-05
Add an invoke task to create a basic vm

ogaca-dd:
What does this PR do?
---------------------

Add an invoke task to create a basic VM.
This is a first step to validate the solution. Other parameters for the VM and other scenarios are going to be added in a later PR. 

```
Usage: inv[oke] [--core-opts] create.vm [--options] [other tasks here ...]

Docstring:
  Create a new virtual machine on the cloud.

Options:
  -a STRING, --agent-version=STRING   The version of the Agent for example
                                      '7.42.0~rc.1-1' or '6.39.0'
  -i, --install-agent                 Install the Agent (default False).
  -o STRING, --os-type=STRING         The operating system. Possible values are
                                      ['Ubuntu', 'Windows', 'AmazonLinux', 'Debian',
                                      'RedHat', 'Suse', 'Fedora']. Default 'Ubuntu'
  -s STRING, --stack-name=STRING      An optional name for the stack. This parameter
                                      is useful when you need to create several
                                      environments. Note: 'invoke destroy' may not
                                      work properly
```


pducolin (approved):

tasks/deploy.py
```diff
@@ -0,0 +1,55 @@
...
+    if api_key is None or len(api_key) != 32:
+        raise invoke.Exit("Invalid API KEY. You must define the environment variable 'DD_API_KEY'. " +
+                            "If you don't want an agent installation add '--no-install-agent'.")
+    return api_key
+
+
+stack_name_prefix = "invoke-"
+
+
+def _get_stack_name(scenario_name):
```
vboulineau, line 53: I believe that's too much prefixing. I don't think we really care if the stack was spawned by `invoke` or `pulumi up` or whatever.
However we do care about being able to spawn 2 VMs (so giving name in input).

tasks/destroy.py
```diff
@@ -0,0 +1,51 @@
...
+        for stack_name in stacks:
+            error(" - " + stack_name)
+    else:
+        status = subprocess.call([
+            "aws-vault", "exec", "sandbox-account-admin", "--",
+            "pulumi", "destroy", "-s", stack
+        ])
+        if status == 0:
+            status = subprocess.call([
+                "pulumi", "stack", "rm", "-s", stack, "-y"
```
vboulineau, line 37: nit: Can be inlined in `destroy` code.

vboulineau (changes requested):
Structure LGTM, but implementation currently hardcodes a lot of things that are the added value of the CLI.

tasks/deploy.py
ogaca-dd: > I don't think we really care if the stack was spawned by invoke or pulumi up or whatever.

I think we care. As long as there is a single environment, calling `inv destroy` uses the same config as `pulumi up` and so the code path is the same (the environment is properly destroyed).

tasks/config.py
```diff
@@ -0,0 +1,70 @@
...
+    key_pairs: List[str] = []
+    output = output.decode("utf-8")
+    for line in output.splitlines():
+        parts = line.split(" ")
+        if len(parts) > 0:
+            key_pair_path = os.path.basename(parts[-1])
+            key_pair = os.path.splitext(key_pair_path)[0]
+            key_pairs.append(key_pair)
+
+    if key_pair_to_search not in key_pairs:
```
pducolin, line 65: **💭 thought**
As stated by @vboulineau, this will not work if we follow the new guidelines, using default `id_xxx` key that is uploaded as `<username>` on AWS.

tasks/deploy.py
```diff
@@ -0,0 +1,101 @@
...
+        raise invoke.Exit(
+            f"the os type '{os_type}' is not supported. Possibles values are {os_types}"
+        )
+    return os_type
+
+
+def _deploy(
+    ctx: Context, stack_name: Optional[str], config: Config, flags: Dict[str, Any]
+) -> None:
+    flags["ddinfra:aws/defaultKeyPairName"] = config.key_pair
```
pducolin, line 57: **💬 suggestion**
What about initialising `flags` using `config` in `deploy` and pass only `flags` to `_deploy` ? It is confusing at first sight to see `flags` set both here and in the caller

tasks/deploy.py
```diff
@@ -0,0 +1,101 @@
...
+    flags["ddinfra:env"] = "aws/sandbox"
+
+    if "ddagent:deploy" in flags and flags["ddagent:deploy"]:
+        flags["ddagent:apiKey"] = _get_api_key()
```
pducolin, line 61: **💭 thought**
We might not need this if we use the fakeintake, and we could set a random key or a `0`s key

pducolin (approved):
Nice ! Few comments, the one on the fakeintake can be addressed when we'll support the `fakeintake` option within vm scenario

tasks/config.py
```diff
@@ -0,0 +1,70 @@
...
+
+            defaultPublicKeyPath = _get_optional_value(
+                "ddinfra:aws/defaultPublicKeyPath", config_dict
+            )
+            c = Config()
+            c.key_pair = key_pair
+            c.defaultPublicKeyPath = defaultPublicKeyPath
+            return c
+    except FileNotFoundError:
+        raise invoke.Exit(f"Cannot find the configuration located at {config_path}")
```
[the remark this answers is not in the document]
ogaca-dd: I don't think so but we can provide an example in our documentation? What do you think?

tasks/config.py
```diff
@@ -0,0 +1,70 @@
...
+    config_filename = ".test_infra_config.yaml"
+    config_path = Path.home().joinpath(config_filename)
+    try:
+        with open(config_path) as f:
+            content = f.read()
+            config_dict = yaml.load(content, Loader=yaml.Loader)
+            key_pair: str = _get_value(
+                "ddinfra:aws/defaultKeyPairName", config_dict, config_path
+            )
+            check_key_pair: Optional[bool] = _get_optional_value(
```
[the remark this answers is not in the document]
ogaca-dd: > To simplify line 28

How to simplify the code?

tasks/config.py
ogaca-dd: Indeed, that is why this parameter is optional.

tasks/deploy.py
ogaca-dd: I think this change can be done when the invoke task support the fake intake.

tasks/deploy.py
ogaca-dd: I have improved these functions.

pducolin (approved):
Thank you for addressing the comments !

pducolin:
Can you add 

```
pip install --requirement requirements.txt
```

to the setup in the readme ?

tasks/deploy.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+    return cfg.defaultPublicKeyPath
+
+
+def _deploy(stack_name: Optional[str], flags: Dict[str, Any]) -> None:
+    cmd_args = ["aws-vault", "exec", "sandbox-account-admin", "--", "pulumi", "up"]
+    for key, value in flags.items():
+        if value is not None and value != "":
+            cmd_args.append("-c")
+            cmd_args.append(shlex.quote(f"{key}={value}"))
```
pducolin, line 71: **🔨 warning**
`shlex.quote` is preventing agent version to be properly passed to the pulumi cli, removing it works fine, what is this used for ?

ogaca-dd: Indeed, this is not needed as `subprocess.check_call` already handle space in the argument

amenasria:
Such a good feature 😍🚀 Can't wait !!

tasks/deploy.py
```diff
@@ -0,0 +1,104 @@
...
+            + "If you don't want an agent installation add '--no-install-agent'."
+        )
+    return api_key
+
+
+def get_stack_name_prefix() -> str:
+    return "invoke-"
+
+
+def _get_stack_name(stack_name: Optional[str], scenario_name: str) -> str:
```
vboulineau, line 100: `_get_stack_name` should not be in `deploy.py`.

tasks/deploy.py
vboulineau: Having a prefix does not change this, it's only use to restrict the ability of `inv destroy` to destroy stacks that were created before, which I don't understand why we artificially limit.
However it will fill all object names with `invoke-` everywhere, which, I believe, in some condensed cloud provider UI is just hurting readability.

aws/scenarios/vm/run.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/DataDog/test-infra-definitions/datadog/agent"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func Run(ctx *pulumi.Context) error {
+	env := config.NewCommonEnvironment(ctx)
+	var osType ec2os.Type
+
+	osTypeStr := env.InfraOSType()
```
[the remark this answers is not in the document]
ogaca-dd: Code updated to use case insensitive compare.

common/config/environment.go
```diff
@@ -22,6 +22,7 @@ const (
 	// Infra namespace
 	ddInfraEnvironment       = "env"
 	ddInfraKubernetesVersion = "kubernetesVersion"
+	ddInfraOSType            = "osType"
```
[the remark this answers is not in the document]
ogaca-dd: Code updated.

tasks/config.py
```diff
@@ -0,0 +1,70 @@
...
+import os
+import yaml
+from pathlib import Path
+from .tool import *
+from typing import List, Dict, Any, Optional
+
+
+class Config:
+    key_pair: str
+    defaultPublicKeyPath: Optional[str]
```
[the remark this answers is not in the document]
ogaca-dd: This code was replaced.

tasks/deploy.py
```diff
@@ -0,0 +1,104 @@
...
+default_public_path_key_name = "ddinfra:aws/defaultPublicKeyPath"
+
+
+def deploy(
+    _: Context,
+    scenario_name: str,
+    stack_name: Optional[str] = None,
+    install_agent: Optional[bool] = None,
+    agent_version: Optional[str] = None,
+    os_type: Optional[str] = None,
```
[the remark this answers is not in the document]
ogaca-dd: The responsibility of `deploy` is to map all variable names from all scenarios to the config map. 
If a scenario do not need `os_type` then `os_type` value is `None`. What would be the benefit of using kwargs?

tasks/deploy.py
```diff
@@ -0,0 +1,104 @@
...
+        raise invoke.Exit(f"Error when running {cmd_args}: {e}")
+
+
+def _get_root_path() -> str:
+    folder = pathlib.Path(__file__).parent.resolve()
+    return str(folder.parent)
+
+
+def _get_api_key() -> str:
+    api_key = os.getenv("DD_API_KEY")
```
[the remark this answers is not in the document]
ogaca-dd: Code updated.

tasks/deploy.py
ogaca-dd: Why not? This function is only used in this file.

ogaca-dd:
> It's probably not an error if it the file does not exist.

You don't

tasks/deploy.py
```diff
@@ -0,0 +1,104 @@
...
+def _default_public_path_key_name(cfg: Config, os_type: str) -> Optional[str]:
+    if os_type == "Windows" and cfg.defaultPublicKeyPath is None:
+        raise invoke.Exit(
+            f"You must set {default_public_path_key_name} when using this operating system"
+        )
+    return cfg.defaultPublicKeyPath
+
+
+def _deploy(stack_name: Optional[str], flags: Dict[str, Any]) -> None:
+    cmd_args = ["aws-vault", "exec", "sandbox-account-admin", "--", "pulumi", "up"]
```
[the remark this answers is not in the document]
vboulineau: Azure and Google requires login to be done using browser beforehand, so it won't be possible to wrap the command, it will be a pre-comamnd step.
You can also use a pre-defined service account, which is what I do for AWS and injecting `aws-vault` all the time will break that.

tasks/deploy.py
vboulineau: It should be used in `destroy` otherwise the commands are not symmetric.

tasks/deploy.py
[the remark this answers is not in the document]
vboulineau: I'd rather let the scenarios handle that as they should know the params they take and how to map them, putting only common params in `_deploy`.

pducolin (approved):

vboulineau (approved):

PR #71, opened 2023-04-14
[fakeintake] configure the agent to use fakeintake

pducolin:
What does this PR do?
---------------------

Add exporter for the fakeintake and properly configure the agent to use it as intake.

-------------------

None

Motivation
----------

Add helper to deploy a fakeintake and use it with VMs

Additional Notes
----------------

### DIFF
--- a/aws/ecs/fargateService.go
+++ b/aws/ecs/fargateService.go
@@ -6,7 +6,7 @@ import (
 
 	"github.com/DataDog/test-infra-definitions/aws"
 	"github.com/DataDog/test-infra-definitions/datadog/agent"
-	"github.com/DataDog/test-infra-definitions/datadog/fakeintake"
+	ddfakeintake "github.com/DataDog/test-infra-definitions/datadog/fakeintake"
 	"github.com/cenkalti/backoff/v4"
 
 	classicECS "github.com/pulumi/pulumi-aws/sdk/v5/go/aws/ecs"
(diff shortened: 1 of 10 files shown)


aws/ecs/fargateService.go
```diff
@@ -126,30 +126,31 @@ func getFirelensLogConfiguration(source, service, apiKeyParamName pulumi.StringI
...
-	fargateService, err := FargateService(e, e.Namer.ResourceName("fakeintake-srv"), pulumi.String(fakeintake.FargateClusterArn), taskDef.TaskDefinition.Arn())
-
+	fargateService, err := FargateService(e, e.Namer.ResourceName("fakeintake-srv"), pulumi.String(e.ECSFargateFakeintakeClusterArn()), taskDef.TaskDefinition.Arn())
 	// Hack passing taskDef.TaskDefinition.Arn() to execute apply function
 	// when taskDef has an ARN, thus it is defined on AWS side
 	ipAddress = pulumi.All(taskDef.TaskDefinition.Arn(), fargateService.Service.Name()).ApplyT(func(args []any) (string, error) {
 		var ipAddress string
 		err := backoff.Retry(func() error {
-			fmt.Println("waiting for task private ip")
+			fmt.Println("waiting for fakeintake task private ip")
```
ogaca-dd, line 143: Are these lines displayed when using pulumi API?

datadog/fakeintake/fakeintake.go
```diff
@@ -0,0 +1,58 @@
...
+		}
+		url, ok := outputs.Value.(string)
+		if !ok {
+			return nil, fmt.Errorf("the type %v is not valid for the key %v", reflect.TypeOf(outputs.Value), exporter.stackKey)
+		}
+		return &ClientData{Url: url}, nil
+	}
+}
+
+type PulumiExporter struct {
```
ogaca-dd, line 44: I know we discussed this topic in person but when looking at the code I think it would be easier to understand if:
 - `PulumiExporter` is defined in the same file as its constructor `NewEcsFakeintake`
 - Is there a reason to named the struct `PulumiExporter` and not something like `FakeIntake`?
 - Because there are two fields in `PulumiExporter` (`Url` and `stackKey`), I think having only one struct is enough instead of having 3 (`PulumiExporter`, `exporter` and `PulumiData`)

datadog/agent/install.go
```diff
@@ -68,11 +67,12 @@ func NewInstaller(vm vm.VM, options ...func(*Params) error) (*Installer, error)
...
 	for _, cmd := range serviceManager.RestartAgentCmd() {
 		restartAgentRes := commonNamer.ResourceName("restart-agent")
+		cmdPulumiStr := pulumi.String(cmd)
 		lastCommand, err = runner.Command(
 			restartAgentRes,
 			&command.Args{
-				Create:   pulumi.String(cmd),
-				Triggers: pulumi.Array{pulumi.String(utils.StrHash(cmd, configHash, integsHash))},
+				Create:   cmdPulumiStr,
+				Triggers: pulumi.Array{utils.PulumiStrHash(cmdPulumiStr, configHash, pulumi.String(integsHash))},
```
ogaca-dd, line 75: Can you explain why you need `PulumiStrHash`. Can't `pulumi.Array` contains directly `cmdPulumiStr`?

datadog/agent/params.go
```diff
@@ -94,7 +96,20 @@ func WithTelemetry() func(*Params) error {
...
 }
+
+func WithFakeintake(fakeintake fakeintake.PulumiData) func(*Params) error {
+	return func(p *Params) error {
+		// configure metrics and check run intake
+		extraConfig := pulumi.Sprintf(`dd_url: http://%s:80
+logs_config.logs_dd_url: %s:80
+logs_config.logs_no_ssl: true
+logs_config.force_use_http: true
+logs_enabled: true`, fakeintake.Url, fakeintake.Url)
```
ogaca-dd: `logs_enabled: true` we should let the user decider to enable it or not? What do you think?

aws/ecs/fargateService.go
pducolin: Yes, they are still logged

datadog/fakeintake/fakeintake.go
pducolin: Names can be improved, I wanted to split the three parts that are here:

* the exporter, that requires as stack key, and that know where to serialise and where to deserialise
* the data pulumi side, where `url` is a `pulumi.StringOutput`
* the data on the runtime side, where `url` is a `string`

For instance, the deserialise function does not need the pulumi data. This is why it is the `exporter` that has a `Deserialise` function.

On the other side the export action does not need to know about the runtime. I find having everything in the same struct confusing, but I agree names are not explicit enough.

datadog/agent/install.go
pducolin: I used an hash to be closer to the previous version, but actually as triggers is an Array of StringInput, passing them directly should be enough

ogaca-dd (approved):
LGTM

PR #73, opened 2023-04-18
Add RemoteServiceConnector

ogaca-dd:
What does this PR do?
---------------------
This PR adds `RemoteServiceConnector` which is a convenient struct to help implementing `GetClientDataDeserializer()`.
Here is an example of usage when you want to serialize a `Connection` and a `string`.
```go
type ClientData struct {
	     MyString   string
	     Connection utils.Connection
}

type MyRemoteService {
	     // This implement GetClientDataDeserializer() func(auto.UpResult) (*ClientData, error)
	     *RemoteServiceConnector[ClientData]
}

remoteServiceConnector := utils.NewRemoteServiceConnector(ctx, ClientData{})
remoteServiceConnector.Register(stackKey, "Connection", connection)
remoteServiceConnector.Register("test", "MyString", pulumi.String("Hello"))

return &MyRemoteService{
     MyRemoteService: remoteServiceConnector
}
```

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/utils/remoteServiceConnector.go
+++ b/common/utils/remoteServiceConnector.go
@@ -50,13 +50,13 @@ func NewRemoteServiceConnector[T any](ctx *pulumi.Context, initialValue T) *Remo
 // stackKeyName is the key in the stack name and should be unique
 // outputFieldName is the name of the field in the struct
 // value is the object to be saved in the stack
-func (s *RemoteServiceConnector[T]) Register(stackKeyName string, outputFieldName string, value pulumi.Input) {
-	s.ctx.Export(stackKeyName, value)
-	s.stackKeyFieldNameMap[stackKeyName] = outputFieldName
+func (c *RemoteServiceConnector[T]) Register(stackKeyName string, outputFieldName string, value pulumi.Input) {
+	c.ctx.Export(stackKeyName, value)
+	c.stackKeyFieldNameMap[stackKeyName] = outputFieldName
 }
 
-func (s *RemoteServiceConnector[T]) GetClientDataDeserializer() func(auto.UpResult) (*T, error) {
-	return s.deserializeFromUpResult
+func (c *RemoteServiceConnector[T]) GetClientDataDeserializer() func(auto.UpResult) (*T, error) {
+	return c.deserializeFromUpResult
 }
 
 func (c *RemoteServiceConnector[T]) deserializeFromUpResult(upResult auto.UpResult) (*T, error) {
@@ -80,7 +80,9 @@ func (c *RemoteServiceConnector[T]) deserializeFromUpResult(upResult auto.UpResu
 		}
 
 		if field.Kind() == reflect.Struct {
-			deserializeStruct(field, outputs)
+			if err := deserializeStruct(field, outputs); err != nil {
+				return nil, err
+			}
 		} else {
 			field.Set(reflect.ValueOf(outputs.Value))
 		}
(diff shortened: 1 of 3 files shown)


common/utils/remoteServiceConnector.go
```diff
@@ -0,0 +1,109 @@
...
+//	 }
+//
+//		remoteServiceConnector := utils.NewRemoteServiceConnector(ctx, ClientData{})
+//		remoteServiceConnector.Register(stackKey, "Connection", connection)
+//		remoteServiceConnector.Register("test", "MyString", pulumi.String("Hello"))
+//
+//		return &MyRemoteService{
+//		    MyRemoteService: remoteServiceConnector
+//		}
+type RemoteServiceConnector[T any] struct {
```
pducolin, line 32: **❓ question**
There is no way in golang to limit T to structs made of `pulumi.Input`like [type guards in TS](https://www.typescriptlang.org/docs/handbook/2/narrowing.html), right ?

common/utils/remoteServiceConnector.go
```diff
@@ -0,0 +1,109 @@
...
+//	     Connection utils.Connection
+//	 }
+//
+//	 type MyRemoteService {
+//	     // This implement GetClientDataDeserializer() func(auto.UpResult) (*ClientData, error)
+//	     *RemoteServiceConnector[ClientData]
+//	 }
+//
+//		remoteServiceConnector := utils.NewRemoteServiceConnector(ctx, ClientData{})
+//		remoteServiceConnector.Register(stackKey, "Connection", connection)
```
pducolin, line 26: **💭 thought**
Would be cool if `Register` was accepting a `pulumi.Input` version of `ClientData`, with a matching type matrix
```
string <-> pulumi.StringInput
int <-> pulumi.IntInput
utils.Connection <-> remote.Connection
```

common/utils/remoteServiceConnector.go
```diff
@@ -0,0 +1,109 @@
...
+		ctx:                  ctx,
+		value:                initialValue,
+		stackKeyFieldNameMap: make(map[string]string)}
+}
+
+// Register registers a field.
+// stackKeyName is the key in the stack name and should be unique
+// outputFieldName is the name of the field in the struct
+// value is the object to be saved in the stack
+func (s *RemoteServiceConnector[T]) Register(stackKeyName string, outputFieldName string, value pulumi.Input) {
```
pducolin, line 53: **💬 suggestion**
Could we add an explicit "Check all fields of type T are registered" to be called to ensure all fields required by the deserialiser were registered ?

pducolin (approved):
Great work ! This helps centralising registration / deserialisation. I have a suggestion for a further helper check, too bad we cannot narrow type of `T` to `ClientData`.

common/utils/remoteServiceConnector.go
ogaca-dd: I don't know how to do it :). Also this code will be used by a very limited number of users and a very limited number of times, so I think this is fine.

common/utils/remoteServiceConnector.go
ogaca-dd: The current implementation allows to pass an initial value for `T` in `NewRemoteServiceConnector`, so you can store non pulumi type like `int` for example.

pducolin (approved):

PR #74, opened 2023-04-19
[unix] properly format linux command with sudo or user

pducolin:
What does this PR do?
---------------------

Fix sudo command execution by properly formatting, handling separately `sudo` and `sudo -u`

-------------------

dockerVM, all

Motivation
----------

https://github.com/DataDog/test-infra-definitions/pull/44 broke docker vm scenario and docker commands, as it prefixed command with `sudo bash -c '<command>'` instead of `sudo <command>`

This ended up in running subshell commands as `root` user instead of default user as we were expecting

```
ubuntu@ip:~$ sudo  bash -c 'echo $(whoami)'
root
ubuntu@ip:~$ bash -c 'sudo echo $(whoami)'
ubuntu
```

Additional Notes
----------------

### DIFF
--- a/command/unixOSCommand.go
+++ b/command/unixOSCommand.go
@@ -59,24 +59,29 @@ func (fs unixOSCommand) GetTemporaryDirectory() string {
 }
 
 func (fs unixOSCommand) BuildCommandString(command pulumi.StringInput, env pulumi.StringMap, sudo bool, user string) pulumi.StringInput {
-	var prefix string
-
-	if sudo {
-		prefix = "sudo"
-	} else if user != "" {
-		prefix = fmt.Sprintf("sudo -u %s", user)
-	}
+	formattedCommand := formatCommandIfNeeded(command, sudo, user)
 
 	var envVars pulumi.StringArray
 	for varName, varValue := range env {
 		envVars = append(envVars, pulumi.Sprintf(`%s="%s"`, varName, varValue))
 	}
 
-	return buildCommandString(command, envVars, func(envVarsStr pulumi.StringOutput) pulumi.StringInput {
-		commandEscaped := command.ToStringOutput().ApplyT(func(command string) string {
-			return shellescape.Quote(command)
-		}).(pulumi.StringOutput)
-
-		return pulumi.Sprintf("%s %s bash -c %s", prefix, envVarsStr, commandEscaped)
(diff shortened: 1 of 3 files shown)


vboulineau (approved):

command/unixOSCommand.go
```diff
@@ -59,24 +59,29 @@ func (fs unixOSCommand) GetTemporaryDirectory() string {
...
-
-		return pulumi.Sprintf("%s %s bash -c %s", prefix, envVarsStr, commandEscaped)
+	return buildCommandString(formattedCommand, envVars, func(envVarsStr pulumi.StringOutput) pulumi.StringInput {
+		return pulumi.Sprintf("%s %s", envVarsStr, formattedCommand)
 	})
 }
+
+func formatCommandIfNeeded(command pulumi.StringInput, sudo bool, user string) pulumi.StringInput {
+	if !sudo && user == "" {
+		return command
```
ogaca-dd, line 76: When sudo is `true` or `user == ""`, the code doesn't use `bash -c`. Can it lead to a difference of behaviour when having `sudo` versus not having `sudo`?

pducolin: The difference is on how we expect to run subshell commands within `command`.

Using a specific `user` requires running all subcommands as that `user`, so we wrap the command and run it in a new shell as that user (`sudo -u <user> bash -c '<command>'`)

In other cases we expect the top command to run as `sudo`, and children as default user, like what you would do manually.

This was indeed breaking 

```golang
usermod, err := d.runner.Command(
  d.namer.ResourceName("usermod-g-docker"),
  &Args{
	Create: pulumi.String("usermod -a -G docker $(whoami)"),
	Sudo:   true,
  },
  pulumi.DependsOn([]pulumi.Resource{dockerInstall}))
```  

as we were expecting to run `usermod` as `root`, with `sudo`, and the subcommand `$(whoami)` as standard user. But it was instead running all as `root`

pducolin: Using both `sudo` and `user` should return a warning, adding it.

pducolin: As I will probably forget about this, I suggest changing the `usermod` command to use `whoami` command output executed in a separate step, like 
```
Create: pulumi.Sprintf("usermod -a -G docker %s", whoamiCmd),
```

@vboulineau what do you think ?

pducolin: Changed !

PR #84, opened 2023-05-02
Add default providers for resources

usamasaqib:
What does this PR do?
---------------------
This PR adds default providers for some common resources.

-------------------

Motivation
----------
Package level default providers are disabled in the automation API. The user must now explicitly set the provider for all resources.

Additional Notes
----------------

### DIFF
--- a/aws/scenarios/microVMs/microvms/domain.go
+++ b/aws/scenarios/microVMs/microvms/domain.go
@@ -8,8 +8,8 @@ import (
 	"github.com/DataDog/test-infra-definitions/aws/scenarios/microVMs/microvms/resources"
 	"github.com/DataDog/test-infra-definitions/aws/scenarios/microVMs/vmconfig"
 	"github.com/DataDog/test-infra-definitions/common/namer"
+	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/pulumi/pulumi-libvirt/sdk/go/libvirt"
-	"github.com/pulumi/pulumi-random/sdk/v4/go/random"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
@@ -38,10 +38,12 @@ func generateDomainIdentifier(vcpu, memory int, vmsetName, tag, arch string) str
 	return fmt.Sprintf("ddvm-%s-%s-%s-%d-%d", vmsetName, arch, tag, vcpu, memory)
 }
 func generateNewUnicastMac(ctx *pulumi.Context, domainID string) (pulumi.StringOutput, error) {
-	pulumiRandStr, err := random.NewRandomString(ctx, "random-"+domainID, &random.RandomStringArgs{
-		Length:  pulumi.Int(6),
-		Special: pulumi.Bool(true),
-	})
+	r, err := utils.NewRandom(ctx)
+	if err != nil {
+		return pulumi.StringOutput{}, err
+	}
+
+	pulumiRandStr, err := r.RandomString(domainID, 6, true)
 	if err != nil {
 		return pulumi.StringOutput{}, err
 	}
--- a/aws/scenarios/microVMs/microvms/run.go
+++ b/aws/scenarios/microVMs/microvms/run.go
@@ -149,7 +149,10 @@ func configureInstance(instance *Instance, m *config.DDMicroVMConfig) ([]pulumi.
 
 	env := *instance.e.CommonEnvironment
 	osCommand := command.NewUnixOSCommand()
-	localRunner := command.NewLocalRunner(env, osCommand)
+	localRunner, err := command.NewLocalRunner(env, osCommand)
+	if err != nil {
+		return waitFor, err
+	}
 	if instance.Arch != LocalVMSet {
 		remoteRunner, err := command.NewRunner(
 			env,
--- a/command/runner.go
+++ b/command/runner.go
@@ -5,6 +5,8 @@ import (
 
 	"github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/namer"
+	"github.com/pulumi/pulumi-command/sdk/go/command"
+	pulumiCommand "github.com/pulumi/pulumi-command/sdk/go/command"
 	"github.com/pulumi/pulumi-command/sdk/go/command/local"
 	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
@@ -42,11 +44,18 @@ type Runner struct {
 	waitCommand *remote.Command
 	config      runnerConfiguration
 	osCommand   OSCommand
+	provider    *command.Provider
 }
 
 func WithUser(user string) func(*Runner) {
 	return func(r *Runner) {
 		r.config.user = user
+	}
+}
+
+func WithProvider(provider *pulumiCommand.Provider) func(*Runner) {
+	return func(r *Runner) {
+		r.provider = provider
 	}
 }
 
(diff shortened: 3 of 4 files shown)


common/utils/random.go
```diff
@@ -0,0 +1,47 @@
+package utils
+
+import (
+	"github.com/DataDog/test-infra-definitions/common/namer"
+	"github.com/pulumi/pulumi-random/sdk/v4/go/random"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type Random struct {
```
pducolin, line 9: **❓ question**
Is this a `RandomProvider` or `RandomGenerator` ?

common/utils/random.go
```diff
@@ -0,0 +1,47 @@
...
+	provider *random.Provider
+}
+
+func WithProvider(provider *random.Provider) func(*RandomGenerator) {
+	return func(r *RandomGenerator) {
+		r.provider = provider
+	}
+}
+
+func NewRandom(ctx *pulumi.Context, options ...func(*RandomGenerator)) (*RandomGenerator, error) {
```
pducolin, line 21: **🥜 nitpick**
Struct constructor should match struct name - better replace from your IDE than apply this suggestion, it will break the linting

```suggestion
func NewRandomGenerator(ctx *pulumi.Context, options ...func(*RandomGenerator)) (*RandomGenerator, error) {
```

aws/scenarios/microVMs/microvms/run.go
```diff
@@ -149,7 +149,10 @@ func configureInstance(instance *Instance, m *config.DDMicroVMConfig) ([]pulumi.
 
 	env := *instance.e.CommonEnvironment
 	osCommand := command.NewUnixOSCommand()
-	localRunner := command.NewLocalRunner(env, osCommand)
+	localRunner, err := command.NewLocalRunner(env, osCommand)
+	if err != nil {
+		return waitFor, err
```
pducolin, line 154: **💬 suggestion**
`waitFor` is probably empty, following other error handlings you should return `nil, err
```suggestion
		return nil, err
```

aws/scenarios/microVMs/microvms/domain.go
```diff
@@ -38,10 +38,12 @@ func generateDomainIdentifier(vcpu, memory int, vmsetName, tag, arch string) str
...
-	pulumiRandStr, err := random.NewRandomString(ctx, "random-"+domainID, &random.RandomStringArgs{
-		Length:  pulumi.Int(6),
-		Special: pulumi.Bool(true),
-	})
+	r, err := utils.NewRandom(ctx)
+	if err != nil {
+		return pulumi.StringOutput{}, err
+	}
+
+	pulumiRandStr, err := r.RandomString(domainID, 6, true)
```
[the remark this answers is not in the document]
pducolin: What is the advantage of using pulumi random generator vs go random ?

command/runner.go
```diff
@@ -70,7 +78,15 @@ func NewRunner(
 		opt(runner)
 	}
 
+	// set default provider if none set
 	var err error
+	if runner.provider == nil {
+		runner.provider, err = pulumiCommand.NewProvider(e.Ctx, runner.namer.ResourceName("provider", connName), &pulumiCommand.ProviderArgs{})
```
pducolin, line 84: **❓ question**
How is the proper provider configured, for instance aws vs kubernetes ?

command/runner.go
```diff
@@ -86,14 +102,15 @@ func (r *Runner) Command(name string, args *Args, opts ...pulumi.ResourceOption)
 	if args.Sudo && r.config.user != "" {
 		r.e.Ctx.Log.Info(fmt.Sprintf("warning: running sudo command on a runner with user %s, discarding user", r.config.user), nil)
 	}
-	return remote.NewCommand(r.e.Ctx, r.namer.ResourceName("cmd", name), args.toRemoteCommandArgs(r.config, r.osCommand), opts...)
+	return remote.NewCommand(r.e.Ctx, r.namer.ResourceName("cmd", name), args.toRemoteCommandArgs(r.config, r.osCommand), append(opts, pulumi.Provider(r.provider))...)
```
pducolin, line 105: **❓ question**
Is this meant to add `r.provider` to the command dependencies ? If yes, I would rather explicitly create a resource dependencies array, and extend it with opts and r.provider

common/utils/random.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"github.com/DataDog/test-infra-definitions/common/namer"
+	"github.com/pulumi/pulumi-random/sdk/v4/go/random"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type RandomGenerator struct {
+	ctx      *pulumi.Context
+	namer    namer.Namer
+	provider *random.Provider
```
pducolin, line 12: **❓ question**
Where is this provider used ?

pducolin:
> Package level default providers are disabled in the automation API. The user must now explicitly set the provider for all resources.

Was this always true ? Is this a recent change in Automation API side ?

aws/scenarios/microVMs/microvms/domain.go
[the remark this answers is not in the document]
usamasaqib: A MAC address can be any 48-bit identifier. Some bytes have special meanings. In this case we only want to ensure that our mac address is unicast, which is done [here](https://github.com/DataDog/test-infra-definitions/blob/usama.saqib/default-providers/aws/scenarios/microVMs/microvms/domain.go#L54-L56C9) 

Go random would cause pulumi to detect that a mac address is changed on every update, and thus recreate the resource, and its dependencies. 
Pulumi random on the other hand does not create new mac addresses on every update, i.e. random resources are generated only when the stack is created. Therefore, there aren't any spurious resource updates.

command/runner.go
usamasaqib: `pulumi.Command` defines `ProviderArgs` as an empty struct. This is essentially a dummy provider.
The user can also define their own `Provider` if needed.

common/utils/random.go
usamasaqib: Ah yes. It should be added as a dependent resource in `NewRandomString`

usamasaqib:
> 

I have begun to run into this recently yes. Seems to have been added [here](https://github.com/DataDog/datadog-agent/pull/16173/files#diff-2dcd6655a4bcdec22348f7aa726f1df91b8473d2ec50953c844ee258dd5338f8R168)

brycekahle (approved):

aws/scenarios/microVMs/microvms/domain.go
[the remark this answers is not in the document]
pducolin: Ah nice, thank you for the clarification!

pducolin (approved):

pducolin:
Please wait for @vboulineau review before merging, as he has more knowledge over this code than me.

common/config/environment.go
```diff
@@ -166,6 +171,24 @@ func (e *CommonEnvironment) GetIntWithDefault(config *sdkconfig.Config, paramNam
 	return defaultValue
 }
 
+func (e *CommonEnvironment) CommandProvider(namer namer.Namer, name string) (*command.Provider, error) {
```
vboulineau, line 174: You don't need to take any parameter in, `Provider` are not currently named with namer for simplicity: https://github.com/DataDog/test-infra-definitions/blob/bf98888bd09fe6e3e760f49f6c0fb34e146ab36f/aws/environment.go#L73-L86

Also I don't think we should return an error on that, it will allow to do `Pulumi.Provider(e.CommandProvider())`

You can `panic` if you get an error there.

PR #91, opened 2023-05-05
[invoke] 🐛 check keypair only when required in config

pducolin:
What does this PR do?
---------------------

Go through keypair validation only when required in config

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -14,7 +14,6 @@ default_public_path_key_name = "ddinfra:aws/defaultPublicKeyPath"
 def deploy(
     _: Context,
     scenario_name: str,
-    key_pair_required: bool = False,
     public_key_required: bool = False,
     stack_name: Optional[str] = None,
     install_agent: Optional[bool] = None,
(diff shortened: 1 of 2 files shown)


tasks/vm.py
```diff
@@ -39,7 +39,6 @@ def create_vm(
     deploy(
         ctx,
         scenario_name,
-        key_pair_required=True,
```
ogaca-dd, line 42: This variable is set at `True` when a scenario need a key pair so we must keep it.

tasks/deploy.py
```diff
@@ -42,7 +41,7 @@ def deploy(
     if install_agent:
         flags["ddagent:apiKey"] = _get_api_key()
 
-    if key_pair_required:
+    if cfg.get_options().checkKeyPair:
```
ogaca-dd, line 44:
```suggestion
    if key_pair_required and cfg.get_options().checkKeyPair:
```

ogaca-dd (approved):
Thanks for the fix.

tasks/deploy.py
pducolin: Ooops, thank you !

PR #93, opened 2023-05-09
Simplify the deserialisation of remote service

ogaca-dd:
What does this PR do?
---------------------

Simplify the deserialisation of remote service. Change from `GetClientDataDeserializer() func(auto.UpResult) (*T, error)` to `Deserialize(upResult auto.UpResult) (*T, error)`


pducolin (approved):
I like it, makes sense !

PR #97, opened 2023-05-12
Add EKS invoke task

ogaca-dd:
What does this PR do?
---------------------

Add the invoke task `create-eks`.
Note: Installing the agent doesn't work.
```
Usage: inv[oke] [--core-opts] create-eks [--options] [other tasks here ...]

Docstring:
  Create a new EKS environment. It lasts around 20 minutes.

Options:
  --, --eks-bottlerocket-node-group   Install a bottlerocket node group (default False)
  -a STRING, --agent-version=STRING   The container version of the Agent for example '7.45.0-rc.3' (default `latest`)'
  -e, --[no-]eks-linux-node-group     Install a Linux node group (default True)
  -i, --install-agent                 Install the Agent (default False).
  -k, --eks-linux-arm-node-group      Install a Linux ARM node group (default False)
  -s STRING, --stack-name=STRING      An optional name for the stack. This parameter is useful when you need to create several environments.
  -w, --eks-windows-node-group        Install a Windows node group (default False)
```

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -4,7 +4,7 @@ import os
 import invoke
 import subprocess
 from invoke.context import Context
-from typing import List, Optional, Dict, Any
+from typing import Callable, List, Optional, Dict, Any
 import pathlib
 from . import tool
 
@@ -16,6 +16,7 @@ def deploy(
     scenario_name: str,
     key_pair_required: bool = False,
     public_key_required: bool = False,
+    app_key_required: bool = False,
     stack_name: Optional[str] = None,
     install_agent: Optional[bool] = None,
     agent_version: Optional[str] = None,
@@ -50,6 +51,10 @@ def deploy(
     for namespace in stackParams:
         for key, value in stackParams[namespace].items():
             flags[f"{namespace}:{key}"] = value
+
+
+    if app_key_required:
+        flags["ddagent:appKey"] = _get_app_key(cfg)
 
     return _deploy(stack_name, flags, debug)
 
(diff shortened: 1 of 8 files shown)


tasks/eks.py
```diff
@@ -0,0 +1,87 @@
...
+    _show_connection_message(full_stack_name)
+
+
+def _show_connection_message(full_stack_name: str):
+    outputs = tool.get_stack_json_outputs(full_stack_name)
+    kubeconfig = outputs["kubeconfig"]
+    kubeconfig_content = yaml.dump(kubeconfig)
+
+    f = os.open(
+        path="config.yaml", flags=(os.O_WRONLY | os.O_CREAT | os.O_TRUNC), mode=0o600
```
vboulineau, line 66: What happens with multiple stacks?
Ideally we want to merge newly created `kubeconfig` and `~/.kube/config`. This can be achieved in Python code or by using something like `KUBECONFIG=~/.kube/config:<new_file> kubectl config view --raw --merge --flatten`

tasks/eks.py
```diff
@@ -0,0 +1,87 @@
...
+        "eks_linux_node_group": doc.linux_node_group,
+        "eks_linux_arm_node_group": doc.linux_arm_node_group,
+        "eks_bottlerocket_node_group": doc.bottlerocket_node_group,
+        "eks_windows_node_group": doc.windows_node_group,
+    }
+)
+def create_eks(
+    ctx: Context,
+    stack_name: Optional[str] = None,
+    install_agent: Optional[bool] = False,
```
vboulineau, line 29: Not sure why we don't install the Agent by default?

ogaca-dd: Code updated.

tasks/deploy.py
```diff
@@ -100,17 +105,28 @@ def _get_root_path() -> str:
 
 
 def _get_api_key(cfg: Optional[Config]) -> str:
+    return _get_key("API KEY", cfg, lambda c: c.get_agent().apiKey, "E2E_API_KEY", 32)
+
```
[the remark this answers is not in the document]
ogaca-dd: Code updated.

tasks/eks.py
ogaca-dd: I have added the name of the stack in the configuration file name so multiple stacks can be used at the same time.

PR #99, opened 2023-05-15
Add some fake application workloads to help validating the agent

L3n41c:
What does this PR do?
---------------------

This PR adds resources to deploy fake applications that can be used to validate some agent features.
* The `nginx` deployment can be used to validate annotation-based auto-discovery.
* The `redis` deployment can be used to validate auto-conf (auto-discovery based on image name).
* The `nginx` and `redis` hpa can be used to validate the cluster-agent external metrics server.
* The `dogstatsd-ud[sp]` deployments can be used to validate origin detection with dogstatsd with both UDS and UDP sockes.
* The `prometheus` deployment can be used to validate Prometheus annotations based scraping.
* `kubernetes_state_core` check can be validated.

-------------------

* `aws/eks`. So far, the added test resources are usable on any Kubernetes implementation.

Motivation
----------

We need some workload to trigger the agent features we want to test.

Additional Notes
----------------


pducolin:
Wonder if this should rather be part of a Test definition in `datadog-agent`, as it is not strictly related to the setup required to deploy the agent.

vboulineau:
> Wonder if this should rather be part of a Test definition in `datadog-agent`, as it is not strictly related to the setup required to deploy the agent.

It makes it more difficult to use in manual tests, that's why we believe it's more convenient to have it here.

L3n41c:
> Wonder if this should rather be part of a Test definition in datadog-agent, as it is not strictly related to the setup required to deploy the agent.

The boundary between what goes in `datadog-agent` and what goes in `test-infra-definitions` might be a little bit blurry, but here are the reasons why I think that it’s valuable to put the “fake applications” used to validate the agent in this repository:

Those “fake applications” are not only useful for the automated CI tests of the `datadog-agent` repository.
I’d use them a lot for manual QA, as a starting point for building reproducers of support cases, as a test environment during dev, etc.
I very often need to spawn an environment, to deploy a dd agent and to create some activity inside the agent with some workload.
Spawning an agent which remains idle is mostly useless.
I think it would be more user friendly to be able to spawn a whole env (with the agent and the fake apps) from a single `pulumi` command from a single repository rather than having to go back and forth between `datadog-agent` and `test-infra-definitions`.

As said above, those fake applications are extremely valuable during the QA. So, having them inside a repository which is frozen during the QA isn’t super convenient.

Generally speaking, `datadog-agent` is already huge and those fake applications are not really part of the agent. They’re more related to the infrastructure needed to test it, which is what this repository seems to be meant for.

pducolin:
@L3n41c @vboulineau I understand the value of adding such helpers for manual QA, this is not limited to kube environments only: VMs and docker env can use extra apps to generate logs and metrics, `ddev` use lots of them for instance.

The idea we had was to keep only generic env helpers in test-infra and have feature specific helpers and code in `datadog-agent`. I also find it counter intuitive to have to go back and forth between repos, we can come back on our decision and have anything about environments here in `test-infra`. 

The big advantage here is that we have invoke tasks ready to use. The big disadvantage is that this can make `test-infra` grow big, with many run functions. 

How much do we want `test-infra` to know about specificities of environments ? If it's a component library for environments, it should not know about apps, if it's more than that, it can. 

From our decision log we noted "Any scenario meant for a specific agent test case should be implemented in datadog-agent. Scenarios in datadog-agent should have the smallest custom code as possible.Examples: snmp, process."

If there was an invoke task in `datadog-agent`, letting you spin up your generic env + apps infrastructure with one command, would it change your current perception of "test-infra is easier for this" ?

L3n41c:
Hello @pducolin,
Thank you for your feedback.

> If there was an invoke task in `datadog-agent`, letting you spin up your generic env + apps infrastructure with one command, would it change your current perception of "test-infra is easier for this" ?

Being able to spin up everything is only a tiny part of the iceberg.
In particular, having an `invoke` task inside the `datadog-agent` repository wouldn’t help for the following points:

* `datadog-agent` repository is frozen 2 weeks every 6 weeks. The “testing materials” won’t benefit from such a freeze. Also, we can expect to have more people wanting to update “testing materials” during the agent QA periods, which is when `datadog-agent` is frozen.

* `datadog-agent` CI takes more than 2 hours. AFAIK, there’s no mechanism in the CI to build and test only what has changed. This means that people working on “testing materials” will “suffer from the `datadog-agent` CI“ for no reason.

* People from [**#container-ecosystems**](https://github.com/orgs/DataDog/teams/container-ecosystems) team will want to leverage the testing framework to validate the public Helm chart hosted in the [`DataDog/helm-charts` repository](https://github.com/DataDog/helm-charts/) and the datadog operator hosted inside the [`DataDog/datadog-operator` repository](https://github.com/DataDog/datadog-operator). They won’t benefit from having the “testing materials” scattered between `test-infra-definitions` and `datadog-agent`.

* We regularly have to answer to internal security teams about vulnerabilities they identify inside the `datadog-agent` repository (both in dependencies and in the code itself with CodeQL). Whereas it’s good to have super high quality expectation for the datadog agent itself, we’d rather avoid loosing time explaining them that some pieces of code they identified as poorer quality are for “internal tooling” and don’t deserve the same quality expectation.

pducolin:
Hi @L3n41c ! Circling back on this subject, thank you for detailing pain points of having testing pieces in `datadog-agent`. Some friction  exposed in your analysis can be reduced, like CI time and smart test jobs running only changes, freeze periods with bypasses for internal tools. `datadog-agent` internal tools could be used by other repos, event tests defined there could be executed from other repos, having tasks configurable. Anyway I agree it would be better to have workloads here. 

Some peculiarities of workloads I see:

- it is defined once
- it is used in multiple tests
- it does not change frequently, should not be coupled with a specific feature for instance
- it is used across repositories - this is valid for tests too
- it is used for automated and manual tests

All of these points convinced me to have them defined here in `test-infra`. We're still discussing internally, will make time to ensure we find consensus by end of this week.

.github/workflows/docker-publish.yml
```diff
@@ -0,0 +1,112 @@
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        apps:
+          - dogstatsd
+          - http-client
+          - nginx-server
+          - prometheus
+          - redis-client
```
pducolin, line 49: What about building these in https://github.com/DataDog/docker-library instead ? Do you except these to change often ?

pducolin:
Hi @L3n41c ! Sorry for the late feedback, discussed within the squad, and agreed it makes sense to have custom workloads defined here in `test-infra`. 

I can help with merging master, we refactored to follow a cleaner code structure since https://github.com/DataDog/test-infra-definitions/pull/115/files#diff-de4064221a2d7f60121bae1312193c9e1da79e99259f88c3828afa5bc51480f5

I would move custom resources to `components/kubernetes/crds` or `components/kubernetes/customresources`

.github/workflows/docker-publish.yml
L3n41c: The content of the docker image is as important as the manifest (pulumi code to create the kubernetes pod or ECS task) in order to determine what the application will do and what the agent will do with that workload. And both are often strongly linked.

For example, an application that exposes a Prometheus endpoint needs to set the port in the app source code and in the k8s manifest.
The exact same port value needs to be set here: https://github.com/DataDog/test-infra-definitions/pull/99/files#diff-72c7db7015a1d7c8eecc2486e0bbb760f1fa19cd2fb1f0ab1c4182505efd6672R18 and here: https://github.com/DataDog/test-infra-definitions/pull/99/files#diff-2bf5b21c2f389ead885e1ef5b42a328316e7385401da6e38c559201c72a47eacR81
That’s why I’d rather keep the two close of each other.

Similarly, for an application that emits dogstatsd data, the application needs to know in its source code where is the dogstatsd endpoint and it is set thanks to the k8s downward API in the manifest.
The exact same environment variable name needs to be used here: https://github.com/DataDog/test-infra-definitions/pull/99/files#diff-013ee6d71017bc4436331edbb8d927d71b3ec92f9c1b2968da5f2b341458051fR25 and here https://github.com/DataDog/test-infra-definitions/pull/99/files#diff-a51f5b99d60fcbbd5161bd40b083d24e97da6416dce1f8b0eea3ffd90217dd35R67
Here again, I’d rather keep the two close of each other.

Finally, I’d like to avoid today’s situation where we have several repositories containing some kubernetes manifests that are referencing docker images for which we just have lost the source because they were pushed manually once.

This `docker-library` repository doesn’t seem to have any CI to build and automatically push images.

L3n41c:
Hello @pducolin,
Thank you very much for the feedback.
I’ve re-dispatched the files of this PR inside the new file structure. Let me know if some stuff could have a better location.

pducolin (approved):
Thank you for the huge valuable contribution and for taking into account my comments ! Looks good to me !

vboulineau (approved):

PR #101, opened 2023-05-15
 [env] 🐛  define provider per common environment instance

pducolin:
What does this PR do?
---------------------

Define one providers per environment instance

-------------------

All

Motivation
----------

Default singleton providers are causing a bug when updating the environment, as pulumi keeps re-using the same instance of provider while creating a new environment instance

Additional Notes
----------------

### DIFF
--- a/aws/environment.go
+++ b/aws/environment.go
@@ -61,7 +61,10 @@ type Environment struct {
 }
 
 func NewEnvironment(ctx *pulumi.Context) (Environment, error) {
-	commonEnv := config.NewCommonEnvironment(ctx)
+	commonEnv, err := config.NewCommonEnvironment(ctx)
+	if err != nil {
+		return Environment{}, err
+	}
 
 	env := Environment{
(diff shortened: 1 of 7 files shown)


usamasaqib (approved):
Thanks!

aws/scenarios/fakeintake/run.go
```diff
@@ -0,0 +1,29 @@
+package fakeintake
+
+import (
```
ogaca-dd, line 3: Note: This file doesn't seem to belong to this PR

ogaca-dd (approved):
LGTM

aws/scenarios/fakeintake/run.go
pducolin: ah good catch, thank you !

vboulineau (approved):

pducolin:
@ogaca-dd I forced push a reset of the history, removing the fakeintake run that was not supposed to make it to this PR

PR #103, opened 2023-05-16
Set the default value of checkKeyPair at true

ogaca-dd:
What does this PR do?
---------------------

Set the default value of `checkKeyPair` at `true` in `.test_infra_config.yaml.example`.

Motivation
----------

It is very hard for a new user to know that the ssh key has to be added in the ssh agent.

Additional Notes
----------------


ogaca-dd:
Close as default value is kept to False.

PR #104, opened 2023-05-16
[invoke] add setup script 

pducolin:
What does this PR do?
---------------------

Add setup invoke task to help onboarding 

-------------------

None

Motivation
----------

Reduce steps to setup a local environment. Tested on macOs, it will not work on Windows.

Additional Notes
----------------

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,28 @@ This repository contains IaC code based on Pulumi to provision dynamic test infr
 
 ## Quick start guide
 
-The first step is to download and install Pulumi CLI. Information can be found [here](https://www.pulumi.com/docs/get-started/install/).
+1. Clone this repository
 
-For instance, on MacOS:
+```bash
+cd ~/dd && git clone git@github.com:DataDog/test-infra-definitions.git 
 ```
-brew install pulumi/tap/pulumi
+
+1. Install Python dependencies
+
+```bash
+cd ~/dd/test-infra-definitions && pip install --requirement requirements.txt
+```
+
+1. Add a PULUMI_CONFIG_PASSPHRASE to your Terminal rc file
+
+```bash
+export PULUMI_CONFIG_PASSPHRASE=citest
+```
+
+1. Run and follow the setup script
+
+```bash
+inv setup
 ```
 
 ### Stack & Storage
--- a/tasks/config.py
+++ b/tasks/config.py
@@ -6,6 +6,8 @@ from typing import Dict, Optional
 from pydantic import BaseModel, Extra, ValidationError
 
 
+profile_filename = ".test_infra_config.yaml"
+
 class Config(BaseModel, extra=Extra.forbid):
     class Params(BaseModel, extra=Extra.forbid):
         class Aws(BaseModel, extra=Extra.forbid):
@@ -60,16 +62,22 @@ class Config(BaseModel, extra=Extra.forbid):
             return {}
         return self.stackParams
 
+    def save_to_local_config(self):
+        profile_path = Path.home().joinpath(profile_filename)
+        try:
+            with open(profile_path, 'w') as outfile:
+                yaml.dump(self.dict(), outfile)
+        except e:
+            raise invoke.Exit(f"Error saving config file {profile_path}: {e}")
+        info(f"Configuration file saved at {profile_path}")
 
 def get_local_config() -> Config:
-    profile_filename = ".test_infra_config.yaml"
     profile_path = Path.home().joinpath(profile_filename)
     try:
         with open(profile_path) as f:
             content = f.read()
             config_dict = yaml.load(content, Loader=yaml.Loader)
             return Config.parse_obj(config_dict)
-
     except FileNotFoundError:
         return Config.parse_obj({})
     except ValidationError as e:
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -1,0 +1,61 @@
+import os.path
+from pathlib import Path
+import getpass
+
+from invoke import task
+from invoke.context import Context
+
+from .config import Config, profile_filename
+from .tool import ask, debug, info, warn
+
+@task
+def setup(ctx: Context):
+    """
+    Setup a local environment interactively
+    """
+    info("🤖 Let's configure your environment for e2e tests! Press ctrl+c to stop me")
+    config = Config()
+    config.configParams = Config.Params()
+    config.configParams.agent = Config.Params.Agent()
+    config.configParams.aws = Config.Params.Aws()
+
+    config.configParams.aws.keyPairName = getpass.getuser()
+    keyPairName = ask(f"🔑 Key pair name - stored in AWS Sandbox, default [{config.configParams.aws.keyPairName}]: ")
+    if len(keyPairName) > 0:
+        config.configParams.aws.keyPairName = keyPairName
+
+    checkKeyPair = ask("Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent [Y/N]? Default No: ")
+    if len(checkKeyPair) > 0:
+        config.options = Config.Options()
+        config.options.checkKeyPair = checkKeyPair.lower() == "y" or checkKeyPair.lower() == "yes"
+
+    config.configParams.aws.publicKeyPath = Path.home().joinpath(".ssh", "id_ed25519.pub")
+    publicKeyPath = ask(f"🔑 Path to your public ssh key: default [{config.configParams.aws.publicKeyPath}]")
+    if len(publicKeyPath) > 0:
+        config.configParams.aws.publicKeyPath = publicKeyPath
+    while not os.path.isfile(config.configParams.aws.publicKeyPath):
+        warn(f"{config.configParams.aws.publicKeyPath} is not a falid ssh key")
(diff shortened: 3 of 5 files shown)


tasks/setup.py
```diff
@@ -0,0 +1,69 @@
...
+    config.configParams = Config.Params()
+    config.configParams.agent = Config.Params.Agent()
+    config.configParams.aws = Config.Params.Aws()
+
+    config.configParams.aws.keyPairName = getpass.getuser()
+    keyPairName = ask(f"🔑 Key pair name - stored in AWS Sandbox, default [{config.configParams.aws.keyPairName}]: ")
+    if len(keyPairName) > 0:
+        config.configParams.aws.keyPairName = keyPairName
+
+    checkKeyPair = ask("Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent [Y/N]? Default No: ")
```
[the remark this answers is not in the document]
pducolin: Will rephrase this

README.md
```diff
@@ -4,11 +4,28 @@ This repository contains IaC code based on Pulumi to provision dynamic test infr
...
-For instance, on MacOS:
+```bash
+cd ~/dd && git clone git@github.com:DataDog/test-infra-definitions.git 
 ```
-brew install pulumi/tap/pulumi
+
+1. Install Python dependencies
+
+```bash
+cd ~/dd/test-infra-definitions && pip install --requirement requirements.txt
```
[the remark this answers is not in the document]
pducolin: Need to check if it is part of the generic setup or the agent dev setup. If that is the case, will redirect to that as a prerequisite.

README.md
```diff
@@ -4,11 +4,28 @@ This repository contains IaC code based on Pulumi to provision dynamic test infr
...
+```
+
+1. Add a PULUMI_CONFIG_PASSPHRASE to your Terminal rc file
+
+```bash
+export PULUMI_CONFIG_PASSPHRASE=citest
```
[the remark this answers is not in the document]
pducolin: Will change this

README.md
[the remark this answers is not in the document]
pducolin: Adding prerequisites

README.md
```diff
@@ -2,13 +2,37 @@
...
+
+```bash
+cd ~/dd && git clone git@github.com:DataDog/test-infra-definitions.git 
+```
+
+1. Install Python dependencies
 
-For instance, on MacOS:
+```bash
+cd ~/dd/test-infra-definitions && pip install --requirement requirements.txt
```
vboulineau, line 23:
```suggestion
cd ~/dd/test-infra-definitions && pip3 install --requirement requirements.txt
```

vboulineau (approved):

README.md
```diff
@@ -2,13 +2,37 @@
...
+1. Add a PULUMI_CONFIG_PASSPHRASE to your Terminal rc file. Create a random password using 1Password and store it there
+
+```bash
+export PULUMI_CONFIG_PASSPHRASE=<random password stored in 1Password>
+```
+
+1. Run and follow the setup script
+
+```bash
+inv setup
```
ogaca-dd, line 35: We should add a note that it works only for MacOS and provides manual steps for Linux / Windows (In other page to keep this one short?). What do you think?

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+    keyPairName = ask(f"🔑 Key pair name - stored in AWS Sandbox, default [{config.configParams.aws.keyPairName}]: ")
+    if len(keyPairName) > 0:
+        config.configParams.aws.keyPairName = keyPairName
+
+    checkKeyPair = ask("Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent when creating manual environments or running e2e tests [Y/N]? Default No: ")
+    if len(checkKeyPair) > 0:
+        config.options = Config.Options()
+        config.options.checkKeyPair = checkKeyPair.lower() == "y" or checkKeyPair.lower() == "yes"
+
+    config.configParams.aws.publicKeyPath = Path.home().joinpath(".ssh", "id_ed25519.pub")
```
ogaca-dd, line 37: Is `id_ed25519.pub` really the default public key name?

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+    publicKeyPath = ask(f"🔑 Path to your public ssh key: default [{config.configParams.aws.publicKeyPath}]")
+    if len(publicKeyPath) > 0:
+        config.configParams.aws.publicKeyPath = publicKeyPath
+    while not os.path.isfile(config.configParams.aws.publicKeyPath):
+        warn(f"{config.configParams.aws.publicKeyPath} is not a falid ssh key")
+        config.configParams.aws.publicKeyPath = ask("Path to your public ssh key: ")
```
ogaca-dd, line 43: Is it expected that the message is not exactly the same as line 38?

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+    
+    config.configParams.agent.apiKey = "00000000000000000000000000000000"
+    apiKey = ask("🐶 Datadog API key - default [00000000000000000000000000000000]: ")
```
ogaca-dd, line 46: I don't think we should provide a default API Key as `00000000000000000000000000000000` is not valid. What do you think?

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+    if len(apiKey) == 32:
```
ogaca-dd, line 47: You may probably want to retry if the length of the key is not 32 as it can be a mistake.

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+        config.configParams.agent.apiKey = apiKey
+    
+    config.configParams.agent.appKey = "0000000000000000000000000000000000000000"
+    appKey = ask("🐶 Datadog APP key - default [0000000000000000000000000000000000000000]: ")
+    if len(appKey) == 40:
```
ogaca-dd, line 52: Same comment as for API key

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+        config.configParams.agent.appKey = appKey
+    
+    config.save_to_local_config()
+    profile_path = Path.home().joinpath(profile_filename)
+    debug("=========================================")
+    with open(profile_path, 'r') as f:
+        debug(f.read())
```
ogaca-dd, line 59: I am not sure we want to write the API key and APP key on the console. Maybe we can provide the link to the file created and add a message to check its content? What do you think?

tasks/config.py
```diff
@@ -60,16 +62,22 @@ def get_stack_params(self) -> Dict[str, Dict[str,str]]:
             return {}
         return self.stackParams
 
+    def save_to_local_config(self):
+        profile_path = Path.home().joinpath(profile_filename)
```
ogaca-dd, line 66: Note: I would create a function that get the full path.

tasks/setup.py
```diff
@@ -0,0 +1,66 @@
...
+    if len(checkKeyPair) > 0:
+        config.options = Config.Options()
+        config.options.checkKeyPair = checkKeyPair.lower() == "y" or checkKeyPair.lower() == "yes"
+
+    config.configParams.aws.publicKeyPath = Path.home().joinpath(".ssh", "id_ed25519.pub")
+    publicKeyPath = ask(f"🔑 Path to your public ssh key: default [{config.configParams.aws.publicKeyPath}]")
+    if len(publicKeyPath) > 0:
+        config.configParams.aws.publicKeyPath = publicKeyPath
+    while not os.path.isfile(config.configParams.aws.publicKeyPath):
+        warn(f"{config.configParams.aws.publicKeyPath} is not a falid ssh key")
```
ogaca-dd, line 42:
```suggestion
        warn(f"{config.configParams.aws.publicKeyPath} is not a valid ssh key")
```

ogaca-dd (commented):
Great PR!

README.md
pducolin: Will add a note as per this is tested only on macos.

tasks/setup.py
pducolin: I will change this to copy to clipboard the `cat` command

tasks/setup.py
pducolin: Changing this

tasks/setup.py
pducolin: This is a chicken and egg issue. With fakeintake there is no need for an api key, but our current VM script requires an API key. Plans is to skip API key requirement when using fakeintake, that is the default, but we need to change the script. So, in the meantime, I suggest configuring a fake API key - we still need an API key as the agent will expect to send one to fakeintake, but fakeintake will validate anything.

tasks/setup.py
vboulineau: To enhance compatibility you should probably fallback to `id_rsa.pub` for people (majority) that onboarded before this change.

tasks/setup.py
ogaca-dd: Indeed, I forgot the fact that fake intake doesn't need a real API KEY.

ogaca-dd (approved):
LGTM

PR #106, opened 2023-05-16
Add a way to specify the version of the cluster agent

L3n41c:
What does this PR do?
---------------------

Add a way to specify the version of the cluster agent.
Align the version of the cluster checks runners with the version of the node agents.

-------------------

Cluster agent and cluster checks runners are specific to Kubernetes setups.

Motivation
----------

We need a way to specify the version of the cluster agent to use.
We could enforce to use the same version for the cluster agent and the agent but in any case, they are in two different docker images.

Additional Notes
----------------


datadog/agent/docker.go
```diff
@@ -44,8 +46,36 @@ func DockerImageTag(e *config.CommonEnvironment) string {
...
 		agentImageTag = agentVersion.String()
 	} else {
-		e.Ctx.Log.Debug("Unable to parse Agent version, using latest", nil)
+		e.Ctx.Log.Debug("Unable to parse agent version, using latest", nil)
 	}
 
 	return agentImageTag
 }
+
+func DockerClusterAgentFullImagePath(e *config.CommonEnvironment, repositoryPath string) string {
```
vboulineau, line 55: We can probably have a single function by passing default values in

common/config/environment.go
```diff
@@ -27,11 +27,13 @@ const (
 	DDInfraOSFamily          = "osFamily"
 
 	// Agent Namespace
-	DDAgentDeployParamName        = "deploy"
-	DDAgentVersionParamName       = "version"
-	DDAgentFullImagePathParamName = "fullImagePath"
-	DDAgentAPIKeyParamName        = "apiKey"
-	DDAgentAPPKeyParamName        = "appKey"
+	DDAgentDeployParamName               = "deploy"
+	DDAgentVersionParamName              = "agentVersion"
```
vboulineau, line 31: We probably don't need to rename, we have `ddagent:version`, having `ddagent:agentVersion` is a bit redundant.

common/config/helpers.go
```diff
@@ -24,11 +24,19 @@ func SetConfigDefaultValue(config auto.ConfigMap, key, value string) {
 	}
 }
 
-func AgentSemverVersion(e *CommonEnvironment) (*semver.Version, error) {
-	version := e.AgentVersion()
+func anyAgentSemverVersion(anyAgentVersion func() string) (*semver.Version, error) {
```
ogaca-dd, line 27: What is the reason to pass a `func() string` and not the version directly?

ogaca-dd (approved):
LGTM

PR #121, opened 2023-05-25
Add an OS independent way to run Agent command

ogaca-dd:
What does this PR do?
---------------------

Add an OS independent way to run Agent command.

### DIFF
--- a/components/os/unix.go
+++ b/components/os/unix.go
@@ -29,6 +29,10 @@ func (*Unix) GetType() Type {
 	return UnixType
 }
 
+func (*Unix) GetRunAgentCmd(parameters string) string {
+	return "sudo datadog-agent " + parameters
+}
+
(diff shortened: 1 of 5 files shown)


components/os/unix.go
```diff
@@ -29,6 +29,10 @@ func (*Unix) GetType() Type {
 	return UnixType
 }
 
+func (*Unix) GetRunAgentCmd(parameters string) string {
+	return "sudo datadog-agent " + parameters
```
pducolin, line 33: **🔨 warning**
Should we have sudo and non sudo linux commands ? This one might not work on debian, out of the box

ogaca-dd: It works on debian.

pducolin: Ah ! I remember seeing a difference among Linux OS in kitchen tests, let's merge it as is and iterate later if needed, great PR !

pducolin (approved):

PR #126, opened 2023-05-31
Make the Helm agent deployment also deploy it on Windows nodes

L3n41c:
What does this PR do?
---------------------

Make the Helm datadog agent deployment method also deploy the agent on Windows nodes.

-------------------

* `aws/eks`

Motivation
----------

We want to be able to validate the Windows support.

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/helm.go
+++ b/components/datadog/agent/helm.go
@@ -24,11 +24,28 @@ type HelmInstallationArgs struct {
 	ValuesYAML   pulumi.AssetOrArchiveArrayInput
 }
 
-func NewHelmInstallation(e config.CommonEnvironment, args HelmInstallationArgs, opts ...pulumi.ResourceOption) (*kubeHelm.Release, error) {
+type HelmComponent struct {
+	pulumi.ResourceState
+
+	LinuxHelmReleaseName   pulumi.StringPtrOutput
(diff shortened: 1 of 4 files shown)


components/datadog/agent/helm.go
```diff
@@ -55,25 +74,56 @@ func NewHelmInstallation(e config.CommonEnvironment, args HelmInstallationArgs,
...
+		Values:      buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAgentImagePath, clusterAgentImageTag),
+	}, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	helmComponent.LinuxHelmReleaseName = linux.Name
+	helmComponent.LinuxHelmReleaseStatus = linux.Status
+
+	windows, err := helm.NewInstallation(e, helm.InstallArgs{
```
vboulineau, line 101: Should we make it conditional based on Args when there are no Windows nodes it will create objects for nothing.

L3n41c: @vboulineau This change https://github.com/DataDog/test-infra-definitions/pull/126/commits/8b78a6e870013e6d3af0ba45f21425fdc3c06ead should ensure we deploy Windows agents only on clusters where there are Windows nodes.

PR #127, opened 2023-06-01
Amit.slavin/support all agent version

amitslavin:
What does this PR do?
---------------------

Add support for custom agent image to unix os for the invoke create-vm command

-------------------

Motivation
----------

Additional Notes
----------------
In order to run the command you only need to know you pipeline id from gitlab.
Example:
`invoke create-vm -i -p=16543115 -o=ubuntu`


components/os/unix.go
```diff
@@ -33,6 +37,10 @@ func (*Unix) GetRunAgentCmd(parameters string) string {
 	return "sudo datadog-agent " + parameters
 }
 
+func getAgentMajorVersion(version AgentVersion) string {
+	return string(version.RepoBranch[len(version.RepoBranch)-1])
```
pducolin, line 41: **🔨 warning**
This function works only if `version.RepoBranch` is not empty. It is also used when version is custom. What if you parsed the major version once in `parseVersion` in `params.go`, get rid of this function and use `agent.MajorVersion` in `GetInstallCmd` ?

components/os/os.go
```diff
@@ -19,9 +19,11 @@ const (
...
 
 type AgentVersion struct {
-	Major       string
-	Minor       string // Empty means latest
-	BetaChannel bool
+	Major         string
+	Minor         string // Empty means latest
+	BetaChannel   bool
+	RepoBranch    string
+	IsCustomImage bool
```
pducolin, line 26: **💬 suggestion**
These 2 options are valid only together. They could merge to a unique property `PipelineId`, and parser would go like

 ```go
version, err := parseVersion("pipeline-16362517-a7")
assert.NoError(t, err)
assert.Equal(t, version, os.AgentVersion{
  PipelineId:    "16362517",
  Major: "7",
  Minor: "",
  BetaChannel: false
})
```

**Review by pducolin — COMMENTED** [review]
Nice !!! Couple of comments, the one on `MajorVersion` should be handled before merge

File: components/os/os.go (hunk 2 above)
Comment (pducolin) [reply components/os/os.go:26]: Not related to this PR, I wonder why `AgentVersion` is here, it should belong to agent packages

File: components/os/unix.go [hunk 3]
```diff
@@ -45,6 +53,14 @@ func getDefaultInstanceType(env config.Environment, arch Architecture) string {
 }
 
 func getUnixInstallFormatString(scriptName string, version AgentVersion) string {
```
bkabrda, line 55: I think it would be great to do some refactoring here to make this function accept distro type as an argument. It should be possible to also use this on RPM-based platforms (either RH-based or SUSE-based) with `TESTING_YUM_URL` and `TESTING_YUM_VERSION_PATH`.

components/os/unix.go
```diff
@@ -45,6 +53,14 @@ func getDefaultInstanceType(env config.Environment, arch Architecture) string {
...
+	if version.IsCustomImage {
+		commandLine := fmt.Sprintf(`TESTING_APT_URL=apttesting.datad0g.com TESTING_APT_REPO_VERSION="%v %v" `, version.RepoBranch, getAgentMajorVersion(version))
```
pducolin, line 57: **💬 suggestion**
Following @bkabrda hint, this suggestion would make the script work with all linux versions 

```suggestion
		testEnvVars := []string{}
		testEnvVars = append(testEnvVars, "TESTING_APT_URL=apttesting.datad0g.com")
		testEnvVars = append(testEnvVars, fmt.Sprintf(`TESTING_APT_REPO_VERSION="%v %v"`, version.RepoBranch, getAgentMajorVersion(version)))
		testEnvVars = append(testEnvVars, "TESTING_YUM_URL=yumtesting.datad0g.com")
		testEnvVars = append(testEnvVars, fmt.Sprintf("TESTING_YUM_VERSION_PATH=testing/%v/%v", version.RepoBranch, getAgentMajorVersion(version)))
		commandLine := strings.Join(testEnvVars, " ")
```

pducolin:
I wonder if we should rather pass `--pipeline=xxxxxx` instead of `pipeline-xxxxxx-a7`. We'll use this mostly for agent 7, when opening a pipeline we get the id, it seems more intuitive to use the `id` directly instead of `pipeline-xxxxx-a7`

Side note: once merged we could have the test command printed as a comment in a PR on github, something like

"To test this version run `inv create-vm --pipeline=xxxxxxx` from test-infra"

components/os/unix.go
pducolin: ⚠️ my suggestion above would break the code, you'll need to add import of `strings`, sorry for this

components/os/unix.go
amitslavin: mm not sure I fully understood this part, this function will be called only when the IsCustomImage will be true, which will make sure that the repo branch is not empty. In the start I do added a variable that is holding only the major version, but then I have noticed that for example for pipeline-16370943-a7 it will include only the major - 7, I am not sure how useful it will be to test major 6. But lets discuss about it on private, if you think that the agent version should always be 7 I may remove this part and validate in early stages that it exists.

components/os/os.go
amitslavin: agreed, removed the redundant variable and changed the name as you suggested

amitslavin:
> I wonder if we should rather pass `--pipeline=xxxxxx` instead of `pipeline-xxxxxx-a7`. We'll use this mostly for agent 7, when opening a pipeline we get the id, it seems more intuitive to use the `id` directly instead of `pipeline-xxxxx-a7`
> 
> Side note: once merged we could have the test command printed as a comment in a PR on github, something like
> 
> "To test this version run `inv create-vm --pipeline=xxxxxxx` from test-infra"

I do agree that it sounds better with the --pipeline, I just started with the -a variable because it was related to the agent as the support I have added, but if you support this option, I do think it is cleaner.

components/os/unix.go
amitslavin: changed and added the needed import, thanks :)

amitslavin:
refactor the code after the changes we talked about, @pducolin please take a look :)

pducolin (approved):
Thank you for addressing my comments! LGTM, please wait for @ogaca-dd review as he has more context on the invoke task

components/datadog/agent/params.go
```diff
@@ -66,8 +70,18 @@ func WithVersion(version string) func(*Params) error {
 	}
 }
 
+// WithPipelineID use a specific version of the Agent by pipeline id. For example: `16497585` returns `pipeline-16497585`
```
ogaca-dd, line 73:
```suggestion
// WithPipelineID use a specific version of the Agent by pipeline id. For example: `16497585` uses the version `pipeline-16497585`
```

components/os/unix.go
```diff
@@ -40,11 +44,29 @@ func getDefaultInstanceType(env config.Environment, arch Architecture) string {
...
 	default:
-		panic("Architecture not supportede")
+		panic("Architecture not supported")
 	}
 }
 
 func getUnixInstallFormatString(scriptName string, version AgentVersion) string {
+	if version.PipelineID != "" {
+		testEnvVars := []string{}
+		testEnvVars = append(testEnvVars, "TESTING_APT_URL=apttesting.datad0g.com")
```
ogaca-dd, line 54: Note: You can reduce the code size by initialising the array directly
```
testEnvVars := []string{
     "TESTING_APT_URL=apttesting.datad0g.com",
    "fmt.Sprintf(`TESTING_APT_REPO_VERSION="%v-a7 7"`, version.PipelineID)",
.....
}

ogaca-dd (approved):
LGTM

components/os/unix.go
amitslavin: I do like the way it is, because it makes the docs more clear for meaning of each line, so I may reduce a bit, but in my opinion it will look less clear what is going on

PR #133, opened 2023-06-06
Add fakeintake for the aws/eks scenario

L3n41c:
What does this PR do?
---------------------

Add the possibility to deploy a fakeintake (on ECS) for the `aws/eks` scenario.

-------------------

* `aws/eks`

Motivation
----------

We need the agent to ship data to the fakeintake in order to be able to assert metrics and logs.

Additional Notes
----------------


components/datadog/agent/helm.go
```diff
@@ -237,3 +249,24 @@ func buildWindowsHelmValues(installName string, agentImagePath, agentImageTag, _
 		},
 	}
 }
+
+func configureFakeintake(values pulumi.Map, fakeintake *ddfakeintake.ConnectionExporter) pulumi.MapOutput {
+	return fakeintake.IPAddress.ToStringOutput().ApplyT(func(ipAddress string) pulumi.Map {
+		values["agents"].(pulumi.Map)["useConfigMap"] = pulumi.Bool(true)
+		values["agents"].(pulumi.Map)["customAgentConfig"] = pulumi.Map{
+			"additional_endpoints": pulumi.Map{
+				fmt.Sprintf("http://%s", ipAddress): pulumi.String("FAKE"),
```
vboulineau, line 258:
```suggestion
				fmt.Sprintf("http://%s", ipAddress): pulumi.String("FAKEAPIKEY"),
```

components/datadog/fakeintake/connectionExporter.go
```diff
@@ -9,37 +9,37 @@ import (
...
 const (
-	stackKey = "fakeintake-url"
+	stackKey = "fakeintake-ip-address"
 )
 
 // ConnectionExporter contains pulumi side data and the export key
 type ConnectionExporter struct {
-	URL      pulumi.StringInput
-	stackKey string
+	IPAddress pulumi.StringInput
```
vboulineau, line 17: I'd say `host`, it could be a DNS name.

L3n41c: Agreed !
I’ve renamed `IPAddress` into `Host` in https://github.com/DataDog/test-infra-definitions/pull/133/commits/30b5719537cfb2b960ccdc81bdd2b7f0ce971ff8.

resources/aws/environmentDefaults.go
```diff
@@ -76,7 +76,7 @@ func sandboxDefault() environmentDefault {
 		ddInfra: ddInfra{
 			defaultVPCID:               "vpc-d1aac1a8",
 			defaultSubnets:             []string{"subnet-b89e00e2", "subnet-8ee8b1c6", "subnet-3f5db45b"},
-			defaultSecurityGroups:      []string{"sg-46506837", "sg-7fedd80a"},
+			defaultSecurityGroups:      []string{"sg-46506837", "sg-7fedd80a", "sg-019a8e84384954381"},
```
vboulineau, line 79: You need the same in other tenants.

components/datadog/agent/helm.go
```diff
@@ -243,3 +255,22 @@ func buildWindowsHelmValues(installName string, agentImagePath, agentImageTag, _
...
+		pulumi.Map{
+			"name":  pulumi.String("DD_LOGS_CONFIG_ADDITIONAL_ENDPOINTS"),
+			"value": pulumi.Sprintf("[{\"host\": \"%s\", \"port\": 80, \"is_reliable\": true, \"usessl\": false}]", fakeintake.Host),
+		},
+	}
+
+	if values["datadog"].(pulumi.Map)["env"] == nil {
+		values["datadog"].(pulumi.Map)["env"] = additionalEndpointsEnvVar
+	} else {
+		values["datadog"].(pulumi.Map)["env"] = append(values["datadog"].(pulumi.Map)["env"].(pulumi.MapArray), additionalEndpointsEnvVar...)
```
ogaca-dd, line 274: Is there any chance that `values["datadog"].(pulumi.Map)["env"]` failed? If yes you should handle it.

scenarios/aws/eks/fakeintake.go
```diff
@@ -0,0 +1,16 @@
+package eks
+
+import (
+	ddfakeintake "github.com/DataDog/test-infra-definitions/components/datadog/fakeintake"
+	"github.com/DataDog/test-infra-definitions/resources/aws"
+	"github.com/DataDog/test-infra-definitions/resources/aws/ecs"
+)
+
+func NewEcsFakeintake(env aws.Environment) (*ddfakeintake.ConnectionExporter, error) {
```
ogaca-dd, line 9: Can you not export this function, ie rename to `NewEcsFakeintake` because e2e user can use it instead of `ecs.NewEcsFakeintake`

components/datadog/agent/helm.go
```diff
@@ -243,3 +255,22 @@ func buildWindowsHelmValues(installName string, agentImagePath, agentImageTag, _
 		},
 	}
 }
+
+func configureFakeintake(values pulumi.Map, fakeintake *ddfakeintake.ConnectionExporter) {
+	additionalEndpointsEnvVar := pulumi.MapArray{
+		pulumi.Map{
+			"name":  pulumi.String("DD_ADDITIONAL_ENDPOINTS"),
+			"value": pulumi.Sprintf("{\"http://%s\": [\"FAKEAPIKEY\"]}", fakeintake.Host),
```
ogaca-dd, line 263: You can use  backquote to avoid escaping `"`

ogaca-dd (approved):
LGTM, can you at least unexport NewEcsFakeintake?

components/datadog/fakeintake/connectionExporter.go
```diff
@@ -9,37 +9,37 @@ import (
 )
 
 const (
-	stackKey = "fakeintake-url"
+	stackKey = "fakeintake-ip-address"
```
pducolin, line 12: **❓ question**
Should this be `fakeintake-host` ? If I remember well, the exported value is the ip address plus the port

components/datadog/fakeintake/connectionExporter.go
```diff
@@ -9,37 +9,37 @@ import (
...
 type ConnectionExporter struct {
-	URL      pulumi.StringInput
+	Host     pulumi.StringInput
 	stackKey string
 }
 
 // ClientData client side data
 type ClientData struct {
-	URL string
+	Host string
```
pducolin, line 23: **💭 thought**
This will require an update on the dd-agent side, I agree it is less misleading - fun fact I misled myself as I believed it was containing the protocol prefix

scenarios/aws/eks/fakeintake.go
```diff
@@ -0,0 +1,16 @@
...
+	ipAddress, err := ecs.FargateServiceFakeintake(env)
```
pducolin, line 10: Oh nice, in the end you could reuse the same deployment used for VMs, editing helm values to configure the agent to talk to fakeintake. Cool !

pducolin (approved):

vboulineau (approved):

components/datadog/agent/helm.go
```diff
@@ -243,3 +255,22 @@ func buildWindowsHelmValues(installName string, agentImagePath, agentImageTag, _
...
+		},
+		pulumi.Map{
+			"name":  pulumi.String("DD_LOGS_CONFIG_ADDITIONAL_ENDPOINTS"),
+			"value": pulumi.Sprintf("[{\"host\": \"%s\", \"port\": 80, \"is_reliable\": true, \"usessl\": false}]", fakeintake.Host),
+		},
+	}
+
+	if values["datadog"].(pulumi.Map)["env"] == nil {
```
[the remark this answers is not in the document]
L3n41c: I agree there are some repeated pieces of code in
```go
	if values["datadog"].(pulumi.Map)["env"] == nil {
		values["datadog"].(pulumi.Map)["env"] = additionalEndpointsEnvVar
	} else {
		values["datadog"].(pulumi.Map)["env"] = append(values["datadog"].(pulumi.Map)["env"].(pulumi.MapArray), additionalEndpointsEnvVar...)
	}
```

But as `values["datadog"].(pulumi.Map)["env"]` appears on the left hand side of the assignment, we would need a variable to capture a reference on it and not its value.

Something like:
```go
	env := &values["datadog"].(pulumi.Map)["env"]  // We need to get a reference here, not a copy of the value.
	if *env == nil {
		*env = additionalEndpointsEnvVar
	} else {
		*env = append(env, additionalEndpointsEnvVar...)
	}
```

But unfortunately, I don’t think it’s possible in GO to get a reference to a map element: https://github.com/golang/go/issues/11865.

components/datadog/agent/helm.go
L3n41c: The `values` manipulated here has been pre-filled by either `buildLinuxHelmValues` or `buildWindowsHelmValues` which are both filling `values` with:
```go
	return pulumi.Map{
		"datadog": pulumi.Map{
…
```

So, `values["datadog"]` cannot be `nil`;
`values["datadog"]` cannot be anything else than a `pulumi.Map`;
and we handle both cases where `env` key is present or not in `values["datadog"].(pulumi.Map)`
So, I think we’re fine because of how `values` is filled before the call to this function.

components/datadog/agent/helm.go
L3n41c: Good point !!!

scenarios/aws/eks/fakeintake.go
L3n41c: Good suggestion.
I made this function private: https://github.com/DataDog/test-infra-definitions/commit/ee56648ad2565da9a046af34a93de24d4cd9a09b#diff-a3d0fb2a3a7b9f7cbf5b219226efa06731aadded47235389dd6646941789f4ffR9.

components/datadog/fakeintake/connectionExporter.go
L3n41c: Definitively !
In fact, we currently put an IP address: https://github.com/DataDog/test-infra-definitions/blob/ee56648ad2565da9a046af34a93de24d4cd9a09b/scenarios/aws/ecs/fakeintake.go#L17

That’s why I initially replaced `url` by `ip address`.
But Vincent made me notice that, in practice, it could also be a DNS host name: https://github.com/DataDog/test-infra-definitions/pull/133#discussion_r1221696948.
Hence the renaming to `host`: https://github.com/DataDog/test-infra-definitions/pull/133/commits/30b5719537cfb2b960ccdc81bdd2b7f0ce971ff8.
But I forgot this one.

Thanks for having caught it !!!

I made the change: https://github.com/DataDog/test-infra-definitions/pull/133/commits/ee56648ad2565da9a046af34a93de24d4cd9a09b#diff-aa24a3296acc5428619072ea3e9efb11e4424e206c7d0803f9a610461f67c51bR12

scenarios/aws/eks/fakeintake.go
L3n41c: There’s undoubtedly pros and cons for both solutions: putting the fakeintake in ECS as it is already implemented today and used for the VM scenario, and putting the fakeintake in Kubernetes EKS.

I thought I would start with the simpler solution, that favors code re-use and generates the fewer maintenance.
We can always reconsider this choice later and move the fakeintake directly inside EKS if it happens to bring value.
But today, I’m happy with reusing the existing fakeintake deployment in ECS Fargate.

components/datadog/fakeintake/connectionExporter.go
L3n41c: @pducolin Here are the required changes on `datadog-agent` side: DataDog/datadog-agent#17566.

components/datadog/agent/helm.go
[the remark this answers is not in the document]
ogaca-dd: Indeed but you can still have a local variable for `values["datadog"].(pulumi.Map)`

PR #151, opened 2023-06-07
Enable possibility to give an AMI id to launch a specific VM

chouetz:
What does this PR do?
---------------------
Add a new option in invoke task to pass an AMI id as input

-------------------
AWS/VM

Motivation
----------
Possibility to run on older distribution than the ones we get by default today

Additional Notes
----------------

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -26,6 +26,8 @@ const (
 	DDInfraEnvironment       = "env"
 	DDInfraKubernetesVersion = "kubernetesVersion"
 	DDInfraOSFamily          = "osFamily"
+	DDInfraOSArchitecture    = "osArchitecture"
+	DDInfraOSAmiID           = "osAmiId"
 
 	// Agent Namespace
 	DDAgentDeployParamName               = "deploy"
@@ -86,6 +88,14 @@ func (e *CommonEnvironment) InfraEnvironmentNames() []string {
 
 func (e *CommonEnvironment) InfraOSFamily() string {
 	return e.GetStringWithDefault(e.InfraConfig, DDInfraOSFamily, "")
+}
+
+func (e *CommonEnvironment) InfraOSArchitecture() string {
+	return e.GetStringWithDefault(e.InfraConfig, DDInfraOSArchitecture, "")
+}
+
+func (e *CommonEnvironment) InfraOSAmiId() string {
+	return e.GetStringWithDefault(e.InfraConfig, DDInfraOSAmiID, "")
 }
 
 func (e *CommonEnvironment) KubernetesVersion() string {
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -94,7 +94,7 @@ func (e *CommonEnvironment) InfraOSArchitecture() string {
 	return e.GetStringWithDefault(e.InfraConfig, DDInfraOSArchitecture, "")
 }
 
-func (e *CommonEnvironment) InfraOSAmiId() string {
+func (e *CommonEnvironment) InfraOSAmiID() string {
 	return e.GetStringWithDefault(e.InfraConfig, DDInfraOSAmiID, "")
 }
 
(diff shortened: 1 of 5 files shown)


tasks/tool.py
```diff
@@ -63,6 +67,18 @@ def get_stack_json_outputs(full_stack_name: str) -> Any:
     return json.loads(output)
 
 
+def get_image_description(ami_id: str) -> Any:
+    r = subprocess.run(
+        f"aws-vault exec build-stable-developer -- aws ec2 describe-images --image-ids {ami_id}".split(),
```
vboulineau, line 72: Wrong AWS account

common/config/environment.go
```diff
@@ -26,6 +26,8 @@ const (
 	DDInfraEnvironment       = "env"
 	DDInfraKubernetesVersion = "kubernetesVersion"
 	DDInfraOSFamily          = "osFamily"
+	DDInfraOSArchitecture    = "osArchitecture"
```
vboulineau, line 29: This information somewhat conflicts with `instanceType`, we could see if we improve that.
`osAmiId` is also AWS specific and this a generic package, so I'm not sure if it's the right place for it.

tasks/vm.py
```diff
@@ -88,3 +95,25 @@ def _get_os_family(os_family: Optional[str]) -> str:
             f"The os family '{os_family}' is not supported. Possibles values are {', '.join(os_families)}"
         )
     return os_family
+
+def _get_os_information(os_family: Optional[str], ami_id: Optional[str]) -> tuple:
```
vboulineau, line 99: Isn't that what the Go code already does?

common/config/environment.go
chouetz: The concept of the image ID probably exist with other provider. I can maybe rename as `imageID` as a starting point?

tasks/vm.py
chouetz: Here we have at least an opportunity to guess the os family (rather than defaulting to ubuntu) but maybe it can be overridden later in the go code. Don't hesitate to tell me if I should move things elsewhere.
Plus I guess currently `arch` is forced to amd64

ogaca-dd (changes requested):
Thank you for the contribution :+1.
Note: I put `Request changes` because I have to open a PR before merging this one. See https://github.com/DataDog/test-infra-definitions/pull/151#discussion_r1229370113 for more information.

ogaca-dd:
> Thank you for the contribution :+1.
Note: I put Request changes because I have to open a PR before merging this one. See https://github.com/DataDog/test-infra-definitions/pull/151#discussion_r1229370113 for more information.

I have just opened the PR: https://github.com/DataDog/test-infra-definitions/pull/194

chouetz:
Rebased this PR on main on PR #209

tasks/vm.py
```diff
@@ -91,3 +98,25 @@ def _get_os_family(os_family: Optional[str]) -> str:
             f"The os family '{os_family}' is not supported. Possibles values are {', '.join(os_families)}"
         )
     return os_family
+
```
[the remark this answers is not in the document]
chouetz: Done on #209

tasks/vm.py
```diff
@@ -91,3 +98,25 @@ def _get_os_family(os_family: Optional[str]) -> str:
...
+    family, architecture = None, None
+    os_families = tool.get_os_families()
+    if ami_id is not None:
+        image = tool.get_image_description(ami_id)
+        if os_family is None: # Try to guess the distribution
+            try:
+                family = next(
+                    os
+                    for os in os_families
+                    if os in image["Description"].lower().replace(" ", "")
```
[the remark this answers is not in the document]
chouetz: I had a look to "official" images and this detection is pretty efficient. However, regarding the overall volume of images and the probability of strange information I allows passing the `-o` with the amiID to ease the detection

PR #160, opened 2023-06-12
Add pipieline ID tag

usamasaqib:
What does this PR do?
---------------------
This PR allows to add pipeline-id as a tag to an ec2 instance, if the appropriate environment variable is present.
It also makes the tags more consistent by always using dash ("-") instead of underscore ("_").

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -107,7 +107,7 @@ func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
 	// Map environment variables
 	lookupVars := []string{"TEAM", "PIPELINE_ID"}
 	for _, varName := range lookupVars {
-		if val, ok := os.LookupEnv(varName); ok {
+		if val := os.Getenv(varName); val != "" {
 			defaultTags[strings.ReplaceAll(
 				strings.ToLower(varName), "_", "-")] = pulumi.String(val)
 		}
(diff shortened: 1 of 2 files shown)


common/config/environment.go
```diff
@@ -105,10 +105,11 @@ func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
 	defaultTags["username"] = pulumi.String(user.Username)
 
 	// Map environment variables
-	lookupVars := []string{"DD_TEAM"}
+	lookupVars := []string{"TEAM", "PIPELINE_ID"}
 	for _, varName := range lookupVars {
-		if val := os.Getenv(varName); val != "" {
-			defaultTags[strings.ToLower(varName)] = pulumi.String(val)
+		if val, ok := os.LookupEnv(varName); ok {
```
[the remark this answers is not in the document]
usamasaqib: Gotcha. I will revert

common/config/environment.go
```diff
@@ -105,10 +105,11 @@ func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
 	defaultTags["username"] = pulumi.String(user.Username)
 
 	// Map environment variables
-	lookupVars := []string{"DD_TEAM"}
+	lookupVars := []string{"TEAM", "PIPELINE_ID"}
```
brycekahle, line 108: This changes from `DD_TEAM` to `TEAM`, is that intentional?

usamasaqib: Yes. `DD_` prefix seemed unnecessary. I checked and could not find usage of this tag anywhere. So I changed it to `TEAM`. 
No strong reason for changing it however.

vboulineau (approved):

PR #169, opened 2023-06-15
[task] Add cleanup task

pducolin:
What does this PR do?
---------------------

Add a task to clean pulumi stacks and lock files accidentally broken. It will remove any lock files and destroy and remove any stack prefixed by username, standard used in our tasks and in e2e tests to prefix stacks.

Stacks manually created with pulumi and not prefixed by the username will be skipped, to be discussed if we want to add them to the cleanup too.

-------------------

None

Motivation
----------

During our workshop one of the most requested feature was a cleanup script to unlock stacks that accidentally break during setup. This can happen for various reasons, most common cause is the aws token expired, or breaking a test execution while in timeout.

Additional Notes
----------------

### DIFF
--- a/tasks/clean.py
+++ b/tasks/clean.py
@@ -1,0 +1,31 @@
+import os
+import os.path
+from pathlib import Path
+
+from invoke.tasks import task
+from invoke.context import Context
+
+from .destroy import destroy
+from .tool import info, is_windows, warn
+
+@task
+def clean(_: Context) -> None:
+    """
+    Clean any environment created with invoke tasks or e2e tests
+    """
+    info("🧹 Clean up lock files")
+
+    if is_windows():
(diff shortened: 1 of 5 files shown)


tasks/clean.py
```diff
@@ -0,0 +1,31 @@
...
+
+from .destroy import destroy
+from .tool import info, is_windows, warn
+
+@task
+def clean(_: Context) -> None:
+    """
+    Clean any environment created with invoke tasks or e2e tests
+    """
+    info("🧹 Clean up lock files")
```
pducolin, line 16: From @vboulineau : check that this is a local state config

pducolin:
Moving this to datadog-agent as it makes more sense to have it there

PR #171, opened 2023-06-15
Update Go version and Azure SDK to get rid of dummy stack for runner

vboulineau:
What does this PR do?
---------------------

-------------------

Motivation
----------

Additional Notes
----------------


ogaca-dd (approved):
LGTM

pducolin (approved):

pducolin:
Should we change go version in go.mod too ?

vboulineau:
> Should we change go version in go.mod too ?

I don't think so, it bumps the minimum required version

PR #179, opened 2023-06-16
[task] ignore Task error

pducolin:
What does this PR do?
---------------------

Ignore general type issue in tasks collection

-------------------

None

Motivation
----------

This error seems to be on invoke side: `task` decorator does not return the proper type and does dynamic casting, which is not working well with pyright static type checking

Additional Notes
----------------


ogaca-dd (approved):
LGTM

PR #180, opened 2023-06-16
Olivierg/usage improvements

ogaca-dd:
What does this PR do?
---------------------

This PR makes some improvements to simplify the usage of ec2vm and azurevm:
- Fix a case mismatch between import "github.com/DataDog/test-infra-definitions/scenarios/aws/vm/ec2VM" and package name `ec2vm`
- Simplify the usage of `components/vm/params.go` by remove 2 generic types
- Rename `scenarios/aws/vm/os/` to `scenarios/aws/vm/ec2os` to avoid confusion with `/components/os`.
- WithImageName can automatically deduce the instance type
- Move VM option to its own package (`ec2params`) to improve discoverability. 
- Move agent/params.go to its own package.


scenarios/aws/vm/ec2vm/ec2VM.go
```diff
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"os"
 
-	commonos "github.com/DataDog/test-infra-definitions/components/os"
+	componentos "github.com/DataDog/test-infra-definitions/components/os"
```
pducolin, line 7: **❓ question**
Why do you need an alias ?

ogaca-dd: Because I have to import `os` for [os.ReadFile](https://pkg.go.dev/os#ReadFile)

PR #184, opened 2023-06-19
Fix UTF8 char in disable default providers. Add Lazy instantiation of providers. Use invoke to run commands

vboulineau:
What does this PR do?
---------------------

Fix UTF8 character in `Pulumi.yaml` that prevented `disable-default-providers` to work.
Provision providers lazily and provide helpers to set them in resources.
Use `invoke` runner to run commands instead of directly `subprocess`, make `debug` work for `eks`.

-------------------

All.

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -4,7 +4,8 @@ from .config import Config, get_full_profile_path
 import os
 import subprocess
 from invoke.context import Context
-from invoke.exceptions import Exit 
+from invoke.runners import Local, Result
+from invoke.exceptions import Exit
 from typing import Callable, List, Optional, Dict, Any
 import pathlib
 from . import tool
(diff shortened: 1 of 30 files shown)


brycekahle (approved):

tasks/deploy.py
```diff
@@ -121,7 +110,13 @@ def _get_app_key(cfg: Optional[Config]) -> str:
...
 
 
-def _get_key(key_name: str, cfg: Optional[Config], get_key: Callable[[Config], Optional[str]], env_key_name: str, expected_size: int) -> str:
+def _get_key(
+    key_name: str,
+    cfg: Optional[Config],
+    get_key: Callable[[Config], Optional[str]],
+    env_key_name: str,
+    expected_size: int,
+) -> str:
```
ogaca-dd, line 119: We should add a formater in order to not change the formatting of the code in each PR.

ogaca-dd (approved):

PR #194, opened 2023-06-22
Inject existing AWS environment into EC2 VM.

ogaca-dd:
What does this PR do?
---------------------

This PR adds methods to inject exiting environment into EC2 VM.

### DIFF
--- a/scenarios/aws/vm/ec2vm/ec2VM.go
+++ b/scenarios/aws/vm/ec2vm/ec2VM.go
@@ -27,39 +27,15 @@ type EC2VM struct {
 
 // NewEc2VM creates a new EC2 instance. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
 func NewEc2VM(ctx *pulumi.Context, options ...ec2params.Option) (*EC2VM, error) {
-	return newVM(ctx, options...)
-}
-
-type EC2UnixVM struct {
-	Infra
-	*commonvm.UnixVM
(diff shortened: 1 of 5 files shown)


scenarios/aws/vm/ec2vm/ec2VM.go
```diff
@@ -105,6 +81,39 @@ func newVM(ctx *pulumi.Context, options ...ec2params.Option) (*EC2VM, error) {
...
+		return nil, err
+	}
+
+	return &EC2UnixVM{
+		UnixVM: unixVM,
+		Infra:  vm.Infra,
+	}, nil
+}
+
+// NewUnixEc2VMW creates a new EC2 instance with an existing aws.NewEnvironment. By default use WithOS(os.UbuntuOS, os.AMD64Arch).
```
vboulineau, line 107: typo: incorrect comment copy/pasted from above

vboulineau (approved):

scenarios/aws/vm/ec2vm/ec2VM.go
vboulineau: The comment is actually wrong: `creates a new EC2 instance with an existing aws.NewEnvironment. By default use WithOS(os.UbuntuOS, os.AMD64Arch).`

While it does create the `aws.Environment` itself.

vboulineau (approved):

PR #196, opened 2023-06-22
Add xsltproc to the list of dependencies in the Dockerfile

usamasaqib:
What does this PR do?
---------------------
Adds `xsltproc` to the list of dependencies in the Dockerfile. `xsltproc` is used by the libvirt-sdk to modify the XML files defining the VM environment.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,9 @@ RUN apt-get update -y && \
   docker-ce \
   google-cloud-sdk \
   google-cloud-sdk-gke-gcloud-auth-plugin \
-  kubectl && \
+  kubectl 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -46,7 +46,9 @@ RUN apt-get update -y && \
   docker-ce \
   google-cloud-sdk \
   google-cloud-sdk-gke-gcloud-auth-plugin \
-  kubectl && \
+  kubectl
```
vboulineau, line 49:
```suggestion
  kubectl \
```

PR #197, opened 2023-06-23
Deploy the fake intake and the fake apps in the Kind on VM scenario

L3n41c:
What does this PR do?
---------------------

Deploy the fake intake and the fake applications in the “Kind on VM” scenario.

-------------------

* `aws/kindvm`

Motivation
----------

Test the Kubernetes-related features on an environment which is faster to spawn than EKS.

Additional Notes
----------------


pducolin (approved):
Thank you for the contribution. Not related to this PR, it looks like the `NewEcsFakeintake` could move out of scenarios as it is common code. Might move it to deployer function that can be injected to the fakeintake definition in components/datadog.

PR #203, opened 2023-06-26
Fix error when using the invoke task from outside the repository

pgimalac:
What does this PR do?
---------------------
Fix a typo with an uncalled function when using the invoke task from outside the repository.

-------------------
None as this is specifically for the invoke task.

Motivation
----------
Being able to use the invoke task from outside the repository.

Additional Notes
----------------
The `_get_root_path` function wasn't called in the argument list, which resulted in a corrupted command being executed, eg.
```
pulumi  -C <function _get_root_path at 0x1057d75e0> stack ls
```
which resulted in an error like
```
/bin/bash: function: No such file or directory
```

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -80,8 +80,9 @@ def _deploy(ctx: Context, stack_name: Optional[str], flags: Dict[str, Any], debu
     up_flags = ""
 
     # Checking root path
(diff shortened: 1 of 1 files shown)


ogaca-dd (approved):
LGTM 👍

PR #204, opened 2023-06-26
[docker] fix agent install not working after instance restart

pducolin:
What does this PR do?
---------------------

Install docker agent storing agent compose in home directory rather than in temporary directory

-------------------

docker

Motivation
----------

Temporary folder is cleaned up at reboot. Thus after the reboot of an instance where we installed agent on docker, the compose file created at deployment time was gone, like the container. Saving the compose file on home has the advantage of 
* not needing sudo powers
* allowing persistency of the file after reboot
* being available cross-platform

The helper can be used elsewhere.

Additional Notes
----------------

### DIFF
--- a/components/command/windowsOSCommand.go
+++ b/components/command/windowsOSCommand.go
@@ -55,7 +55,9 @@ func (fs windowsOSCommand) GetTemporaryDirectory() string {
 }
 
 func (fs windowsOSCommand) GetHomeDirectory() string {
-	return "$env:HOMEPATH\\$env:HOMEPATH"
+	// %HOMEDRIVE% returns the disk drive where home directory is located
+	// %HOMEPATH% returns the path to the home directory related to HOMEDRIVE
+	return "$env:HOMEDRIVE$env:HOMEPATH"
 }
 
 func (fs windowsOSCommand) BuildCommandString(
--- a/components/command/windowsOSCommand.go
+++ b/components/command/windowsOSCommand.go
@@ -54,6 +54,10 @@ func (fs windowsOSCommand) GetTemporaryDirectory() string {
 	return "$env:TEMP"
 }
 
+func (fs windowsOSCommand) GetHomeDirectory() string {
+	return "$env:HOMEPATH\\$env:HOMEPATH"
+}
+
 func (fs windowsOSCommand) BuildCommandString(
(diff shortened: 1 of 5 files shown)


components/command/windowsOSCommand.go
```diff
@@ -54,6 +54,10 @@ func (fs windowsOSCommand) GetTemporaryDirectory() string {
 	return "$env:TEMP"
 }
 
+func (fs windowsOSCommand) GetHomeDirectory() string {
+	return "$env:HOMEPATH\\$env:HOMEPATH"
```
ogaca-dd, line 58: Did you mean `"$env:HOMEPATH"`?

ogaca-dd (approved):
LGTM but please check my comment.

components/command/windowsOSCommand.go
pducolin: Oh thank you, I meant `$env: HOMEDRIVE$env:HOMEPATH`, good catch !

vboulineau:
Note:

```
not needing sudo powers
allowing persistency of the file after reboot
being available cross-platform
```

Tempdir does not require `sudo` and is available cross-platform (`%Temp%`).
The only change is the persistency on restart.
However note that `$HOME` is per user, so using `sudo` will affect the selected directory.

pducolin:
> Note:
> 
> ```
> not needing sudo powers
> allowing persistency of the file after reboot
> being available cross-platform
> ```
> 
> Tempdir does not require `sudo` and is available cross-platform (`%Temp%`). The only change is the persistency on restart. However note that `$HOME` is per user, so using `sudo` will affect the selected directory.

Thank you for your comment. Indeed, Temp is available cross-platform, but home is too thanks to `%HOMEDRIVE%%HOMEPATH%. Is there an exception you are aware of ? Unfortunately for docker we require persistency at reboot, and temp does not guarantee it.

I was aware that home is per user, what is the scenario you have in mind where using `sudo` affects the selected directory ? Doesn't it apply to temp too ? I can create a file/directory with sudo in temp and change its ownership. Curious to know more, and also if you think this is not the proper fix, if you have alternatives. I thought about adding a startup script that creates the file at startup, thought it was too complex and more work to have it working crossplatform, but I can rework that solution

PR #207, opened 2023-06-27
Create AWS environment once

usamasaqib:
What does this PR do?
---------------------
Fix bug introduced in #206 
The bug resulted in the creation of a new AWS environment for every instance created.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/microVMs/microvms/run.go
+++ b/scenarios/aws/microVMs/microvms/run.go
@@ -95,6 +95,9 @@ func newMetalInstance(instanceEnv *InstanceEnvironment, name, arch string, m con
 	var ami string
 
 	awsEnv := instanceEnv.Environment
+	if awsEnv == nil {
+		panic("no aws environment setup")
+	}
 
 	namer := namer.NewNamer(awsEnv.Ctx, fmt.Sprintf("%s-%s", awsEnv.Ctx.Stack(), arch))
 	if arch == ec2.AMD64Arch {
@@ -271,6 +274,10 @@ func run(e commonConfig.CommonEnvironment) (*ScenarioDone, error) {
 	}
 
 	instanceEnv := &InstanceEnvironment{&e, nil}
+	// We only setup an AWS environment if we need to launch
+	// a remote ec2 instance. This is determined by whether there
+	// is a non-local vmset in the configuration file. The following
+	// loop checks for this.
 	for arch := range archs {
 		if arch != LocalVMSet {
 			awsEnv, err := aws.NewEnvironment(instanceEnv.Ctx, aws.WithCommonEnvironment(&e))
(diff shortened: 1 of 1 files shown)


scenarios/aws/microVMs/microvms/run.go
```diff
@@ -273,9 +270,21 @@ func run(e commonConfig.CommonEnvironment) (*ScenarioDone, error) {
...
+	instanceEnv := &InstanceEnvironment{&e, nil}
+	for arch := range archs {
+		if arch != LocalVMSet {
+			awsEnv, err := aws.NewEnvironment(instanceEnv.Ctx, aws.WithCommonEnvironment(&e))
+			if err != nil {
+				return nil, err
+			}
+			instanceEnv.Environment = &awsEnv
+			break
+		}
```
val06, line 282: add comment please, explain why do you break after first non local arch

scenarios/aws/microVMs/microvms/run.go
```diff
@@ -90,14 +90,11 @@ func newEC2Instance(awsEnv aws.Environment, name, ami, arch, instanceType, keyPa
...
-func newMetalInstance(e commonConfig.CommonEnvironment, name, arch string, m config.DDMicroVMConfig) (*Instance, error) {
+func newMetalInstance(instanceEnv *InstanceEnvironment, name, arch string, m config.DDMicroVMConfig) (*Instance, error) {
 	var instanceType string
 	var ami string
 
-	awsEnv, err := aws.NewEnvironment(e.Ctx, aws.WithCommonEnvironment(&e))
-	if err != nil {
-		return nil, err
-	}
+	awsEnv := instanceEnv.Environment
```
val06, line 97: should we check the pointer for nil?

scenarios/aws/microVMs/microvms/run.go
```diff
@@ -123,23 +120,23 @@ func newMetalInstance(e commonConfig.CommonEnvironment, name, arch string, m con
...
 		Connection:    conn.ToConnectionOutput(),
 		Arch:          arch,
 		instanceNamer: namer,
 	}, nil
 }
 
-func newInstance(e commonConfig.CommonEnvironment, arch string, m config.DDMicroVMConfig) (*Instance, error) {
-	name := e.Ctx.Stack() + "-" + arch
+func newInstance(instanceEnv *InstanceEnvironment, arch string, m config.DDMicroVMConfig) (*Instance, error) {
+	name := instanceEnv.Ctx.Stack() + "-" + arch
```
val06, line 132: should we check the pointer for nil?

val06 (changes requested):
cred

scenarios/aws/microVMs/microvms/run.go
usamasaqib: I think it would be fine to panic here if the pointer is nil, since a valid environment can be regarded as an invariant for this program..

scenarios/aws/microVMs/microvms/run.go
usamasaqib: Same as below. We should just panic here.

val06 (approved):

PR #211, opened 2023-06-28
[DO NOT MERGE][task] use sandbox

pducolin:
What does this PR do?
---------------------

-------------------

Motivation
----------

Additional Notes
----------------


vboulineau (approved):

tasks/deploy.py
```diff
@@ -91,7 +91,7 @@ def _create_stack(ctx: Context, stack_name: str, global_flags: str):
 
 def _deploy(ctx: Context, stack_name: Optional[str], flags: Dict[str, Any], debug: Optional[bool]) -> str:
     stack_name = tool.get_stack_name(stack_name, flags["scenario"])
-    run_wrapper = "aws-vault exec sso-agent-sandbox-account-admin"
+    aws_account = flags["ddinfra:env"][len("aws/"):]
```
ogaca-dd, line 94: What do you think about passing `aws_account` as a function parameter instead of parsing `ddinfra:env`?

tasks/deploy.py
```diff
@@ -111,7 +111,7 @@ def _deploy(ctx: Context, stack_name: Optional[str], flags: Dict[str, Any], debu
 
     _create_stack(ctx, stack_name, global_flags)
 
-    cmd = f"{run_wrapper} -- pulumi {global_flags} up --yes -s {stack_name} {up_flags}"
+    cmd = f"{tool.get_aws_wrapper(aws_account)} -- pulumi {global_flags} up --yes -s {stack_name} {up_flags}"
```
ogaca-dd, line 114: 💯

ogaca-dd (approved):
LGTM 👍 .
Please note that https://github.com/DataDog/test-infra-definitions/pull/209 may need to be updated.

tasks/deploy.py
pducolin: Though about that, but thought also that we have the info in flags. True that flags are for the config map, but I was worried about passing twice the same info

pducolin: We could pass a confimap struct similar to what we have on agent side, and have a function to serialise it to the actual config map dictionary.

PR #214, opened 2023-06-29
Deploy the fake applications on ECS

L3n41c:
What does this PR do?
---------------------

Deploy the fake applications on ECS.

-------------------

* `aws/ecs`

Motivation
----------

Create some workload monitored by the agent so that we can test it.

Additional Notes
----------------

It’s the same fake applications as the ones we use for Kubernetes (#99), except:
* `nginx` and `redis` servers don’t have any autoscaler.
  ECS services can have an autoscaler but it cannot be plugged to Datadog, unlike on kubernetes where the cluster-agent provides a custom metrics server.
  So, adding an autoscaler on ECS service wouldn’t help validating any extra-feature of the agent.
* For `dogstatsd`, there’s only the UDS flavor because origin detection doesn’t work on ECS with UDP.

This PR contains only workload for ECS-EC2. ECS-Fargate will be in another PR.


components/datadog/apps/redis/ecs.go
```diff
@@ -0,0 +1,135 @@
+package redis
```
pducolin, line 1: **💭 thought**
We should probably have something similar for the fakeintake, I'm thinking moving it out of scenarios. Not related to this PR

pducolin (approved):

PR #222, opened 2023-07-03
Add team tag support in config file and config map

pducolin:
What does this PR do?
---------------------

- Add config map tags parameter (could not find any existing yet, can remove if it duplicates another parameter)
- Tag resources with config map tags
- Add team tag in tasks
 
-------------------

All

Motivation
----------

Add team tag to any AWS instance managed by pulumi and created locally

Additional Notes
----------------

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -26,6 +26,7 @@ const (
 	DDInfraOSFamily          = "osFamily"
 	DDInfraOSArchitecture    = "osArchitecture"
 	DDInfraOSAmiID           = "osAmiId"
+	DDInfraResourcesTags     = "resourcesTags"
 
 	// Agent Namespace
 	DDAgentDeployParamName               = "deploy"
@@ -92,7 +93,7 @@ func (e *CommonEnvironment) KubernetesVersion() string {
 }
 
 func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
-	defaultTags := pulumi.StringMap{
+	tags := pulumi.StringMap{
 		"managed-by": pulumi.String("pulumi"),
 	}
 
@@ -101,18 +102,29 @@ func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
 	if err != nil {
 		panic(err)
 	}
-	defaultTags["username"] = pulumi.String(user.Username)
+	tags["username"] = pulumi.String(user.Username)
 
 	// Map environment variables
 	lookupVars := []string{"TEAM", "PIPELINE_ID"}
 	for _, varName := range lookupVars {
 		if val := os.Getenv(varName); val != "" {
-			defaultTags[strings.ReplaceAll(
+			tags[strings.ReplaceAll(
 				strings.ToLower(varName), "_", "-")] = pulumi.String(val)
 		}
 	}
 
-	return defaultTags
+	// inject tags from config map
+	tagsFromConfigMap := e.GetStringListWithDefault(e.InfraConfig, DDInfraResourcesTags, []string{})
+	for _, tag := range tagsFromConfigMap {
+		nameAndValue := strings.Split(tag, "=")
+		if len(nameAndValue) != 2 {
+			continue
+		}
+		tags[strings.ReplaceAll(
+			strings.ToLower(nameAndValue[0]), "_", "-")] = pulumi.String(nameAndValue[1])
+	}
+
+	return tags
 }
 
 // Agent Namespace
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -50,6 +50,10 @@ def deploy(
     aws_account = cfg.get_aws().get_account()
     flags["ddinfra:env"] = "aws/" + aws_account
 
+    if cfg.get_aws().teamTag is None:
+        raise Exit("Error in config, missing configParams.aws.teamTag. Run `inv setup` again and provide a valid team name")
+    flags["ddinfra:resourcesTags"] = f"TEAM={cfg.get_aws().teamTag}"
+
     if install_agent:
         flags["ddagent:apiKey"] = _get_api_key(cfg)
 
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -31,30 +31,44 @@ def setup(_: Context) -> None:
     except Exception:
         config = Config.parse_obj({})
         
+    
+    # AWS config
+    setupAWSConfig(config)
+    # Agent config
+    setupAgentConfig(config)
 
+    config.save_to_local_config()
+    cat_profile_command = f"cat {get_full_profile_path()}"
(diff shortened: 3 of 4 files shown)


common/config/environment.go
```diff
@@ -26,6 +26,7 @@ const (
 	DDInfraOSFamily          = "osFamily"
 	DDInfraOSArchitecture    = "osArchitecture"
 	DDInfraOSAmiID           = "osAmiId"
+	DDInfraResourcesTags     = "resourcesTags"
```
vboulineau, line 29: Something like `extraResourceTags` is probably more accurate as this option does not replace default resource tags.

tasks/deploy.py
```diff
@@ -50,6 +50,10 @@ def deploy(
     aws_account = cfg.get_aws().get_account()
     flags["ddinfra:env"] = "aws/" + aws_account
 
+    if cfg.get_aws().teamTag is None:
+        raise Exit("Error in config, missing configParams.aws.teamTag. Run `inv setup` again and provide a valid team name")
+    flags["ddinfra:resourcesTags"] = f"TEAM:{cfg.get_aws().teamTag}"
```
vboulineau, line 55: s/TEAM/team

common/config/environment.go
```diff
@@ -101,18 +102,29 @@ func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
...
 	for _, varName := range lookupVars {
 		if val := os.Getenv(varName); val != "" {
-			defaultTags[strings.ReplaceAll(
+			tags[strings.ReplaceAll(
 				strings.ToLower(varName), "_", "-")] = pulumi.String(val)
 		}
 	}
 
-	return defaultTags
+	// inject tags from config map
```
vboulineau, line 116: Usually env vars have priority than config parameters

tasks/setup.py
```diff
@@ -84,7 +98,22 @@ def setup(_: Context) -> None:
...
-    # Agent config
+    checkKeyPair = ask(f"Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: ")
+    if len(checkKeyPair) > 0:
+        config.options.checkKeyPair = checkKeyPair.lower() == "y" or checkKeyPair.lower() == "yes"
+
+    if config.configParams.aws.teamTag is None:
+        config.configParams.aws.teamTag = ""
+    while True:
+        teamTag = ask(f"🔖 What is your team? It will tag all your resources by `TEAM:<team>`. Default [{config.configParams.aws.teamTag}]: ")
+        if len(teamTag) > 0:
```
ogaca-dd, line 109: Why this test is needed as `if len(config.configParams.aws.teamTag) > 0:` performs the same test?

common/config/environment.go
pducolin: It overwrite tags with the same key (ie if you pass `managed-by:totoro` it will replace `managed-by:pulumi`

common/config/environment.go
pducolin: I will do 
- default
- config
- env vars

tasks/setup.py
```diff
@@ -84,7 +98,22 @@ def setup(_: Context) -> None:
             break
         warn(f"{config.configParams.aws.publicKeyPath} is not a valid ssh key")
     
-    # Agent config
+    checkKeyPair = ask(f"Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: ")
```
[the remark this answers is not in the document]
pducolin: Bad copy paste, thank you

tasks/setup.py
pducolin: This is meant to handle a setup with an existing config file. We load parameters from the config file and they become the default. If the user insert some input, I replace the value in the config map. If the user does not input anything, it means they want to keep the default, old value. Now we have two options: it is a fresh setup, user di not add anything, they need to pass something. Or we are in a second run of the setup, we keep the old value.

tasks/setup.py
```diff
@@ -84,7 +98,22 @@ def setup(_: Context) -> None:
...
+    if len(checkKeyPair) > 0:
+        config.options.checkKeyPair = checkKeyPair.lower() == "y" or checkKeyPair.lower() == "yes"
+
+    if config.configParams.aws.teamTag is None:
+        config.configParams.aws.teamTag = ""
+    while True:
+        teamTag = ask(f"🔖 What is your team? It will tag all your resources by `TEAM:<team>`. Default [{config.configParams.aws.teamTag}]: ")
```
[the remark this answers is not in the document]
pducolin: Isn't it already the same tag ? 

```golang
f"TEAM:{cfg.get_aws().teamTag}"
```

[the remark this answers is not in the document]
pducolin: Do you mean that I should `s/TEAM/team` also here ?

common/config/environment.go
```diff
@@ -101,18 +102,29 @@ func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
...
+	tagsFromConfigMap := e.GetStringListWithDefault(e.InfraConfig, DDInfraResourcesTags, []string{})
+	for _, tag := range tagsFromConfigMap {
+		keyAndValue := strings.Split(tag, ":")
+		if len(keyAndValue) != 2 {
+			continue
```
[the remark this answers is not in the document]
pducolin: Added !

common/config/environment.go
```diff
@@ -63,6 +65,13 @@ func NewCommonEnvironment(ctx *pulumi.Context) (CommonEnvironment, error) {
 		CommonNamer:           namer.NewNamer(ctx, ""),
 		providerRegistry:      newProviderRegistry(ctx),
 	}
+	// store username
+	user, err := user.Current()
+	if err != nil {
+		panic(err)
```
ogaca-dd, line 71: Suggestion: Why not returning an error?

common/config/environment.go
```diff
@@ -101,37 +101,38 @@ func (e *CommonEnvironment) KubernetesVersion() string {
...
-	}
+func (e *CommonEnvironment) ResourcesTags() pulumi.StringMap {
+	tags := pulumi.StringMap{}
+
+	// default tags
+	extendTagsMap(tags, e.DefaultResourceTags())
+	// extended resource tags
+	extendTagsMap(tags, e.ExtraResourcesTags())
+	// env variable tags
+	extendTagsMap(tags, EnvVariableResourceTags())
```
ogaca-dd, line 135: 👍

ogaca-dd (approved):

common/config/environment.go
pducolin: Ah true !

PR #225, opened 2023-07-03
Put a load balancer in front of the fake intake

L3n41c:
What does this PR do?
---------------------

* Put a load balancer in front of the fake intake so that it keeps the same IP when it is OOM-killed.
* Use the `awsx` provider to define the other load balancers.

-------------------

* all

Motivation
----------

#### Put a load balancer in front of the fake intake

Although I agree that the OOM kills of the fake intake should be addressed independently of this workaround, I think there is a value to guarantee a stable IP to target the fake intake.
It will keep on working (although the data will vanish) when the fake intake will restart either involuntarily (in case of OOM kill or crash) or voluntarily (in case of deployment of a new dev version for ex.) without requiring a full rollout of the agents to update their config.

It has the side effect of removing the more or less hacky piece of code retrieving the IP of the fake intake task.

#### Use the `awsx` provider to define the other load balancers.

It simplifies a little bit our code as a single pulumi object is created to hold the three AWS objects (the load balancer itself, its listener and the target group).

Additional Notes
----------------

Load balancers are living outside of the ECS cluster.
As a consequence, they can be started to be provisioned very early, in parallel to the ECS capacity providers and the underlying VMs.
So, even if provisioning load balancers can take time, it shouldn’t impact the total elapsed time to spawn a stack because it can be provisioned in parallel with everything else and can be started at the beginning of the stack provisioning.

### DIFF
--- a/components/datadog/fakeintake/ecsFargate.go
+++ b/components/datadog/fakeintake/ecsFargate.go
@@ -16,22 +16,22 @@ const (
 	port          = 80
 )
 
-type Component struct {
+type Instance struct {
 	pulumi.ResourceState
 
 	Host pulumi.StringOutput
 }
 
-func FargateLinuxComponentDefinition(e aws.Environment) (*Component, error) {
+func NewECSFargateInstance(e aws.Environment) (*Instance, error) {
 	namer := e.Namer.WithPrefix("fakeintake")
 	opts := []pulumi.ResourceOption{e.WithProviders(config.ProviderAWS, config.ProviderAWSX)}
 
-	component := &Component{}
-	if err := e.Ctx.RegisterComponentResource("dd:fakeintake", namer.ResourceName("grp"), component, opts...); err != nil {
+	instance := &Instance{}
+	if err := e.Ctx.RegisterComponentResource("dd:fakeintake", namer.ResourceName("grp"), instance, opts...); err != nil {
 		return nil, err
 	}
 
-	opts = append(opts, pulumi.Parent(component))
+	opts = append(opts, pulumi.Parent(instance))
 
 	alb, err := lb.NewApplicationLoadBalancer(e.Ctx, namer.ResourceName("lb"), &lb.ApplicationLoadBalancerArgs{
 		Name:           e.CommonNamer.DisplayName(pulumi.String("fakeintake")),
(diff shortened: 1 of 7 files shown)


components/datadog/fakeintake/ecsFargate.go
```diff
@@ -14,41 +16,101 @@ const (
 	port          = 80
 )
 
-const (
-	DefaultFakeintakePort = 80
-)
+type Component struct {
```
vboulineau, line 19: Perhaps it should be called `Instance`, what we what people to see externally is `fakeintake.Instance`.

components/datadog/fakeintake/ecsFargate.go
```diff
@@ -14,41 +16,101 @@ const (
...
-			containerName: *fargateLinuxContainerDefinition(),
-		},
-		Cpu:    pulumi.StringPtr("256"),
-		Memory: pulumi.StringPtr("512"),
-		ExecutionRole: &awsx.DefaultRoleWithPolicyArgs{
-			RoleArn: pulumi.StringPtr(e.ECSTaskExecutionRole()),
+	Host pulumi.StringOutput
+}
+
+func FargateLinuxComponentDefinition(e aws.Environment) (*Component, error) {
```
vboulineau, line 25:
```suggestion
func NewECSFargateInstance(e aws.Environment) (*Component, error) {
```

L3n41c: That’s definitively a better name: https://github.com/DataDog/test-infra-definitions/pull/225/commits/e0a571d3c744c8b18fa3d40094eef2be83091f14.

vboulineau (approved):

components/datadog/fakeintake/ecsFargate.go
```diff
@@ -14,41 +16,101 @@ const (
...
-		TaskRole: &awsx.DefaultRoleWithPolicyArgs{
-			RoleArn: pulumi.StringPtr(e.ECSTaskRole()),
+		Listener: &lb.ListenerArgs{
+			Port:     pulumi.IntPtr(port),
+			Protocol: pulumi.StringPtr("HTTP"),
 		},
-		Family: e.CommonNamer.DisplayName(pulumi.String("fakeintake-ecs")),
-	}, e.WithProviders(config.ProviderAWS, config.ProviderAWSX))
-}
+	}, opts...)
```
ogaca-dd, line 52: This code is very close to https://github.com/DataDog/test-infra-definitions/pull/225/files#diff-5368ecf6a73f221e6f5376fbae6989ec734652bba2afbf1e3428aca810fe796cR29-R43. What do you think about having a helper function to reduce the verbosity of the code?

L3n41c: It’s very close because it’s the creation of the same resource type.
And it’s already a single function call (`lb.NewLoadBalancer`), but most arguments change.
image
I’m not sure that replacing one pulumi standard function call by our own custom function with as many arguments, but in flat instead of being structured would provide a big gain.

PR #226, opened 2023-07-05
Check for os family in name if no description is provided

KevinFairise2:
What does this PR do?
---------------------
When guessing the os family, if no Description is provided with the image, try to find in name.

-------------------
Create-vm invoke tasks used with custom AMI

Motivation
----------
Fix an error when AMI does not have a Description specified

Additional Notes
----------------


tasks/vm.py
```diff
@@ -109,11 +109,19 @@ def _get_os_information(
...
+                        for os in os_families
+                        if os in image["Description"].lower().replace(" ", "")
+                    )
+                else: # If the image has no description, check with name
+                    family = next(
+                        os
+                        for os in os_families
+                        if os in image["Name"].lower().replace(" ", "")
+                    )
+
```
ogaca-dd, line 124: What do you think about something like:
```python
if "Description" in image:
   image_info = image["Description"]
else:
   image_info =  image["Name"]
image_info =  image_info.lower().replace(" ", "")
family = next(
   os
   for os in os_families
   if os in image_info
)
```

ogaca-dd (approved):
LGTM

tasks/vm.py
KevinFairise2: Looks good

PR #227, opened 2023-07-05
[vm] add centos aws os type support

pducolin:
What does this PR do?
---------------------

Add support for CentOS 7

-------------------

VM

Motivation
----------

We frequently have to test CentOS 7 in QA as it is one of the supported Agent OS, the image is added to both `sandbox` and `agent-sandbox` public available AMI

Additional Notes
----------------


ogaca-dd (approved):
LGTM

PR #228, opened 2023-07-06
Fix running the invoke task from outside the directory

pgimalac:
What does this PR do?
---------------------
Fix a bug preventing using the invoke task from outside the directory.

It also uses the task's `Context` to run a command which was ran with `subprocess` previously.

-------------------
None as this is only related to the invoke task.

Motivation
----------
Being able to use the invoke task from outside the directory without having a crash.
The logic is very similar to what is done in `deploy.py`.

Additional Notes
----------------
I believe this bug started when I updated pulumi to 1.74 but I'm not 100% sure.

To reproduce the bug, run
`inv -r ~/dd/test-infra-definitions create-vm --no-install-agent --no-use-fakeintake` from outside the `test-infra-definitions` directory.

The root error is 
```error: if you're using the --stack flag, pass the fully qualified name (organization/project/stack)```
which happens because pulumi cannot find the stack file in the current directory.

### DIFF
--- a/tasks/tool.py
+++ b/tasks/tool.py
@@ -1,5 +1,7 @@
 import getpass
 import json
+import os
+import pathlib
 import platform
 import subprocess
 from io import StringIO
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: tasks/tool.py [hunk 1]
```diff
@@ -63,12 +65,20 @@ def get_stack_name_prefix() -> str:
...
-def get_stack_json_outputs(full_stack_name: str) -> Any:
-    output = subprocess.check_output(
-        ["pulumi", "stack", "output", "--json", "-s", full_stack_name]
+def get_stack_json_outputs(ctx: Context, full_stack_name: str) -> Any:
+    buffer = StringIO()
+    global_flags = ""
+
+    # Checking root path
+    root_path = _get_root_path()
+    if root_path != os.getcwd():
```
pducolin, line 74: **💬 suggestion**
You can achieve this by encapsulating the `ctx run` in ` with ctx.cd(root_path)`. I had to do a similar fix in the clean up task just seconds away https://github.com/DataDog/datadog-agent/pull/18052/files#diff-2e14236948e1a1fe87cd7e14b35b0beeb5bfa024b11f8e183e704ff37ab16a9fR155-R165

This works too by the way !

pducolin (approved):
Thank you for fixing this !

tasks/tool.py
pgimalac: I followed your suggestion and I think it's much clearer, thanks !

ogaca-dd (approved):

PR #216, opened 2023-06-30
Truncate the EKS node group name prefix if longer than 37 characters

L3n41c:
What does this PR do?
---------------------

Truncate the EKS node group name prefix if was supposed to be longer than 37 characters.

Which scenarios this will impact?
-------------------

* `aws/eks`

Motivation
----------

The [`datadog-agent` is currently failing](https://gitlab.ddbuild.io/DataDog/datadog-agent/-/jobs/289311691) with the following error:
```
aws:eks:NodeGroup aws-bottlerocket-ng  error: aws:eks/nodeGroup:NodeGroup resource 'aws-bottlerocket-ng' has a problem: expected length of node_group_name_prefix to be in the range (0 - 37), got ci-17317712-4670-eks-cluster-bottlerocket-ng. Examine values at 'NodeGroup.NodeGroupNamePrefix'.
```

Additional Notes
----------------

In the case of the CI, with this change, the node group name prefixes will become:
```
ci-17317712-4670-eks-cluster-linux-ng
ci-17317712-4670-eks-clust-linux-a-ng
ci-17317712-4670-eks-clus-bottlero-ng
```
See https://go.dev/play/p/7WiUMh_MQwd

PR #229, opened 2023-07-06
🐛 [fakeintake] hash LB and Target Group names

pducolin:
What does this PR do?
---------------------

Use md5 hash to name LB and Target Group name in fakeintake deployment

-------------------

VM

Motivation
----------

AWS Load Balancer and Target Group names must be 32 characters max and unique across an account. We were using `namer.DisplayName("fakeintake")` which was prefixing `fakeintake` with the stack name. The stack name is by default based on the environment flavour - for example `aws-vm-testsuite` - plus the username, and this is great as it creates unique resources per TestSuite / manually created environment. But it is easily longer than 32 chars. Using **md5** respects the 32 chars limit with a sufficiently low collision probability.

Additional Notes
----------------

### DIFF
--- a/common/namer/namer.go
+++ b/common/namer/namer.go
@@ -1,10 +1,9 @@
 package namer
 
 import (
-	"crypto/md5"
-	"encoding/hex"
 	"strings"
 
+	"github.com/DataDog/test-infra-definitions/common/utils"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
@@ -62,21 +61,10 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
 	}).(pulumi.StringOutput)
 }
 
-// HashedName return the md5 32-chars hex pulumi.StringInput of `parts` prefixed
+// DisplayNameWithMaxLen return pulumi.StringInput the concatanation of pulumi.StringInput `parts` prefixed
 // with the namer prefix
-func (n Namer) HashedName(parts ...pulumi.StringInput) pulumi.StringInput {
-	return n.DisplayName(parts...).ToStringOutput().ApplyT(func(name string) string {
-		hmd5 := md5.Sum([]byte(name))
-		return hex.EncodeToString(hmd5[:])
-	}).(pulumi.StringOutput)
-}
-
-// DisplayHashedName return the first 15 chars of the display name plus the first 16 chars
-// of the md5 32-chars hex pulumi.StringInput of display name
-func (n Namer) DisplayHashedName(parts ...pulumi.StringInput) pulumi.StringInput {
-	return pulumi.All(n.DisplayName(parts...), n.HashedName(parts...)).ApplyT(func(args []interface{}) string {
-		displayName := args[0].(string)
-		hashedName := args[1].(string)
-		return strings.Join([]string{displayName[:15], hashedName[:16]}, "-")
+func (n Namer) DisplayNameWithMaxLen(maxLen int, parts ...pulumi.StringInput) pulumi.StringInput {
+	return n.DisplayName(parts...).ToStringOutput().ApplyT(func(s string) string {
+		return utils.StrUniqueWithMaxLen(s, maxLen)
 	}).(pulumi.StringOutput)
 }
--- a/common/namer/namer.go
+++ b/common/namer/namer.go
@@ -1,6 +1,8 @@
 package namer
 
 import (
+	"crypto/md5"
+	"encoding/hex"
 	"strings"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
@@ -29,6 +31,8 @@ func (n Namer) WithPrefix(prefix string) Namer {
 	return childNamer
 }
 
+// ResourceName return the concatenation of `parts` prefixed
+// with namer prefix
 func (n Namer) ResourceName(parts ...string) string {
 	if len(parts) == 0 {
 		panic("Resource name requires at least one part to generate name")
@@ -42,6 +46,8 @@ func (n Namer) ResourceName(parts ...string) string {
 	return resourceName + strings.Join(parts, nameSep)
 }
 
+// DisplayName return pulumi.StringInput the concatanation of pulumi.StringInput `parts` prefixed
+// with the namer prefix
 func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
 	var convertedParts []interface{}
 	for _, part := range parts {
@@ -53,5 +59,14 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
 			strArgs = append(strArgs, arg.(string))
 		}
 		return n.ResourceName(strArgs...)
+	}).(pulumi.StringOutput)
+}
+
+// HashedName return the md5 32-chars hex pulumi.StringInput of `parts` prefixed
+// with the namer prefix
+func (n Namer) HashedName(parts ...pulumi.StringInput) pulumi.StringInput {
+	return n.DisplayName(parts...).ToStringOutput().ApplyT(func(name string) string {
+		hmd5 := md5.Sum([]byte(name))
+		return hex.EncodeToString(hmd5[:])
 	}).(pulumi.StringOutput)
 }
--- a/common/namer/namer.go
+++ b/common/namer/namer.go
@@ -70,3 +70,13 @@ func (n Namer) HashedName(parts ...pulumi.StringInput) pulumi.StringInput {
 		return hex.EncodeToString(hmd5[:])
 	}).(pulumi.StringOutput)
 }
+
+// DisplayHashedName return the first 15 chars of the display name plus the first 16 chars
+// of the md5 32-chars hex pulumi.StringInput of display name
+func (n Namer) DisplayHashedName(parts ...pulumi.StringInput) pulumi.StringInput {
+	return pulumi.All(n.DisplayName(parts...), n.HashedName(parts...)).ApplyT(func(args []interface{}) string {
+		displayName := args[0].(string)
+		hashedName := args[1].(string)
+		return strings.Join([]string{displayName[:15], hashedName[:16]}, "-")
+	}).(pulumi.StringOutput)
+}
--- a/common/randomgenerator/randomgenerator.go
+++ b/common/randomgenerator/randomgenerator.go
@@ -1,4 +1,4 @@
-package utils
+package randomgenerator
 
 import (
 	"github.com/DataDog/test-infra-definitions/common/config"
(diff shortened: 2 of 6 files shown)


common/namer/namer.go
```diff
@@ -55,3 +61,12 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
 		return n.ResourceName(strArgs...)
 	}).(pulumi.StringOutput)
 }
+
+// HashedName return the md5 32-chars hex pulumi.StringInput of `parts` prefixed
+// with the namer prefix
+func (n Namer) HashedName(parts ...pulumi.StringInput) pulumi.StringInput {
+	return n.DisplayName(parts...).ToStringOutput().ApplyT(func(name string) string {
```
vboulineau, line 68: I'm not sure this is the logic we want as we're loosing info about stack name. We should perhaps just review the naming, size of the stack name and then generate something like `<stack-name|16>-<hash|16>`.

pducolin: What if `stack-name` is > 32 chars ?

vboulineau: Unfortunately not all objects get tags (not all objects get names, even). We should review the naming patterns to understand the size of each piece but it's probably reasonable to limit the stack name to, say `16` chars

pducolin: There would be a higher risk of collisions, as we cannot control where the unique part comes from. The disadvantage here is that we don't know which username an LB belongs to, at first sight, right ?

pducolin: What about `{username.split(".")[0][0]}.{username.split(".")[1]}-{hash[:remaining_length]}` and we use the current hash, with the stack name included ?

pducolin: This would give something like `p.ducolin-1234BAD6` - for 32 chars total. We can remove the point.

vboulineau: That's the logic used in `datadog-agent` already: https://github.com/DataDog/datadog-agent/blob/39703bcc5900903d0eeac29f6e727dae00544710/test/new-e2e/runner/local-profile.go#L31-L58

The code inside `test-infra` should not try to do things with stack name, except validating the length if we want to do that.

pducolin: Tried with your suggestion and we I got something like `paola-ducolin-aw6727d0b5bb5b88da`

pducolin: Updated !

vboulineau: Do we need the `HashedName` function then?

L3n41c:
I already faced an issue linked to to the size of object names in the past, but it was for a different object type.
It was for EKS node group name prefixes.
Here is the corresponding PR: #216.
I think we should have a common way of limiting the size of object names. So, maybe we should revisit what I did in #216 if the proposition of this PR is better.

L3n41c (commented):
I already faced an issue linked to to the size of object names in the past, but it was for a different object type.
It was for EKS node group name prefixes.
Here is the corresponding PR: https://github.com/DataDog/test-infra-definitions/pull/216.
I think we should have a common way of limiting the size of object names. So, maybe we should revisit what I did in https://github.com/DataDog/test-infra-definitions/pull/216 if the proposition of this PR is better.

common/namer/namer.go
L3n41c: As the goal of object display names is to be as explicit as possible for users,
(if the goal was only to be unique, we could just not specify it and let pulumi generate a unique name for us)
wouldn’t it be valuable to try to keep the current untruncated concatenation of full stack name and full object name when it fits in the object name max size and switch to the truncated + hash name only when the full name doesn’t fit ?

pducolin:
@L3n41c agreed, some names might require more readable context then others, at minima we probably want a reference to the username with `<name[0]><familyName>` to easily identify resources. Should we have a function in namer that return a best effort of `<displayName[:maxLen/2 - 1]>-<hash(displayName)[:rest_of_the_available_length>`, passing `maxLen` to the function I added to `namer` ?

common/namer/namer.go
```diff
@@ -55,3 +61,22 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
 		return n.ResourceName(strArgs...)
 	}).(pulumi.StringOutput)
 }
+
+// HashedName return the md5 32-chars hex pulumi.StringInput of `parts` prefixed
+// with the namer prefix
+func (n Namer) HashedName(parts ...pulumi.StringInput) pulumi.StringInput {
```
ogaca-dd, line 67: May you can rename to `HashDisplayName`?

common/namer/namer.go
```diff
@@ -55,3 +61,22 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
...
+		return hex.EncodeToString(hmd5[:])
+	}).(pulumi.StringOutput)
+}
+
+// DisplayHashedName return the first 15 chars of the display name plus the first 16 chars
+// of the md5 32-chars hex pulumi.StringInput of display name
+func (n Namer) DisplayHashedName(parts ...pulumi.StringInput) pulumi.StringInput {
+	return pulumi.All(n.DisplayName(parts...), n.HashedName(parts...)).ApplyT(func(args []interface{}) string {
+		displayName := args[0].(string)
+		hashedName := args[1].(string)
```
ogaca-dd, line 79: If `len(displayName) <= 32`, don't you want to keep the full `displayName`?

common/namer/namer.go
```diff
@@ -55,3 +61,22 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
...
+	return n.DisplayName(parts...).ToStringOutput().ApplyT(func(name string) string {
+		hmd5 := md5.Sum([]byte(name))
+		return hex.EncodeToString(hmd5[:])
+	}).(pulumi.StringOutput)
+}
+
+// DisplayHashedName return the first 15 chars of the display name plus the first 16 chars
+// of the md5 32-chars hex pulumi.StringInput of display name
+func (n Namer) DisplayHashedName(parts ...pulumi.StringInput) pulumi.StringInput {
+	return pulumi.All(n.DisplayName(parts...), n.HashedName(parts...)).ApplyT(func(args []interface{}) string {
```
ogaca-dd, line 77: As `HashedNamed` is only used in this function, you can inline the code in this function and write something like
```go
return n.DisplayName(parts...).ToStringOutput().ApplyT(func(names ...string) string {
 displayName :=   n.DisplayName(names...)
if len(displayName) <= 32 {
    return displayName 
 }
hmd5 := md5.Sum([]byte(displayName))
hash := hex.EncodeToString(hmd5[:])
 return strings.Join([]string{displayName[:15], hashedName[:16]}, "-")
}
```

ogaca-dd (approved):

common/namer/namer.go
pducolin: As per the discussion with @L3n41c we will use this in other spots, wherever AWS is expecting a limited string size. I will add a max size to the `HashDisplayName` function and return the display name if the size is fine, or half of the string and half of the hash if it is too long

common/namer/namer.go
pducolin: Right !

common/namer/namer.go
pducolin: I'm considering having a `Limiter` out of namer, better the limiter logic doesn't have to know about `DisplayName`. It takes a string and limit it to half of it plus adds a substring of its hash to try to reduce collisions

L3n41c (changes requested):
`func (n Namer) DisplayName(resourceName)` was concatenating the name of the stack with the name of the resource.
So, the result could exceed or not the max name field length depending on both the stack name and the resource name.

But the `DisplayHashedName(resourceName)` uses only the resource name given in parameter, which is a literal.
In this case, all the logic is useless: let’s just check that the static resource name string is short enough. There’s nothing dynamic.

pducolin:
> `func (n Namer) DisplayName(resourceName)` was concatenating the name of the stack with the name of the resource. So, the result could exceed or not the max name field length depending on both the stack name and the resource name.
> 
> But the `DisplayHashedName(resourceName)` uses only the resource name given in parameter, which is a literal. In this case, all the logic is useless: let’s just check that the static resource name string is short enough. There’s nothing dynamic.

Agreed ! ---> https://github.com/DataDog/test-infra-definitions/pull/229#discussion_r1255728250

common/namer/namer.go
```diff
@@ -55,3 +60,11 @@ func (n Namer) DisplayName(parts ...pulumi.StringInput) pulumi.StringInput {
 		return n.ResourceName(strArgs...)
 	}).(pulumi.StringOutput)
 }
+
+// DisplayNameWithMaxLen return pulumi.StringInput the concatanation of pulumi.StringInput `parts` prefixed
+// with the namer prefix
+func (n Namer) DisplayNameWithMaxLen(maxLen int, parts ...pulumi.StringInput) pulumi.StringInput {
+	return n.DisplayName(parts...).ToStringOutput().ApplyT(func(s string) string {
+		return utils.StrUniqueWithMaxLen(s, maxLen)
+	}).(pulumi.StringOutput)
```
L3n41c, line 69: This function is:
* first concatenating the stack name with its arguments (resource name) and then
* truncate the result if it’s too long.

As a consequence, if the stack name is super verbose and the resource names very concise, the whole resource name will be systematically fully truncated whereas it would have been much more valuable to truncate the verbose stack name.

Ex.:
| `n.Ctx.Stack()` | `parts` | `maxLen` | Output of `DisplayNameWithMaxLen(maxLen, parts)` | What I would have liked |
|--------|--------|--------|--------|--------|
| `ci-17317712-4670-eks-cluster` | `fakeintake` | 32 | `ci-17317712-4670-eks-cluster-880` | `ci-17317712-4670-eks-fakeint-880` |
| `ci-17317712-4670-eks-cluster` | `nginx` | 32 | `ci-17317712-4670-eks-cluster-877` | `ci-17317712-4670-eks-cl-ngin-877` |
| `ci-17317712-4670-eks-cluster` | `redis` | 32 | `ci-17317712-4670-eks-cluster-7d0` | `ci-17317712-4670-eks-cl-redi-7d0` |

L3n41c (changes requested):
My concern with current implementation is that, given the length of the stack names in the CI, the current algorithm will just completely remove the resource name.
It means that the name of the LB (`fakeintake`, `nginx`, `redis`, etc.) will not be part of the final name at all, not even a truncated version of it.

If the final name doesn’t convey any information valuable for humans to understand what this resource is about, we can just let pulumi assign a completely random name for us.

common/namer/namer.go
pducolin: What if the namer had more knowledge of parts ? What if we had a "StackNamer" maybe ? We could have it knowing about "stackname", resource name and user name.

PR #230, opened 2023-07-07
Add serialization for Docker 

ogaca-dd:
What does this PR do?
---------------------
This PR adds the serialization for Docker.
It also performs some renamings.

-------------------

Motivation
----------

Additional Notes
----------------


components/datadog/agent/docker/agentDockerInstaller.go
```diff
@@ -43,6 +47,7 @@ func NewAgentDockerInstaller(vm *vm.UnixVM, options ...dockerparams.Option) (*Ag
 		for key, value := range dockerAgentParams.Env {
 			env[key] = pulumi.String(value)
 		}
+		agentContainerName = "docker-on-vm-compose-tmp-agent-1" // TODO: Improve the naming and make it more robust
```
pducolin: **❓ question**
Should we rather assign a name in the docker compose using `container_name` ? --> https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name

pducolin: Could be updated at https://github.com/DataDog/test-infra-definitions/blob/main/components/datadog/agent/docker.go#L14

pducolin:
This PR will conflict with https://github.com/DataDog/test-infra-definitions/pull/239, I suggest either first merging https://github.com/DataDog/test-infra-definitions/pull/239 or splitting this one in two, on one side adding the serialisation, on the other moving files

pducolin (approved):

components/datadog/agent/docker/agentDockerInstaller.go
ogaca-dd: As it was trivial I did it in this PR.

PR #233, opened 2023-07-07
Let pulumi assign LB name itself to avoid name too long error

L3n41c:
What does this PR do?
---------------------

Let pulumi assign LB name itself.

-------------------

* all

Motivation
----------

Our current logic to build smart resource names can exceed the maximum name size allowed.

Additional Notes
----------------

This PR will generate LB named like this:
image

### DIFF
--- a/components/datadog/fakeintake/ecsFargate.go
+++ b/components/datadog/fakeintake/ecsFargate.go
@@ -34,12 +34,12 @@ func NewECSFargateInstance(e aws.Environment) (*Instance, error) {
 	opts = append(opts, pulumi.Parent(instance))
 
 	alb, err := lb.NewApplicationLoadBalancer(e.Ctx, namer.ResourceName("lb"), &lb.ApplicationLoadBalancerArgs{
-		Name:           e.CommonNamer.DisplayName(pulumi.String("fakeintake")),
+		// Name:           e.CommonNamer.DisplayName(pulumi.String("fakeintake")),
 		SubnetIds:      pulumi.ToStringArray(e.DefaultSubnets()),
 		Internal:       pulumi.BoolPtr(!e.ECSServicePublicIP()),
 		SecurityGroups: pulumi.ToStringArray(e.DefaultSecurityGroups()),
 		DefaultTargetGroup: &lb.TargetGroupArgs{
-			Name:       e.CommonNamer.DisplayName(pulumi.String("fakeintake")),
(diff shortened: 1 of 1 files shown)


components/datadog/fakeintake/ecsFargate.go
```diff
@@ -34,12 +34,12 @@ func NewECSFargateInstance(e aws.Environment) (*Instance, error) {
 	opts = append(opts, pulumi.Parent(instance))
 
 	alb, err := lb.NewApplicationLoadBalancer(e.Ctx, namer.ResourceName("lb"), &lb.ApplicationLoadBalancerArgs{
```
pducolin, line 36: We can add more info to the logical name according to https://www.pulumi.com/docs/concepts/resources/names/, would name this `namer.ResourceName("fakeintake-lb").

pducolin (approved):

components/datadog/fakeintake/ecsFargate.go
L3n41c: The logical name already contains `fakeintake` because it’s one of the prefix of the used `namer`: https://github.com/DataDog/test-infra-definitions/blob/18a09ca9510a3c42d887ea73ddcc55bdf1f02006/components/datadog/fakeintake/ecsFargate.go#L26

pducolin:
I like this, it simplifies code a lot. The cons is we miss information about which user created the resource, that helps when manually checking resources on aws console and helps monitoring cloud cost. We need to ensure we have tags wherever they are supported.

components/datadog/fakeintake/ecsFargate.go
```diff
@@ -34,12 +34,12 @@ func NewECSFargateInstance(e aws.Environment) (*Instance, error) {
...
-		Name:           e.CommonNamer.DisplayName(pulumi.String("fakeintake")),
+		// Name:           e.CommonNamer.DisplayName(pulumi.String("fakeintake")),
```
ogaca-dd, line 37: Is there a reason to comment this code and not to remove it?

L3n41c: This was only a temporary quick workaround to unblock people, until a smarter fix (#240) is ready.
The goal of keeping the code commented was to not forget that, in the end, we would rather set an explicit name than letting pulumi generate one for us.
In #240, I’ve revisited all the locations where we use `(Name) DisplayName(…)`. If I had completely removed this piece of code, I would have missed those locations.

PR #238, opened 2023-07-08
Reduce the number of IP needed per LB and better spread VM on subnets

L3n41c:
What does this PR do?
---------------------

Make a better use of the IP available in the subnets in order to avoid premature exhaustion of IP.

| Before | After |
|--------|--------|
| All single VM use-cases were spawning the VM always on the same first subnet. This could lead to exhaustion of available IP in the first subnet whereas the other two subnets were barely used. | Single VM use-cases are now spawning the VM on a subnet randomly chosen among the 3 available ones. It spreads the usage on all 3 subnets. |
| Load balancers were consuming 3 IP, one per subnet in which they were present. | It’s not possible to have a load balancer inside a single availability zone. But, instead of putting them in all 3 zones, we now put them only in 2 zones randomly chosen among the 3 available ones. |
| ECS and EKS clusters were spread on the 3 zones. It didn’t consume more IP. | But concentrating them on 2 zones reduces the probability of cross-zone traffic from 2 chances out of 3 to 1 chance out of 2. |

-------------------

all

Motivation
----------

We already faced errors due to IP exhaustion.

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/microVMs/microvms/run.go
+++ b/scenarios/aws/microVMs/microvms/run.go
@@ -73,7 +73,7 @@ func newEC2Instance(awsEnv aws.Environment, name, ami, arch, instanceType, keyPa
 
 	instance, err := awsEc2.NewInstance(awsEnv.Ctx, awsEnv.Namer.ResourceName(name), &awsEc2.InstanceArgs{
 		Ami:                 pulumi.StringPtr(ami),
-		SubnetId:            pulumi.StringPtr(awsEnv.DefaultSubnets()[0]),
+		SubnetId:            awsEnv.RandomSubnets().Index(pulumi.Int(0)),
 		InstanceType:        pulumi.StringPtr(instanceType),
 		VpcSecurityGroupIds: pulumi.ToStringArray(awsEnv.DefaultSecurityGroups()),
 		KeyName:             pulumi.StringPtr(keyPair),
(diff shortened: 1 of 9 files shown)


vboulineau (approved):

scenarios/aws/microVMs/microvms/run.go
```diff
@@ -73,7 +73,7 @@ func newEC2Instance(awsEnv aws.Environment, name, ami, arch, instanceType, keyPa
 
 	instance, err := awsEc2.NewInstance(awsEnv.Ctx, awsEnv.Namer.ResourceName(name), &awsEc2.InstanceArgs{
 		Ami:                 pulumi.StringPtr(ami),
-		SubnetId:            pulumi.StringPtr(awsEnv.DefaultSubnets()[0]),
+		SubnetId:            awsEnv.RandomSubnets().Index(pulumi.Int(0)),
```
vboulineau, line 76: @usamasaqib As you're tweaking the subnets in your test, does that work for you?

usamasaqib: We are tweaking the subnets for two reason.
1. The default subnet in datadog-agent-qa had capacity issues for the metal instance we were spawning. Therefore, we passed different subnets to target.
2. If the stack creation fails due to capacity shortage, we retry the stack with a different subnet.

We haven't tried addressing the IP problem by changing subnets.

usamasaqib, line 76: I think we would like to opt out of randomly selecting the subnets. We did some coarse experimentation to see which AZs had the most capacity for the resources we want, and we are selecting the subnets based on that.

usamasaqib (changes requested):

scenarios/aws/microVMs/microvms/run.go
vboulineau: You can "opt-out" if you pass only 1 subnet, which, I think, is what you do?

usamasaqib (approved):

PR #240, opened 2023-07-10
AETT-186 Ensure `namer.DisplayName` doesn’t generate a name exceeding a maximum

L3n41c:
What does this PR do?
---------------------

Add a new “maximum length” parameter to `(Namer) DisplayName(…)` to ensure that the output string isn’t longer than the maximum length authorized for the `Name` field by the cloud provider.

AWS has different constraints on name field depending on the object.

-------------------

* all

Motivation
----------

We’ve already been hit several times in the past by issues due to names being too long:
* EKS node group name prefix exceeding the 37 characters limit was addressed by #216
* Load balancer name exceeding the 32 characters limit was addressed by #229 and #233

This PR is an attempt to revisit the above mentioned fixes and to provide a longer term solution addressing the issue of name field maximum size for all resource types.

It adds a mandatory “maximum length” parameter to `(Namer) DisplayName(…)` so that the developer should ask themselves what is the limit of the `Name` field they want to fill. It’s usually specified in the comment of the `pulumi` API.
Ex.: https://github.com/pulumi/pulumi-aws/blob/master/sdk/go/aws/applicationloadbalancing/loadBalancer.go#L419-L422

In case a truncation is needed, all the parts that are concatenated to form the name are truncated.
All the parts are truncated with the same proportion so that a big part cannot cause a small part to be completely removed.

In case a truncation is needed, a hash is suffixed.

Additional Notes
----------------

Here are [some real life example of what the new `(Namer) DisplayName(maxLen, parts …)` is generating](https://github.com/DataDog/test-infra-definitions/pull/240/commits/2663f8576c37e2185143c6f76e144fc9a4c73379#diff-56e58fdae39f12fd3e6c22e64f4beebe60f8f974aa9e8f1b209494b517c21235R108-R140):
| `maxLen` | `Stack()` | `parts ...` | Output |
|--------|--------|--------|--------|
| 37 | `ci-17317712-4670-eks-cluster` | `linux`, `ng` | `ci-17317712-4670-eks-cluster-linux-ng` |
| 37 | `ci-17317712-4670-eks-cluster` | `linux-arm`, `ng` | `ci-17317712-4670-eks-c-linux-a-ng-458` |
| 37 | `ci-17317712-4670-eks-cluster` | `bottlerocket`, `ng` | `ci-17317712-4670-eks--bottleroc-n-a2e` |
| 32 | `ci-17317712-4670-eks-cluster` | `fakeintake` | `ci-17317712-4670-eks-fakeint-5f1` |
| 32 | `ci-17317712-4670-eks-cluster` | `nginx` | `ci-17317712-4670-eks-cl-ngin-db3` | 
| 32 | `ci-17317712-4670-eks-cluster` | `redis` | `ci-17317712-4670-eks-cl-redi-7de` |


common/namer/namer.go
```diff
@@ -1,57 +1,131 @@
...
 	"strings"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"github.com/samber/lo"
 )
 
 const nameSep = "-"
 
+// In case of truncation, size of the hash suffix in percentage of the full size
+const hashSizePercent = 10
```
pducolin, line 17: **💭 thought**
I can see this being helpful for short names, should we also have an upper bound with `maxHashSize` ? More for readability than for it being used at anytime, we don't expect to cut names for resources limited to 400 chars and thus using 40 chars of hash

common/namer/namer.go
```diff
@@ -1,57 +1,131 @@
...
+func (n Namer) DisplayName(maxLen int, parts ...pulumi.StringInput) pulumi.StringInput {
 	var convertedParts []interface{}
 	for _, part := range parts {
 		convertedParts = append(convertedParts, part)
 	}
 	return pulumi.All(convertedParts...).ApplyT(func(args []interface{}) string {
-		strArgs := []string{n.ctx.Stack()}
+		strArgs := make([]string, 1, 1+len(n.prefixes)+len(args))
+		strArgs[0] = n.ctx.Stack()
+		strArgs = append(strArgs, n.prefixes...)
```
pducolin, line 63: **💭 thought**
Seen it here it is weird to prepend everything with stack name when we call namer parts "prefixes". We might come back to the naming - not related to this PR

common/namer/namer.go
```diff
@@ -1,57 +1,131 @@
...
+	totalOutputSize := maxLength - len(tokens)*len(nameSep) - hashSize
+	totalInputSize := lo.Sum(lo.Map(tokens, func(s string, _ int) int { return len(s) }))
+	prevY := 0
+	X := 0
+	for _, tok := range tokens {
+		X += len(tok)
+		nextY := (X*totalOutputSize + totalInputSize/2) / totalInputSize
+		tokSize := nextY - prevY
+		prevY = nextY
+		sb.WriteString(tok[:tokSize])
```
pducolin, line 120: **❓ question**
Is there a chance we strip out a token ?

common/namer/namer_test.go
```diff
@@ -0,0 +1,173 @@
...
+			noHash:    true,
+			expected:  "FfffOooo-BbbAaa-BbAa-QU",
+		},
+		{
+			maxLength: 13,
+			tokens:    []string{"FfffOoooOooo", "BbbAaaRrr", "BbAaZz", "QUX"},
+			noHash:    true,
+			expected:  "Ffff-Bbb-Bb-Q",
+		},
+		// Truncation are spread evenly on all tokens
```
pducolin, line 47: **🥜 nitpick**
```suggestion
		// Truncation are spread best effort evenly on all tokens
```

common/namer/namer_test.go
```diff
@@ -0,0 +1,173 @@
...
+func outputOK(maxLength int, tokens []string, output string) bool {
+	full := strings.Join(tokens, "-")
+	if len(full) <= maxLength {
+		return output == full
+	}
+
+	return len(output) == maxLength
+}
+
+func FuzzJoinWithMaxLength(f *testing.F) {
```
pducolin, line 160: **👏 praise**
😍

pducolin (approved):
Nice !

common/namer/namer.go
L3n41c: In fact, there’s already an implicit upper bound to the size of the hash suffix that comes from the hash function.
We’re using a hash function that returns a 64 bits long hash. So, it’s hexadecimal representation is 16 characters long.
In the case we have a field limited to 400 chars and so many long tokens that a truncation is needed, the hash suffix will be exactly 16 char long.
So, introducing yet another `maxHashSize` parameter might be overkill. WDYT ?
But I can add a test with a 400 char limit to validate this behavior and make it more explicit for code readers ?

common/namer/namer.go
L3n41c: Indeed, in the version you reviewed, it was possible to have tokens fully stripped out.
I’ve just slightly modified to algorithm to transition from the “truncated format” to the “hash only” format sooner.
Here is the change that I did: https://github.com/DataDog/test-infra-definitions/pull/240/commits/6d33115c6f03497cb67f5b88e5caa6becea35490

common/namer/namer.go
L3n41c: Indeed, the “prefixes” are before the “parts…”, but the stack name is at the beginning of everything.

common/namer/namer.go
pducolin: Ah good, I thought you would have used 40 chars of hash on a 400 chars name. Ok if it is already upper limited to the max hash size !

pducolin (approved):
Thank you for adding tests, this is the best we can do to get the best of both worlds: context on what / who created a resource and make aws happy with its limited resource name lengths. We might revisit this to have priorities in truncation, so that for example we don't cut pipeline id / username or we have each part keeping the logic on how to truncate itself, but I feel that would be overkilled for the risk of that happening. This is great code we can re-edit in the future, if the need raises.

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
```
ogaca-dd, line 121: Can you explain how do you ensure that `tokSize` is less than `len(tok)`?

ogaca-dd (commented):
Thank for this PR!!!!!!

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
+		}
+		return fullhash
+	}
+
+	var sb strings.Builder
+
+	// Truncate all tokens in the same relative proportion
+	totalOutputSize := maxLength - len(tokens)*len(nameSep) - hashSize
+	totalInputSize := lo.Sum(lo.Map(tokens, func(s string, _ int) int { return len(s) }))
+	prevY := 0
```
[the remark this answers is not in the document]
L3n41c: The goal is to truncate all the tokens with the same ratio.
If the token lengths could be rational numbers, the formula would have just been:
```go
ratio := totalOutputSize / totalInputSize  // At this point, we have the guarantee that totalOutputSize < totalInputSize
tokSize := len(token) * ratio
```

But string lengths are integer, so we have to use integers and Euclidean division.
If we were computing `tokSize = len(token) * totalOutputSize / totalInputSize` for all the tokens, the integer rounding errors would accumulate. And the sum of the length of the truncated tokens at the end wouldn’t be exactly `totalOutputSize`.
So, instead of that, we do a linear interpolation to place the boundaries of the tokens.

Here is what the variables of the algorithm are representing when we are computing the new size of the token at index 2:
image

common/namer/namer.go
L3n41c: As explained above, `tokSize` is a rounded value of `len(token) * totalOutputSize / totalInputSize` and, at this point of the function, we know that we have `totalOutputSize < totalInputSize`
The rest are details to ensure that integer rounding division error do not accumulate.

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
+			return fullhash[:maxLength]
+		}
+		return fullhash
+	}
+
+	var sb strings.Builder
+
+	// Truncate all tokens in the same relative proportion
+	totalOutputSize := maxLength - len(tokens)*len(nameSep) - hashSize
+	totalInputSize := lo.Sum(lo.Map(tokens, func(s string, _ int) int { return len(s) }))
```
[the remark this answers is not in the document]
L3n41c: Indeed, the definition of `totalInputSize` can be moved at the beginning of the function: https://github.com/DataDog/test-infra-definitions/pull/246/files#diff-811a454eee674bac80521cade037c79db0a3cb01f01c3818181b166587e083bbR74

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
+		hash = hash[:hashSize]
+	} else {
+		hashSize = len(hash)
+	}
+
+	// For test purpose, we have an option to completely strip the output of the hash suffix
+	// At this point, `hashSize` is the size of the hash suffix without the dash separator
+	// -1 means that we want to also strip the dash
+	if noHash {
+		hashSize = -1
```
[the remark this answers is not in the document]
L3n41c: Correct !
https://github.com/DataDog/test-infra-definitions/pull/246/files#diff-811a454eee674bac80521cade037c79db0a3cb01f01c3818181b166587e083bbR101

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
 	}
 }
 
 func (n Namer) WithPrefix(prefix string) Namer {
-	childNamer := Namer{}
-	if n.prefix != "" {
-		childNamer.prefix = n.prefix + nameSep
+	return Namer{
+		ctx:      n.ctx,
+		prefixes: append(n.prefixes, prefix),
```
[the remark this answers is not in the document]
L3n41c: Good catch !
This should be fixed by an explicit copy: https://github.com/DataDog/test-infra-definitions/pull/246/files#diff-811a454eee674bac80521cade037c79db0a3cb01f01c3818181b166587e083bbR41-R42.

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
+	"github.com/samber/lo"
 )
 
 const nameSep = "-"
 
+// In case of truncation, size of the hash suffix in percentage of the full size
+const hashSizePercent = 10
+
+// Used only for tests to skip hash suffix to ease test results validation
+var noHash = false
```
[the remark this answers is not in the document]
L3n41c: I fear that adding an extra parameter to the `joinWithMaxLength` function would be interpreted as a part of its API and the function users may wonder what value they should pass.
But, it’s not part of the contract. It should always be `False` in “normal” usage.
I think it’s best to not have testing details visible in APIs.

common/namer/namer.go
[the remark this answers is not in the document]
ogaca-dd: Thanks for the explanation. Can you rename `prevY` to something like `prevOutputSize` and `X` to `totalTokenSize`?

common/namer/namer.go
ogaca-dd: Can you explain why you have `totalInputSize/2` in `nextY := (X*totalOutputSize + totalInputSize/2) / totalInputSize`?

pducolin: Looking at @ogaca-dd question, I can see as I also did not understand this formula. Could you split in "safeRoundedRatio" and then using `X*safeRoundedRatio`?

If I get it right, you are using `+totalInputSize/2` to make the ratio smaller, by increasing by 0.5 the computed `nextY`, as this equals
`round(X*totalOutputSize/totalInputSize + 0.5)` if we were using floats.

pducolin: Is there a reason to use `0.5` instead of another factor ?

pducolin: Thank you @ogaca-dd for asking the question !

L3n41c: So…

L3n41c: If we look at [the graph in the comment above](https://github.com/DataDog/test-infra-definitions/pull/240#discussion_r1261087617), we see that what we are trying to compute is the linear interpolation between points `(0, 0)` and `(totalInputSize, totalOutputSize)` at abscissa `X`.
It’s nothing more than a rule of 3.

So, what we a trying to compute is mathematically :
```
(math) Y = X × totalOutputSize / totalInputSize
```
Except that numbers of characters are integer, so, we have to find the closest integer to that value.
That is mathematically :
```
(math) Y = round(X × totalOutputSize / totalInputSize)
```

The problem is that the Go integer division doesn’t return the round value of the real division, but the quotient of the Euclidean division, which is the floor of the real division.
The output of the following Go expression:
```go
(GO) a / b
```
where `a` and `b` are `int` variables is:
```
(math) floor( a / b )
```

So, we are leveraging the following property:
```
(math) round( x ) = floor( x + ½ )
```

Except that, with integer computing, we cannot represent ½.
So, we do:
```
(math) Y
= round( X × totalOutputSize / totalInputSize )
= floor( X × totalOutputSize / totalInputSize + ½ )
= floor( (X × totalOutputSize + totalInputSize/2) / totalInputSize)
```
And this later expression can be computed in Go:
```go
Y := (X*totalOutputSize + totalInputSize/2) / totalInputSize
```

L3n41c:
> Could you split in "safeRoundedRatio" and then using X*safeRoundedRatio?

No, we can’t because
```go
safeRoundedRatio := totalOutputSize / totalInputSize
```
would always be `0`.

When we write:
```
nextY := X * totalOutputSize / totalInputSize
```
we leverage the left-associativity of the `*` and `/` operators to ensure that the multiplication is done prior to the division.
If we pre-compute `safeRoundedRatio := totalOutputSize / totalInputSize`, the division is done first and the result is `0`.

common/namer/namer.go
```diff
@@ -1,57 +1,132 @@
...
+	} else {
+		hashSize = len(hash)
+	}
+
+	// For test purpose, we have an option to completely strip the output of the hash suffix
+	// At this point, `hashSize` is the size of the hash suffix without the dash separator
+	// -1 means that we want to also strip the dash
+	if noHash {
+		hashSize = -1
+	}
```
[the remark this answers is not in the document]
L3n41c: I’ve extracted the function to compute the hash here: https://github.com/DataDog/test-infra-definitions/pull/252/files#diff-811a454eee674bac80521cade037c79db0a3cb01f01c3818181b166587e083bbR137-R143

common/namer/namer.go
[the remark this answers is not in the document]
L3n41c: As this variable is now used for two different purposes, I’ve renamed it `sumTokensSize`.
https://github.com/DataDog/test-infra-definitions/pull/252/files#diff-811a454eee674bac80521cade037c79db0a3cb01f01c3818181b166587e083bbR74
It better describes it’s content.

common/namer/namer.go
ogaca-dd: Thank you for this clear explanation 💯 . Can you add a little comment in the code explaining that `totalInputSize/2` is used to have the floor value?

L3n41c: As this naive algorithm wasn’t as trivial to understand as I initially expected, I think it’s better to move back to the optimal well-known solution: #252.
At least, explanations about Bresenham’s algorithm are easy to find.

PR #244, opened 2023-07-11
Export proposal using ComponentResources and Marshall

vboulineau:
What does this PR do?
---------------------
```
Current stack outputs (5):
    OUTPUT                           VALUE
    dd-kubeCluster-kind              {"kubeConfig":"apiVersion: v1\nclusters:\n- cluster:\n    insecure-skip-tls-verify: true\n    server: https://<ip>:8443\n  name: kind-kind\ncontexts:\n- context:\n    cluster: kind-kind\n    user: kind-kind\n  name: kind-kind\ncurrent-context: kind-kind\nkind: Config\npreferences: {}\nusers:\n- name: kind-kind\n  user:\n    client-certificate-data: <redacted>\n"}
```

-------------------

Motivation
----------

Additional Notes
----------------


components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+type KubernetesCluster struct {
+	pulumi.ResourceState
+	exportName string
+
+	KubeConfig pulumi.StringOutput `export:"kubeConfig"`
+}
+
+// The target type that's going to be deserialized on the other side.
+// We could unit test the matching if required.
+// The E stands for exported, probably can find something better
```
pducolin, line 28: **💬 suggestion**
Agree, can be confused with `Exporter`. Maybe `KubernetesClusterDeserialiser`

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+	if ek.Key == "" {
+		return errors.New("missing key")
+	}
+
+	val, found := output[ek.Key]
+	if !found {
+		return errors.New("output not found")
+	}
+
+	// it *might* happen that val.Value is a map[interface{}]interface{} or something, which
```
pducolin, line 50: **❓ question**
We should find a way to limit values to json strings, maybe a custom linter

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+	// would mean that we need to implement our own reflection as well to import.
+	return json.Unmarshal(val.Value.([]byte), ek)
+}
+
+// Boilerplate component resource creation
+func newClusterComponent(e *config.CommonEnvironment, name string, opts ...pulumi.ResourceOption) (*KubernetesCluster, error) {
+	comp := &KubernetesCluster{
+		// as we components we can guarantee key unicity without asking user to provide name again
+		exportName: e.CommonNamer.ResourceName("dd", "kubeCluster", name),
```
pducolin, line 59: **👏 praise**

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (k *KubernetesCluster) Export(e *config.CommonEnvironment, exportedObj *EKubernetesCluster) {
+	// To reproduce the current cross-program assignment in `datadog-agent`, not technically required.
+	if exportedObj != nil {
+		exportedObj.Key = k.exportName
+	}
+
+	// This can be easily replaced by reflection traversal but not enough time rn
+	e.Ctx.Export(k.exportName, pulumi.JSONMarshal(pulumi.ToMapOutput((map[string]pulumi.Output{
```
pducolin, line 76: **💭 thought**
We can probably ensure we only use `pulumi.JSONMarshal` by using a generic and having each implementation providing a 

```golang
(t *T) func(dataToExport map[string]pulumi.Output) err
```

so that we are sure we limit exported values to json

pducolin (approved):
I like this proposal ! I like how we keep marshalling and unmarshalling close together, makes it easier to see how data flows. We might want to add some generics / interfaces to guide and limit data types and ensure these marshalling helper classes have required parameters such as `Key`

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+		return errors.New("missing key")
+	}
+
+	val, found := output[ek.Key]
+	if !found {
+		return errors.New("output not found")
+	}
+
+	// it *might* happen that val.Value is a map[interface{}]interface{} or something, which
+	// would mean that we need to implement our own reflection as well to import.
```
ogaca-dd, line 51: Isn't the goal of this proposal to always use JSON?

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+		exportedObj.Key = k.exportName
+	}
+
+	// This can be easily replaced by reflection traversal but not enough time rn
+	e.Ctx.Export(k.exportName, pulumi.JSONMarshal(pulumi.ToMapOutput((map[string]pulumi.Output{
+		"kubeConfig": k.KubeConfig,
+	}))))
+}
+
+// // Usage in datadog-agent
```
ogaca-dd, line 81: Can you elaborate a little the usage in datadog-agent for example what will you keep, what will you remove?

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+// type KubeSuite struct {
+// 	e2e.Suite
+
+// 	cluster *EKubernetesCluster
```
ogaca-dd, line 85: Can you explain how marshalling is related to e2e.Suite?

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+// 	// cluster *EKubernetesCluster `name:"kind"` if `k.Export` is called inside existing runFunc (like scenarios/aws/kindvm/run.go or eks/run.go)
+// }
+
+// If willing to use type matching
+// runFunc = func(ctx) {
```
ogaca-dd, line 90: Can you explain when and by who `runFunc` will be called?

components/kubernetes/component.go
```diff
@@ -0,0 +1,99 @@
...
+//	kubeCluster := newKindCluster()
+//	kubeCuster.Export(e, suite.cluster)
+//}
+
+// func deserialize() {
```
ogaca-dd, line 95: Can you explain when and by who `deserialize` will be called?

components/kubernetes/component.go
pducolin: If I understand it well, a client in `datadog-agent` would get a specific `test-infra-definitions` deserialiser, and would call the `deserialise` at init. What could be confusing is how to keep `Export`, but is defined thanks to the environment that exists only in a pulumi context. Sane fir `Import`, it can be called only with an UpResult, so that would be called by the stack definition. 

Each client would require to have a `Deserialiser` component, and then the environment initialiser would call each component `Import`.

PR #245, opened 2023-07-12
[task] use account wrapper in output commands in ecs and eks

pducolin:
What does this PR do?
---------------------

Use account from local profile to build the helper command to reach eks and ecs infrastructures after an environment is created on an account. 

-------------------

eks, ecs

Motivation
----------

The current command copied to the clipboard is using `sandbox` hardcoded account

Additional Notes
----------------


ogaca-dd (approved):

pducolin (commented):
Oops, thank you for the fix !

PR #260, opened 2023-07-21
Add option to force pulumi destroy

chouetz:
What does this PR do?
---------------------
Add a `-y|--yes` option to `inv destroy-xxx` entrypoints. My initial goal was to add it only for `destroy-vm` and it can maybe be useful for other scenarii

-------------------
all:
- docker
- ecs
- eks
- vm

Motivation
----------
I use only one stack at a time and as the choice selection is not immediate (but after the preview computation) I often forget to do the actual deletion

Additional Notes
----------------
I was wondering if we could even couple the `--yes` with `--skip-preview` as my use case is really: I have started one vm and I want to flush it


pducolin:
I agree with having `--force-yes` skipping the preview, I also fall in the cli waiting for my confirmation when what I want to do is just destroy what I just created

tasks/vm.py
```diff
@@ -80,16 +80,14 @@ def _show_connection_message(ctx: Context, full_stack_name: str):
     )
 
 
-@task(
-    help={
-        "stack_name": doc.stack_name,
-    }
-)
-def destroy_vm(ctx: Context, stack_name: Optional[str] = None):
+@task(help={"stack_name": doc.stack_name, "yes": doc.yes})
```
pducolin, line 83: What about using `force-yes` instead ?
**💬 suggestion**
```suggestion
@task(help={"stack_name": doc.stack_name, "force_yes": doc.yes})
```

pducolin (approved):
Thank you for the contribution, this is helpful !

We need to work on the python linter here 😵‍💫 we have an open task for it

PR #261, opened 2023-07-21
Add same linters as datadog-agent

chouetz:
What does this PR do?
---------------------
Add same python linters as datadog-agent

-------------------
N/A

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,7 +36,7 @@ jobs:
 
           # Optional: show only new issues if it's a pull request. The default value is `false`.
           only-new-issues: true
-          
+
           # Optional:The mode to install golangci-lint. It can be 'binary' or 'goinstall'.
           install-mode: "goinstall"
       - name: Check tidy
@@ -52,4 +52,17 @@ jobs:
         run: pip3 install pyright
       - name: Run pyright
         run: pyright tasks
-  
+  lint-python:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install lint python requirements
+        run: pip3 install -r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/circleci/requirements.txt
+      - name: Run linters
+        run: |
+          flake8 .
+          black --check --diff .
+          isort --check-only --diff .
+          vulture --ignore-decorators @task --ignore-names 'test_*,Test*' tasks
+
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+
+[tool.black]
+include = '\.pyi?$'
+line-length = 120
+py36 = false
+skip-string-normalization = true
+exclude = '''
+(
+  /(
+    | \.git
```
chouetz, line 14: not sure the below list is ok

pducolin: I think we can instead include only `'tasks` folder, we have python code only there

.github/workflows/lint.yml
```diff
@@ -52,4 +52,17 @@ jobs:
...
       - name: Run pyright
         run: pyright tasks
-  
+  lint-python:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install lint python requirements
+        run: pip3 install -r https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/circleci/requirements.txt
```
pducolin, line 61: Local requirements.txt might be enough

pducolin (approved):
Too good !!!

pyproject.toml
chouetz: I launch linters only on `tasks` repo, so no need to `exclude/include`

setup.cfg
```diff
@@ -0,0 +1,11 @@
...
+select = B,C,E,F,FS,G,U,W,B001,B003,B006,B007,B301,B305,B306,B902
+# Ignore:
+# - black-incompatible options: E203
+# - bugbear overlap: E722
+# - style options: W2,W3,W50,E111,E114,E117,E12,E2,E3,E5,E74
+# - unused argument with underscore: U101
+# - Unnecessary dict call: C408
+# - f-string missing prefix FS003
+ignore = E203,W2,W3,W50,E111,E114,E117,E12,E2,E3,E5,E74,E722,U101,C408,FS003
+exclude = .git,.github,.circleci,docs,google-marketplace,pkg-config,releasenotes,venv,venv3
```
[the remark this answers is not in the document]
chouetz: Same as previous

.github/workflows/lint.yml
chouetz: I used this file because it's the one that lists the required linters used in datadog-agent, it will allows remaining aligned in terms of configuration, wdyt?

PR #265, opened 2023-07-25
Use newer Fedora AMI that works

KevinFairise2:
What does this PR do?
---------------------

Change the AMI used for creating a Fedora VM

-------------------

VM scenario
Motivation
----------

Be able to create Fedora VM

Additional Notes
----------------


Pythyu (approved):
LGTM !

PR #266, opened 2023-07-25
Fix CPU configuration and downloading with multiple VMSets

usamasaqib:
What does this PR do?
---------------------
This PR performs two fixes.

1. Fix downloading of qcow2 images from S3 with multiple VMSets.
2. Remove hardcoded configuration of CPUs, so that libvirt can detect the configuration best suited.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/microVMs/microvms/libvirt.go
+++ b/scenarios/aws/microVMs/microvms/libvirt.go
@@ -82,14 +82,19 @@ type VMCollection struct {
 
 func (vm *VMCollection) SetupCollectionFilesystems(depends []pulumi.Resource) ([]pulumi.Resource, error) {
 	var waitFor []pulumi.Resource
+	var err error
 
+	var libvirtPoolReady []pulumi.Resource
 	if vm.instance.Arch != LocalVMSet {
-		libvirtPoolReady, err := setupRemoteLibvirtPool(vm.pool, vm.instance.runner, depends)
-		if err != nil {
-			return []pulumi.Resource{}, err
-		}
-		depends = append(depends, libvirtPoolReady...)
+		libvirtPoolReady, err = setupRemoteLibvirtPool(vm.pool, vm.instance.runner, depends)
+	} else {
+		libvirtPoolReady, err = setupLocalLibvirtPool(vm.instance.e.Ctx, vm.libvirtProvider, vm.pool, depends)
+
 	}
+	if err != nil {
+		return []pulumi.Resource{}, err
+	}
+	depends = append(depends, libvirtPoolReady...)
 
 	for _, set := range vm.vmsets {
 		fs, err := newLibvirtFS(vm.instance.e.Ctx, &set, vm.pool)
--- a/scenarios/aws/microVMs/microvms/libvirt_fs.go
+++ b/scenarios/aws/microVMs/microvms/libvirt_fs.go
@@ -38,6 +38,7 @@ type LibvirtFilesystem struct {
 	pool          *LibvirtPool
 	images        []*filesystemImage
 	baseVolumeMap map[string]*filesystemImage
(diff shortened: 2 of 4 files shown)


scenarios/aws/microVMs/microvms/libvirt.go
```diff
@@ -82,14 +82,19 @@ type VMCollection struct {
...
-			return []pulumi.Resource{}, err
-		}
-		depends = append(depends, libvirtPoolReady...)
+		libvirtPoolReady, err = setupRemoteLibvirtPool(vm.pool, vm.instance.runner, depends)
+	} else {
+		libvirtPoolReady, err = setupLocalLibvirtPool(vm.instance.e.Ctx, vm.libvirtProvider, vm.pool, depends)
+
+	}
+	if err != nil {
+		return []pulumi.Resource{}, err
```
val06, line 95: why not `return nil, err` ?

scenarios/aws/microVMs/microvms/libvirt_fs.go
```diff
@@ -374,18 +376,21 @@ func setupRemoteLibvirtFilesystem(fs *LibvirtFilesystem, runner *Runner, depends
...
+func setupLocalLibvirtPool(ctx *pulumi.Context, provider *libvirt.Provider, pool *LibvirtPool, depends []pulumi.Resource) ([]pulumi.Resource, error) {
+	poolReady, err := libvirt.NewPool(ctx, "create-libvirt-pool", &libvirt.PoolArgs{
 		Type: pulumi.String("dir"),
-		Name: pulumi.String(fs.pool.poolName),
-		Path: pulumi.String(generatePoolPath(fs.pool.poolName)),
+		Name: pulumi.String(pool.poolName),
+		Path: pulumi.String(generatePoolPath(pool.poolName)),
 	}, pulumi.Provider(provider), pulumi.DependsOn(depends))
 	if err != nil {
 		return []pulumi.Resource{}, err
```
val06, line 386: `return nil, err`

val06 (changes requested):
cred

scenarios/aws/microVMs/microvms/libvirt.go
usamasaqib: No good reason I guess. I think I misunderstood that empty slice == nil in pretty much all aspects in golang.

I would like to keep these as-is for now. And then make a seperate PR to change all instances of this behavior in the code, for consistency.

scenarios/aws/microVMs/microvms/libvirt_fs.go
usamasaqib: See above.

brycekahle (approved):

scenarios/aws/microVMs/microvms/libvirt_fs.go
```diff
@@ -220,11 +223,12 @@ func downloadRootfs(fs *LibvirtFilesystem, runner *Runner, depends []pulumi.Reso
 	}
 
 	if webDownload {
+		configPath := fmt.Sprintf("/tmp/aria2-%s.config", fs.fsNamer.ResourceName("aria2c"))
```
brycekahle, line 226:
```suggestion
		configPath := filepath.Join(os.TempDir(), fmt.Sprintf("aria2-%s.config", fs.fsNamer.ResourceName("aria2c")))
```

usamasaqib: `os.TempDir()` would return the default temporary directory on the system where pulumi is running.
On the otherhand, `configPath` targets the temp directory on the remote host where we are running the commands.
Given this, I think it may be better to keep `/tmp` as hardcoded.

PR #268, opened 2023-07-27
Remove entrypoint from Dockerfile

KevinFairise2:
What does this PR do?
---------------------

Remove ENTRYPOINT instruction from e2e runner Dockerfile

-------------------

Motivation
----------

We do not need it, and CI on datadog-agent is working only because on Kubernetes runners ENTRYPOINT is overriden.
By removing this we could use the image in CI with other runners.

Additional Notes
----------------


pducolin (approved):
This looks fair, did you try building an image from this dockerfile and using it in a custom CI job ?

vboulineau:
Note that isn't necessary as you can override `entrypoint` in Gitlab as well: https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#override-the-entrypoint-of-an-image

pducolin:
> Note that isn't necessary as you can override `entrypoint` in Gitlab as well: https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#override-the-entrypoint-of-an-image
> 
> Also removing `entrypoint` will make the image default to `cmd` which is set by parent image to `python3`, so does not really change your problem.

What we want then is probably using `bash` as default `entrypoint`

PR #283, opened 2023-08-07
Add CI_PIPELINE_ID in env var tags, to automatically tags re sources with pipeline id

KevinFairise2:
What does this PR do?
---------------------

It includes `CI_PIPELINE_ID` in the list of env var that will be used as tags for resources. So resources are automatically tagged when running in CI.

-------------------

Motivation
----------

To be able to link resources with CI pipelines. Which is convenient to create a clean up job for example

Additional Notes
----------------

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -115,7 +115,7 @@ func (e *CommonEnvironment) ExtraResourcesTags() map[string]string {
 
 func EnvVariableResourceTags() map[string]string {
 	tags := map[string]string{}
-	lookupVars := []string{"TEAM", "PIPELINE_ID"}
+	lookupVars := []string{"TEAM", "CI_PIPELINE_ID"}
 	for _, varName := range lookupVars {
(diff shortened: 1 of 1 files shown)


ogaca-dd (approved):

common/config/environment.go
```diff
@@ -115,7 +115,7 @@ func (e *CommonEnvironment) ExtraResourcesTags() map[string]string {
 
 func EnvVariableResourceTags() map[string]string {
 	tags := map[string]string{}
-	lookupVars := []string{"TEAM", "PIPELINE_ID"}
+	lookupVars := []string{"TEAM", "PIPELINE_ID", "CI_PIPELINE_ID"}
```
pducolin, line 118: @usamasaqib we can deprecate `PIPELINE_ID` in ebpf tests, `CI_PIPELINE_ID` is defined out of the box by gitlab with the same value

PR #289, opened 2023-08-09
Add agentparam.WithFile() interface for adding files to VM stack def

jeremy-hanna:
What does this PR do?
---------------------

A couple instances have come up where we need to write an arbitrary file to the test VM:
* [`hostname` from a file](https://github.com/DataDog/datadog-agent/blob/main/test/new-e2e/agent-subcommands/hostname_ec2_test.go#L47) 
* [flare included configuration](https://github.com/DataDog/datadog-agent/pull/18562/files#diff-abf69163194531431407d7ef6b1b41e73b779ccfe287dcc1d2c013a34d7d2641R75)
* [`secret` e2e test with backend executable](https://github.com/DataDog/datadog-agent/pull/18771/files#diff-2e3065c0eee76814c7ed58ba65f14a068e7d4f9d5bf16199b550cf32e2a0933fR52)

This PR adds a new interface to the agent stack definition similar to `WithIntegration()` but with an absolute path supplied.

-------------------

Motivation
----------

Additional Notes
----------------


components/datadog/agent/install.go
```diff
@@ -115,27 +116,36 @@ func updateAgentConfig(
...
-	for folderName, content := range integrations {
+	for filePath, content := range files {
 		var err error
-		folderPath := path.Join(configFolder, "conf.d", folderName)
-		confPath := path.Join(folderPath, "conf.yaml")
+		var fullPath string
+		useSudo := false
+		// filePath is absolute path from params.WithFile but relative from params.WithIntegration
+		if !strings.HasPrefix(filePath, "/") && !strings.HasPrefix(filePath, "C:\\") {
+			fullPath = path.Join(configFolder, filePath)
```
jeremy-hanna, line 130: I was hoping to move all this path building to `agentparams.WithIntegration(...)` but the param shouldn't have context around the os so os.GetAgentConfigFolder() isn't known at that time

components/datadog/agentparams/params.go
```diff
@@ -117,7 +120,19 @@ func WithAgentConfig(config string) func(*Params) error {
 // WithIntegration adds the configuration for an integration.
 func WithIntegration(folderName string, content string) func(*Params) error {
 	return func(p *Params) error {
-		p.Integrations[folderName] = content
+		confPath := path.Join("conf.d", folderName, "conf.yaml")
+		p.Integrations[confPath] = content
```
jeremy-hanna, line 124: renaming this to `p.File` might be a bit more correct now, I didn't want to do that until I got some feedback however

components/datadog/agentparams/params.go
```diff
@@ -117,7 +120,19 @@ func WithAgentConfig(config string) func(*Params) error {
...
+		return nil
+	}
+}
+
+// WithFile adds a file to the install with the contents at the given path
+func WithFile(filePath string, content string) func(*Params) error {
+	return func(p *Params) error {
+		if !strings.HasPrefix(filePath, "/") && !strings.HasPrefix(filePath, "C:\\") {
```
jeremy-hanna, line 132: I'm not sure if `C:\` is the correct windows install partition for the agent but I assume so?

components/datadog/agent/install.go
ogaca-dd: You can keep both `Integrations map[string]string` and `Files map[string]string`. Before calling this function you can fill `Files` with `Integrations` as you have access to `os.GetAgentConfigFolder()`. It should allow to remove this code. What do you think?

components/datadog/agentparams/params.go
```diff
@@ -117,7 +120,19 @@ func WithAgentConfig(config string) func(*Params) error {
...
 	return func(p *Params) error {
-		p.Integrations[folderName] = content
+		confPath := path.Join("conf.d", folderName, "conf.yaml")
+		p.Files[confPath] = content
+		return nil
+	}
+}
+
+// WithFile adds a file to the install with the contents at the given path
+func WithFile(filePath string, content string) func(*Params) error {
```
ogaca-dd, line 130: What do you think about reflecting that `filePath` is absolute in its naming (like `absolutePath`)?

components/datadog/agentparams/params.go
```diff
@@ -117,7 +120,19 @@ func WithAgentConfig(config string) func(*Params) error {
...
 func WithIntegration(folderName string, content string) func(*Params) error {
 	return func(p *Params) error {
-		p.Integrations[folderName] = content
+		confPath := path.Join("conf.d", folderName, "conf.yaml")
+		p.Files[confPath] = content
+		return nil
+	}
+}
+
+// WithFile adds a file to the install with the contents at the given path
```
ogaca-dd, line 129: What do you think about having `WithFilemanager`? `WithFile` could use `WithFilemanager` for its implementation.
`FileManager` supports absolute paths, relative paths, embedded paths and `sudo` and I am afraid our user will try to convert everything to absolute paths to use `WithFile` instead of using existing function. What do you think?
Note: Can be done in a separate PR.

ogaca-dd (commented):
Great PR!

components/datadog/agentparams/params.go
jeremy-hanna: 👍 awesome! good catch

jeremy-hanna: hmm on closer look I don't think I can use this as it references the runtime.GOOS to determine what is a valid path for the runtime architecture but we are referencing a VM that could be different from the CI box the test is running on

components/datadog/agent/install.go
```diff
@@ -115,27 +116,36 @@ func updateAgentConfig(
...
 	lastCommand *remote.Command) (*remote.Command, string, error) {
 	configFolder := os.GetAgentConfigFolder()
 	var parts []string
-	for folderName, content := range integrations {
+	for filePath, content := range files {
 		var err error
-		folderPath := path.Join(configFolder, "conf.d", folderName)
-		confPath := path.Join(folderPath, "conf.yaml")
+		var fullPath string
+		useSudo := false
```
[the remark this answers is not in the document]
ogaca-dd: I think we can start thing simple and not provide a way to handle sudo / not sudo as long.

components/datadog/agentparams/params.go
ogaca-dd: Indeed you cannot use `IsAbs`. I didn't realise the path was the remote path.

components/datadog/agent/install.go
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/54784169e31df63e105a4f7862d913f738dc7b62

components/datadog/agentparams/params.go
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/225fcde25381273527fca1578144f561a0c25775

components/datadog/agentparams/params.go
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/febd9a54ac2670ee61776df1c4a658bc9857d48c

components/datadog/agentparams/params.go
```diff
@@ -117,7 +122,19 @@ func WithAgentConfig(config string) func(*Params) error {
...
+		confPath := path.Join("conf.d", folderName, "conf.yaml")
+		p.Integrations[confPath] = content
+		return nil
+	}
+}
+
+// WithFile adds a file to the install with the contents at the given path
+func WithFile(absolutePath string, content string) func(*Params) error {
+	return func(p *Params) error {
+		if !strings.HasPrefix(absolutePath, "/") && !strings.HasPrefix(absolutePath, "C:\\") {
```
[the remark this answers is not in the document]
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/174944878f9a61e2b0daf992f08af1c8baac8766 (went with error checking in installer)

components/datadog/agent/install.go
```diff
@@ -109,29 +109,52 @@ func updateAgentConfig(
...
 		}
 		lastCommand, err = fileManager.CopyInlineFile(
 			pulumi.String(content),
-			confPath, true, utils.PulumiDependsOn(lastCommand))
+			fullPath, useSudo, utils.PulumiDependsOn(lastCommand))
 		if err != nil {
 			return nil, "", err
 		}
-		parts = append(parts, folderName, content)
+		parts = append(parts, fullPath, content)
```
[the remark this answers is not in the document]
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/174944878f9a61e2b0daf992f08af1c8baac8766

components/datadog/agentparams/params.go
```diff
@@ -117,7 +122,19 @@ func WithAgentConfig(config string) func(*Params) error {
...
 func WithIntegration(folderName string, content string) func(*Params) error {
 	return func(p *Params) error {
-		p.Integrations[folderName] = content
+		confPath := path.Join("conf.d", folderName, "conf.yaml")
+		p.Integrations[confPath] = content
+		return nil
+	}
+}
+
+// WithFile adds a file to the install with the contents at the given path
```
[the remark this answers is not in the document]
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/0417a0643c8446c5ab364f505783a4dbe1561fc2

components/datadog/agent/install.go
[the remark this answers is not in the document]
jeremy-hanna: Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/0417a0643c8446c5ab364f505783a4dbe1561fc2

jeremy-hanna:
https://github.com/DataDog/test-infra-definitions/pull/305 fixes the lint-go failure

components/os/unix.go
```diff
@@ -22,6 +23,10 @@ func (u *Unix) GetDefaultInstanceType(arch Architecture) string {
 }
 func (*Unix) GetAgentConfigFolder() string { return "/etc/datadog-agent" }
 
+func (*Unix) CheckIsAbsPath(path string) bool {
```
ogaca-dd, line 26: I don't think it works if you create a Ubuntu VM from a Windows laptop as `filepath.IsAbs` would return if the path is a Windows absolute path.

components/datadog/agentparams/params.go
```diff
@@ -27,20 +27,26 @@ import (
 //   - [WithLogs]
 //
 // [Functional options pattern]: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
+
+type FileDefinition struct {
```
ogaca-dd, line 31: 👍

components/os/unix.go
jeremy-hanna: 👍 

Fixed in https://github.com/DataDog/test-infra-definitions/pull/289/commits/62f077c5fa464c54e484f2f229435941058b6158

ogaca-dd (approved):

PR #309, opened 2023-08-23
Add system-probe and security-agent config

mftoure:
What does this PR do?
---------------------
This PR adds the ability to set configurations for system-probe and the security-agent using `WithSystemProbeConfig` and `WithSecurityAgentConfig`

### DIFF
--- a/components/datadog/agent/install.go
+++ b/components/datadog/agent/install.go
@@ -50,7 +50,7 @@ func NewInstaller(vm vm.VM, options ...agentparams.Option) (*Installer, error) {
 	}
 
 	var updateConfigTrigger pulumi.StringInput
-	lastCommand, updateConfigTrigger, err = updateAgentConfig(
+	lastCommand, _, err = updateAgentConfig(
 		vm.GetFileManager(),
 		env,
 		params.AgentConfig,
--- a/components/datadog/agent/install.go
+++ b/components/datadog/agent/install.go
@@ -49,8 +49,8 @@ func NewInstaller(vm vm.VM, options ...agentparams.Option) (*Installer, error) {
 		return nil, err
 	}
 
-	var updateConfigTrigger pulumi.StringInput
-	lastCommand, _, err = updateAgentConfig(
+	var updateAgentConfigTrigger, updateSystemConfigTrigger, updateSecurityAgentConfigTrigger pulumi.StringInput
+	lastCommand, updateAgentConfigTrigger, err = updateAgentConfig(
 		vm.GetFileManager(),
 		env,
 		params.AgentConfig,
--- a/components/datadog/agent/install.go
+++ b/components/datadog/agent/install.go
@@ -49,16 +49,17 @@ func NewInstaller(vm vm.VM, options ...agentparams.Option) (*Installer, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


components/datadog/agent/install.go
```diff
@@ -109,6 +127,56 @@ func updateAgentConfig(
...
 
+func updateSystemProbeConfig(
+	fileManager *command.FileManager,
+	systemProbeConfig string,
+	os os.OS,
+	lastCommand *remote.Command) (*remote.Command, pulumi.StringInput, error) {
+
+	if systemProbeConfig == "" {
+		// don't update config
+		return lastCommand, nil, nil
```
pducolin, line 138: **🔨 warning**
There is one particular situation where you might still want to update the config, being when you first write a config and then you want to clean it up coming back to the default empty config.  In other words, you probably want to write and empty config if you are called with `WithSystemProbeConfig('')`, while you don't want to touch the file if you are called without any mention of `WithSystemProbeConfig`

I suggest writing an empty file here, if it does no arm to have an empty config file always.

pducolin: cc @ogaca-dd as this was a recent bug you found on the agent config

pducolin (commented):
Thank you for your contribution !

pducolin (approved):

components/datadog/agent/install.go
```diff
@@ -58,6 +58,24 @@ func NewInstaller(vm vm.VM, options ...agentparams.Option) (*Installer, error) {
...
+	lastCommand, updateSystemConfigTrigger, err = updateSystemProbeConfig(
+		vm.GetFileManager(),
+		params.SystemProbeConfig,
+		os,
+		lastCommand)
+	if err != nil {
+		return nil, err
+	}
+
+	lastCommand, updateSecurityAgentConfigTrigger, err = updateSecurityAgentConfig(
```
ogaca-dd, line 70: By extracting `params.ExtraAgentConfig` logic  from `updateSystemProbeConfig`, it would be possible to have something like:
```go
for _, input := range struct{ content: string, target: string } {
   { "datadog.yaml",  wholeConfigContent},
   { "system-probe.yaml",  params.SystemProbeConfig},
 { "security-agent.yaml",  params.SecurityAgentConfig} } {
   updateConfig(...)
}

Comment (mftoure) [reply components/datadog/agent/install.go:70]: Hello Oliver, could you elaborate on this ?

Comment (ogaca-dd) [reply components/datadog/agent/install.go:70]: Note: We discussed it offline.

**Review by nplanel — APPROVED** [review]

File: components/datadog/agent/install.go [hunk 3]
```diff
@@ -49,16 +49,17 @@ func NewInstaller(vm vm.VM, options ...agentparams.Option) (*Installer, error) {
...
-		lastCommand)
-	if err != nil {
-		return nil, err
+	fileChangeTriggers := make(map[string]pulumi.StringInput)
+	var trigger pulumi.StringInput
+	for _, input := range []struct{ path, content string }{
+		{"datadog.yaml", params.AgentConfig},
+		{"system-probe.yaml", params.SystemProbeConfig},
+		{"security-agent.yaml", params.SecurityAgentConfig}} {
+		lastCommand, trigger, err = updateConfig(vm.GetFileManager(), input.path, input.content, params.ExtraAgentConfig, env, os, lastCommand)
```
pducolin, line 58: **👏 praise**
Nice !

PR #310, opened 2023-08-23
Temporarly disable integration test locally to avoid overwritting local config

KevinFairise2:
What does this PR do?
---------------------

Temp fix to avoid overwriting local config when running go test locally

-------------------

Motivation
----------

Not break config while the issue is fixed with a proper fix

Additional Notes
----------------


pducolin (approved):
Perfect ! Thank you

PR #324, opened 2023-09-01
Option to disable the loadbalancer

Pythyu:
What does this PR do?
---------------------

Add the `--use-loadbalancer` option to `inv create-vm`

The loadbalancer is now disabled by default on VM and can be enabled if needed.
The option will only have an effect if `--use-fakeintake` is used as well.

-------------------

Every VM instantiation using a fakeintake

Motivation
----------

Fakeintake deployment can take up to 10min with a loadbalancer.

Additional Notes
----------------


components/datadog/fakeintake/ecsFargate.go
```diff
@@ -105,11 +121,12 @@ func NewECSFargateInstance(e aws.Environment) (*Instance, error) {
 	}, opts...); err != nil {
 		return nil, err
 	}
-
-	if err := e.Ctx.RegisterResourceOutputs(instance, pulumi.Map{
-		"host": alb.LoadBalancer.DnsName(),
-	}); err != nil {
-		return nil, err
+	if os.Getenv("DD_DISABLE_LB") == "" {
```
pgimalac, line 124: **💬 suggestion:** You might want to use LookupEnv instead of GetEnv to make a difference between unset variable and variable set to the empty string (unless the variable has to be explicitly set to a non-empty value)
https://pkg.go.dev/os#LookupEnv

components/datadog/apps/nginx/ecs.go
```diff
@@ -2,31 +2,56 @@ package nginx
...
-func EcsAppDefinition(e aws.Environment, clusterArn pulumi.StringInput, opts ...pulumi.ResourceOption) (*EcsComponent, error) {
-	namer := e.Namer.WithPrefix("nginx")
-	opts = append(opts, e.WithProviders(config.ProviderAWS, config.ProviderAWSX))
-
-	ecsComponent := &EcsComponent{}
-	if err := e.Ctx.RegisterComponentResource("dd:apps", namer.ResourceName("grp"), ecsComponent, opts...); err != nil {
-		return nil, err
-	}
+func GetEcsTargetGroupAndDns(e aws.Environment, namer namer.Namer, opts []pulumi.ResourceOption) (pulumi.StringOutput, pulumi.StringOutput, error) {
+	if os.Getenv("DD_DISABLE_LB") != "" {
```
vboulineau, line 21: Why not using a standard parameter instead of hardcoding an env var?

vboulineau (changes requested):
Goal of the PR is not very clear, code is duplicated between `nginx` and `redis` and I don't understand why code in `components/datadog/fakeintake/ecsFargate.go` was moved?

Pythyu:
Circle dependencies issue, but I'm going to move it back into multiple files into `components/datadog/apps`

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -0,0 +1,202 @@
...
+				ContainerPort:  pulumi.Int(port),
+				TargetGroupArn: alb.DefaultTargetGroup.Arn(),
+			},
+		}
+	} else {
+		var FargateErr error
+		instance.Host, FargateErr = FargateServiceFakeintake(e)
+		if FargateErr != nil {
+			return nil, FargateErr
+		}
```
pducolin, line 79: **💬 suggestion**
No need to create a variable here
```suggestion
		instance.Host, err = FargateServiceFakeintake(e)
		if err != nil {
			return nil, err
		}
```

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -0,0 +1,202 @@
...
+				Protocol:      pulumi.StringPtr("tcp"),
+			},
+		},
+		VolumesFrom: ecs.TaskDefinitionVolumeFromArray{},
+	}
+}
+
+// FargateServiceFakeintake deploys one fakeintake container to a dedicated Fargate cluster
+// Hardcoded on sandbox
+func FargateServiceFakeintake(e aws.Environment) (ipAddress pulumi.StringOutput, err error) {
```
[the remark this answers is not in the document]
pducolin: This could be unexported

tasks/vm.py
```diff
@@ -50,6 +52,7 @@ def create_vm(
     extra_flags = {}
     os_family, os_arch = _get_os_information(ctx, os_family, architecture, ami_id)
     extra_flags["ddinfra:osFamily"] = os_family
+    extra_flags["ddinfra:aws/defaultFargateLoadBalancer"] = use_loadBalancer
```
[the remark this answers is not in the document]
pducolin: I would define it in `ddinfra: deployFakeintakeWithLoadBalancer` as it is not strictly related to aws, you can move its definition to `common/config/environment.go`, you can look at `DDInfraOSArchitecture` as an example

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -0,0 +1,201 @@
...
+				Protocol:      pulumi.StringPtr("tcp"),
+			},
+		},
+		VolumesFrom: ecs.TaskDefinitionVolumeFromArray{},
+	}
+}
+
+// FargateServiceFakeintakeWithoutLoadBalancer deploys one fakeintake container to a dedicated Fargate cluster
+// Hardcoded on sandbox
+func FargateServiceFakeintakeWithoutLoadBalancer(e aws.Environment) (ipAddress pulumi.StringOutput, err error) {
```
pducolin, line 173: **❓ question**
Can this be private ?

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -0,0 +1,201 @@
...
+			"host": alb.LoadBalancer.DnsName(),
+		}); err != nil {
+			return nil, err
+		}
+	}
+
+	return instance, nil
+}
+
+func FargateLinuxTaskDefinition(e aws.Environment, name string) (*ecs.FargateTaskDefinition, error) {
```
pducolin, line 136: **❓ question**
Looks like this could be private too

resources/aws/environment.go
```diff
@@ -166,6 +166,10 @@ func (e *Environment) DefaultShutdownBehavior() string {
 	return e.GetStringWithDefault(e.InfraConfig, DDInfraDefaultShutdownBehavior, e.envDefault.ddInfra.defaultShutdownBehavior)
 }
 
+func (e *Environment) DefaultFargateLoadBalancer() bool {
+	return e.GetBoolWithDefault(e.InfraConfig, DDInfraEcsFargateCapacityProvider, e.envDefault.ddInfra.defaultFargateLoadBalancer)
+}
+
```
pducolin, line 172: **💬 suggestion**
This should be defined in `common/config/environments.go` as `InfraShouldDeployFakeintakeWithLB`

scenarios/aws/fakeintake.go
```diff
@@ -1,12 +1,13 @@
 package aws
 
 import (
+	fakeintake "github.com/DataDog/test-infra-definitions/components/datadog/apps/fakeintake"
 	ddfakeintake "github.com/DataDog/test-infra-definitions/components/datadog/fakeintake"
```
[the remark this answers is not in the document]
Pythyu: can't because of circle dependency

pducolin (approved):
Love this, thank you !

PR #348, opened 2023-09-15
Add new console type.

usamasaqib:
What does this PR do?
---------------------
This PR adds config option to use a file for console output.

-------------------

Motivation
----------
Debug VM startup issues in the CI.

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/microVMs/microvms/resources/resources.go
+++ b/scenarios/aws/microVMs/microvms/resources/resources.go
@@ -99,11 +99,14 @@ type RecipeLibvirtVolumeArgs struct {
 	XMLArgs  map[string]pulumi.StringInput
 }
 
+func GetConsolePath(domainName string) string {
+	return fmt.Sprintf("/tmp/ddvm-%s.log", domainName)
+}
+
 func setupConsole(consoleType, domainName string) (libvirt.DomainConsoleArgs, error) {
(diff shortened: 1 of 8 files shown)


scenarios/aws/microVMs/microvms/resources/resources.go
```diff
@@ -22,8 +23,23 @@ const (
 	PoolPath      = "poolPath"
 	Nvram         = "nvram"
 	Efi           = "efi"
+	fileConsole   = "file"
+	ptyConsole    = "pty"
```
pducolin, line 27: **💬 suggestion**
Should console type be a subset of `string` that allows only `"file|pty"` values ?

pducolin (approved):

scenarios/aws/microVMs/microvms/resources/resources.go
usamasaqib: How would I implement this?

brycekahle: Using a go "enum":
```
type ConsoleType string
const (
    FileConsole ConsoleType = "file"
    PTYConsole ConsoleType = "pty"
)
```

brycekahle (approved):
I think creating a `ConsoleType` enum could help, but isn't required.

PR #355, opened 2023-09-20
[dockerfile] install helm from binaries

pducolin:
What does this PR do?
---------------------

Install helm in pulumi-runner image using binaries instead of the script fetched from github

-------------------

None 

Motivation
----------

This will prevent the github error rate we get when installing using the install script. It also prevents flakiness pinning helm version

Additional Notes
----------------

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,16 +21,16 @@ RUN apt-get update -y && \
   wget \
   unzip && \
   # Get all of the signatures we need all at once.
-  curl -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key  | apt-key add - && \
-  curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg              | apt-key add - && \
-  curl -fsSL https://download.docker.com/linux/debian/gpg          | apt-key add - && \
-  curl -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
-  curl -fsSL https://packages.microsoft.com/keys/microsoft.asc     | apt-key add - && \
+  curl --retry 10 -fsSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key  | apt-key add - && \
+  curl --retry 10 -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg              | apt-key add - && \
+  curl --retry 10 -fsSL https://download.docker.com/linux/debian/gpg          | apt-key add - && \
+  curl --retry 10 -fsSL https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
+  curl --retry 10 -fsSL https://packages.microsoft.com/keys/microsoft.asc     | apt-key add - && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -68,7 +70,12 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 # (e.g. in GitHub actions where $HOME points to /github/home).
 ENV XDG_CONFIG_HOME=/root/.config
 ENV XDG_CACHE_HOME=/root/.cache
-RUN curl -L https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash && \
+RUN curl -fsSLo /tmp/helm.tgz https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
+  echo "${HELM_SHA} /tmp/helm.tgz" | sha256sum -c - && \
+  tar -C /usr/local/helm -xzf /tmp/helm.tgz && \
+  rm /tmp/helm.tgz && \
+  export PATH="/usr/local/helm/linux-amd64:$PATH" && \
```
L3n41c, line 77:
```suggestion
  mkdir /usr/local/helm &&
  tar -C /usr/local/helm -xzf /tmp/helm.tgz --strip-components=1 linux-amd64/helm && \
  rm /tmp/helm.tgz && \
  export PATH="/usr/local/helm:$PATH" && \
```

L3n41c (approved):
Just left a small comment to:
* fix the `tar: /usr/local/helm: Cannot open: No such file or directory` error.
* skip the extraction of `README.md` and `LICENSE` file from the archive.
* remove the `/linux-amd64/` subdirectory.

Dockerfile
```diff
@@ -68,7 +70,12 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 # (e.g. in GitHub actions where $HOME points to /github/home).
 ENV XDG_CONFIG_HOME=/root/.config
 ENV XDG_CACHE_HOME=/root/.cache
-RUN curl -L https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash && \
+RUN curl -fsSLo /tmp/helm.tgz https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
```
L3n41c, line 73: In order to be more resilient to transient failures, we might be interested in “retry” options:

```suggestion
RUN curl -fsSLo --retry=10 /tmp/helm.tgz https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz && \
```

Dockerfile
pducolin: Manually committing as I included the `mkdir` in a recent forced push ✊

Dockerfile
pducolin: Should I add this retry also to go install ?

Dockerfile
vboulineau: Also should be `/usr/local/bin/helm` and hopefully drop the `export PATH`

PR #358, opened 2023-09-21
Use timeout options added to Command plugin

vboulineau:
What does this PR do?
---------------------

With `Command` plugin bumped to `0.9` we can finally control the `Dial` timeout, meaning that we won't get stuck for 2-3 minutes (OS timeout) before retrying SSH connection if the first attempt is made while SSH server is not yet ready.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,15 +101,13 @@ RUN cd /tmp/test-infra && \
 # Remove AWS-related deps as we already install AWS CLI v2
 # Remove PyYAML to workaround issues with cpython 3.0.0
 # https://github.com/yaml/pyyaml/issues/724#issuecomment-1638636728
-RUN curl --retry 10 -fsSL https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/main/requirements.txt | \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -101,15 +101,13 @@ RUN cd /tmp/test-infra && \
...
+RUN curl --retry 10 -fsSL https://raw.githubusercontent.com/DataDog/datadog-agent-buildimages/b468e3cdcbe66e8b8852a29be6c66b03bc08e03d/requirements.txt | \
   grep -ivE "boto3|botocore|awscli|urllib3|PyYAML" > requirements-agent.txt && \
   pip3 install "cython<3.0.0" && \
   pip3 install --no-build-isolation PyYAML==5.4.1 && \
   pip3 install -r requirements-agent.txt && \
   go install gotest.tools/gotestsum@latest
 
-# Install go test requirements
-RUN go install gotest.tools/gotestsum@latest
-
```
pducolin, line 113: **❓ question**
Why is this gone ?

vboulineau: It's duplicated with line `109`

pducolin (approved):

PR #368, opened 2023-09-26
Add --show-secrets to pulumi stack output

mftoure:
What does this PR do?
---------------------
This PR adds `--show-secrets` to the pulumi stack output command, so that we can view debug outputs like `host` and `user`
-------------------

Motivation
----------

Additional Notes
----------------


pducolin (changes requested):
This should not be required. We don't call `pulumi` from e2e tests, we use the automation API. The output normally contains the connection infos for the stack.

PR #370, opened 2023-09-27
Simplify FileManager API:

KevinFairise2:
What does this PR do?
---------------------

Simplify file manager API by removing the need to set a resource name

-------------------

Motivation
----------

Have a simpler API
Additional Notes
----------------

### DIFF
--- a/components/command/filemanager.go
+++ b/components/command/filemanager.go
@@ -27,23 +27,23 @@ func NewFileManager(runner *Runner) *FileManager {
 }
 
 // CreateDirectory if it does not exist
-func (fm *FileManager) CreateDirectory(name string, remotePath pulumi.StringInput, useSudo bool, opts ...pulumi.ResourceOption) (*remote.Command, error) {
-	return fm.command.CreateDirectory(fm.runner, name, remotePath, useSudo, opts...)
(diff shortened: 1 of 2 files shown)


components/command/filemanager.go
```diff
@@ -79,7 +79,7 @@ func (fm *FileManager) CopyAbsoluteFolder(absoluteFolder string, remoteFolder st
 	baseFolder := filepath.Base(absoluteFolder)
 	rootWithoutBase := absoluteFolder[:len(absoluteFolder)-len(baseFolder)]
 	// Use remoteFolder as `absoluteFolder` may be a random file path that is different for each run.
```
pducolin, line 81: We can remove this comment, underlying we keep the same behaviour 
```suggestion
```

pducolin (approved):
Tested on a temporary scenario, it works fine for both temp and home directories !

vboulineau:
This removes the ability to create folder from `Output`, that's why the simplification was not made in the past.

ogaca-dd (approved):

pducolin:
> This removes the ability to create folder from Output, that's why the simplification was not made in the past.

After discussion, we'll add a `CreateFolderWithPulumiString` for such a use case - note so far it is not used.

PR #371, opened 2023-09-27
Add notification after stack is created: MacOS, copy to clipboard now requires to press a key

KevinFairise2:
What does this PR do?
---------------------

Add some desktop notification once the stack is created.
User now needs to press a key before the connection command is copied to clipboard, to avoid unexpected things in the clipboard

-------------------

Motivation
----------

Nice to have notification when the stack can take a long time to create

Additional Notes
----------------

I created the notification on MacOS. I did not implement it for Linux and Windows since I am not able to test anything without a Linux or Windows system with a desktop

### DIFF
--- a/tasks/tool.py
+++ b/tasks/tool.py
@@ -118,7 +118,6 @@ def notify(text):
         notify_windows(text)
     notify_macos(text)
 
-
 def notify_macos(text):
     CMD = '''
     on run argv
--- a/tasks/tool.py
+++ b/tasks/tool.py
@@ -2,6 +2,7 @@ import getpass
 import json
 import pathlib
 import platform
+import subprocess
 from io import StringIO
 from typing import Any, List, Optional
 
@@ -110,6 +111,29 @@ def get_image_description(ctx: Context, ami_id: str) -> Any:
     else:
         return result["Images"][0]
 
+def notify(text):
+    if is_linux():
+        notify_linux(text)
+    if is_windows():
+        notify_windows(text)
+    notify_macos(text)
+
+
+def notify_macos(text):
+    CMD = '''
+    on run argv
+    display notification (item 2 of argv) with title (item 1 of argv)
+    end run
+    '''
+    subprocess.call(['osascript', '-e', CMD, "test-infra-definitions", text])
+
+def notify_linux(text):
+    # TODO: Implement notification on linux. Would require linux computer (with desktop) to test
+    return
+
+def notify_windows(text):
+    # TODO: Implenent notification on windows. Would require windows computer (with desktop) to test
+    return
 
 def _get_root_path() -> str:
     folder = pathlib.Path(__file__).parent.resolve()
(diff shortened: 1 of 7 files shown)


tasks/tool.py
```diff
@@ -2,6 +2,7 @@
 import json
 import pathlib
 import platform
+import subprocess
```
pducolin, line 5: **💬 suggestion**
You should be able to use [invoke context run](https://docs.pyinvoke.org/en/stable/api/runners.html) to execute commands, we use it to call `pulumi` in [deploy](https://github.com/DataDog/test-infra-definitions/blob/main/tasks/deploy.py#L132)

tasks/tool.py
```diff
@@ -111,6 +112,33 @@ def get_image_description(ctx: Context, ami_id: str) -> Any:
...
+    subprocess.call(['osascript', '-e', CMD, "test-infra-definitions", text])
+
+
+def notify_linux():
+    # TODO: Implement notification on linux. Would require linux computer (with desktop) to test
+    return
+
+
+def notify_windows():
+    # TODO: Implenent notification on windows. Would require windows computer (with desktop) to test
```
pducolin, line 138: As a follow up, should use [win10toast](https://stackoverflow.com/questions/12575708/notification-using-python)

tasks/tool.py
```diff
@@ -111,6 +112,33 @@ def get_image_description(ctx: Context, ami_id: str) -> Any:
...
+    CMD = '''
+    on run argv
+    display notification (item 2 of argv) with title (item 1 of argv)
+    end run
+    '''
+    subprocess.call(['osascript', '-e', CMD, "test-infra-definitions", text])
+
+
+def notify_linux():
+    # TODO: Implement notification on linux. Would require linux computer (with desktop) to test
```
pducolin, line 133: As a follow up, can try with https://pypi.org/project/notify-send/ on linux

pducolin (changes requested):
Overall looks great, you should avoid using `subprocess` in favor of invoke runner `run`

tasks/tool.py
KevinFairise2: Can I create follow up cards for this? I saw that some solution exist but since I will not be able to test it I think it is not super safe to try to implement it. It could broke windows/linux users.
Or I can wrap all this code in a `try` and do not fail even if my code do not work

tasks/tool.py
KevinFairise2: Done!

tasks/tool.py
pducolin: Yes absolutely ! You can add cards mentioning this, if a linux or windows dev needs it they will have a trace to it

pducolin (approved):

pducolin:
Sorry I saw I forgot to approve !

pducolin:
**💬 suggestion**
You could add a `--no-interactive` mode to skip the notification, can help fix the integration test

PR #377, opened 2023-09-29
Fallback random version, add install workload parameter for EKS

vboulineau:
What does this PR do?
---------------------

`random 4.14` requires `Go 1.21`, which we are not using yet, making `go mod tidy` unhappy.

-------------------

Motivation
----------

Additional Notes
----------------


pducolin (approved):

pducolin:
I don't get errors at `go mod tidy`, can we check this on the CI ? Why is the CI fine ?

vboulineau:
It probably errors when Go >= 1.21 is present (which is my case locally) but it's not the case in the CI. I'd have expected a warning but apparently it just ignores it.

PR #380, opened 2023-10-06
Add parameters to set docker registry credentials

L3n41c:
What does this PR do?
---------------------

Add new parameters
* `ddagent:imagePullRegistry`
* `ddagent:imagePullUsername`
* `ddagent:imagePullPassword`

to set docker registry credentials.

-------------------

Those based on Helm deployment.

Motivation
----------

Whereas EKS nodes have, by default and out of the box, the credentials to fetch the docker images from the private ECR of the same AWS project, it’s not the case for the kind setup.
The kind testcase is currently failing because all the pods remain in `ImagePullBackoff` because they miss the credentials to fetch the agent docker images from the private ECR registry.

Additional Notes
----------------

https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/

### DIFF
--- a/components/datadog/agent/helm.go
+++ b/components/datadog/agent/helm.go
@@ -1,7 +1,8 @@
 package agent
 
 import (
-	"fmt"
+	"encoding/base64"
+	"encoding/json"
 
 	"golang.org/x/exp/maps"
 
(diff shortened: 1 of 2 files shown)


components/datadog/agent/helm.go
```diff
@@ -162,8 +193,12 @@ func NewHelmInstallation(e config.CommonEnvironment, args HelmInstallationArgs,
 	return helmComponent, nil
 }
 
-func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAgentImagePath, clusterAgentImageTag string, clusterAgentToken pulumi.StringInput) pulumi.Map {
-	return pulumi.Map{
+type HelmValues struct {
```
vboulineau, line 196: Any reason to embed? Why note `type HelmValues pulumi.Map`?

L3n41c: You’re right : there’s no reason.
I’ve switched to type alias instead.

vboulineau (approved):

PR #392, opened 2023-10-16
[task] 🤖 install pulumi plugins at setup

pducolin:
What does this PR do?
---------------------

Install pulumi plugins at `inv setup` time

-------------------

None

Motivation
----------

E2E tests in `datadog-agent` do not run invoke tasks to create resources, so do not install plugins out of the box. Adding explicit plugin install from the setup script, that can be safely re-executed as it reloads a existing local profile, plus it can be interrupted after all install steps are executed

Additional Notes
----------------

### DIFF
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -28,6 +28,8 @@ def setup(_: Context, config_path: Optional[str] = None, interactive: Optional[b
     else:
         os.system("brew install pulumi/tap/pulumi")
 
+    # install plugins
+    os.system("pulumi --non-interactive plugin install")
     # login to local stack storage
     os.system("pulumi login --local")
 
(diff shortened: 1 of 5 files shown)


tasks/setup.py
```diff
@@ -28,13 +28,16 @@ def setup(_: Context, config_path: Optional[str] = None, copy_to_clipboard: Opti
     else:
         os.system("brew install pulumi/tap/pulumi")
 
+    # install plugins
+    os.system("pulumi --non-interactive plugin install")
```
vboulineau, line 32: Note that it needs to be run with CWD == `test-infra-definitions` repo.

pducolin: Worth adding a cwd here, anyhow as the script is defined in `test-infra`, it is mostly called from there. But I know of errors triggered because they called it from somewhere else

pducolin: Also realising we did not replace os.system with `ctx.run` here, on it

pducolin: What am I saying, I should run from the `tasks` parent !

pducolin:
This is breaking the integration test, that fails locally because I get an unexpected EOF. Will come later to this.

KevinFairise2 (approved):

PR #394, opened 2023-10-16
[CONTINT-3310] Set GOMEMLIMIT to the fake intake

L3n41c:
What does this PR do?
---------------------

Set `GOMEMLIMIT` to the fake intake to ¾ of its memory limit.
As `GOMEMLIMIT` is a “soft” memory limit, it’s safer to take a security margin.

-------------------

Motivation
----------

Without it, the memory is never released. The memory consumed by the fake intake keeps on increasing forever until it is OOM killed.

Additional Notes
----------------

image


pducolin (approved):

pducolin:
Nice !!

vboulineau (approved):

PR #395, opened 2023-10-16
Add event platform endpoints to Agent config when using Fake Intake

Kaderinho:
What does this PR do?
---------------------

Add event platform endpoints to the Agent config when using the fake intake.

-------------------

I'm currently testing the diagnose command and the command is failing because the Agent is sending request to the real backend with a dummy API Key.

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agentparams/params.go
+++ b/components/datadog/agentparams/params.go
@@ -196,34 +196,29 @@ func WithTelemetry() func(*Params) error {
 // WithFakeintake installs the fake intake and configures the Agent to use it.
 func WithFakeintake(fakeintake *fakeintake.ConnectionExporter) func(*Params) error {
 	return func(p *Params) error {
-		// configure metrics and check run intake
-		args := make([]interface{}, 12) // number of %s in the configuration (horribly hardcoded)
-		for idx := range args {
-			args[idx] = fakeintake.Host
-		}
-		extraConfig := pulumi.Sprintf(`dd_url: http://%s:80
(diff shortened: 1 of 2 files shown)


components/datadog/agentparams/params.go
```diff
@@ -197,11 +197,33 @@ func WithTelemetry() func(*Params) error {
...
+network_devices.snmp_traps.forwarder.logs_dd_url: %s:80
+network_devices.snmp_traps.forwarder.logs_no_ssl: true
+network_devices.netflow.forwarder.logs_dd_url: %s:80
+network_devices.netflow.forwarder.logs_no_ssl: true
+container_lifecycle.logs_dd_url: %s:80
+container_lifecycle.logs_no_ssl: true
+container_image.logs_dd_url: %s:80
+container_image.logs_no_ssl: true
+sbom.logs_dd_url: %s:80
+sbom.logs_no_ssl: true`, args...)
```
pgimalac, line 226: `fmt.Sprinf` can use indexed arguments which would make this much more readable, and avoid the hacky arg slice at the beginning (https://go.dev/play/p/7xVTDDBUTV9)

Kaderinho: One day a wise person said `tous les héros ne portent pas de cape` and now I understand what he meant, thanks ❤️

pducolin (approved):
Approved with change suggested by @pgimalac - amazing !

PR #403, opened 2023-10-20
Add FakeIntakeParams to be able to disable LB from datadog-agent tests

KevinFairise2:
What does this PR do?
---------------------

Expose option to be able to disable Load Balancer from tests

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/fakeintake.go
+++ b/scenarios/aws/fakeintake.go
@@ -4,10 +4,11 @@ import (
 	"github.com/DataDog/test-infra-definitions/components/datadog/apps/fakeintake"
 	ddfakeintake "github.com/DataDog/test-infra-definitions/components/datadog/fakeintake"
 	resourcesAws "github.com/DataDog/test-infra-definitions/resources/aws"
+	"github.com/DataDog/test-infra-definitions/scenarios/aws/fakeintake/fakeintakeparams"
 )
 
-func NewEcsFakeintake(env resourcesAws.Environment) (*ddfakeintake.ConnectionExporter, error) {
-	fargateInstance, err := fakeintake.NewECSFargateInstance(env)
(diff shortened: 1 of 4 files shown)


scenarios/aws/fakeintake.go
```diff
@@ -4,10 +4,11 @@ import (
 	"github.com/DataDog/test-infra-definitions/components/datadog/apps/fakeintake"
 	ddfakeintake "github.com/DataDog/test-infra-definitions/components/datadog/fakeintake"
 	resourcesAws "github.com/DataDog/test-infra-definitions/resources/aws"
+	"github.com/DataDog/test-infra-definitions/scenarios/aws/fakeintake/fakeintakeparams"
 )
 
-func NewEcsFakeintake(env resourcesAws.Environment) (*ddfakeintake.ConnectionExporter, error) {
-	fargateInstance, err := fakeintake.NewECSFargateInstance(env)
+func NewEcsFakeintake(env resourcesAws.Environment, options ...fakeintakeparams.Option) (*ddfakeintake.ConnectionExporter, error) {
```
pducolin, line 10: **👏 praise**
Exactly what I was thinking ! @pgimalac Can you rebase on this change and add `WithName` option ?

pducolin (approved):

scenarios/aws/fakeintake.go
pgimalac: Haha I initially did it that way but ended up reverting to something simpler
But that's perfect, I'll rebase ! :)

KevinFairise2:
> **💬 suggestion** Can we change the integration test to create a VM without LB ?

I can for sure

KevinFairise2:
> **💬 suggestion** Can we change the integration test to create a VM without LB ?

After checking, the invoke task is configured to not use the load balancer by default. So the integration tests are actually not creating a load balancer with the VM

PR #409, opened 2023-10-25
[ASCII-749] Add WithIntakeHostname

pgimalac:
What does this PR do?
---------------------
Add a `WithIntakeHostname` agent option.

-------------------
Agent using a Fakeintake, although the change should be transparent (backward compatible).

Motivation
----------
Use a custom intake name in a test, without copy pasting the implementation of `WithFakeintake`.

Additional Notes
----------------
Implementation is exactly the same as `WithFakeintake`.

### DIFF
--- a/components/datadog/agentparams/params.go
+++ b/components/datadog/agentparams/params.go
@@ -224,14 +224,18 @@ sbom.logs_no_ssl: true`, hostname)
 	}
 }
 
-// WithIntakeName configures the agent to use the given hostname as intake
-// This option is incompatible with `WithFakeintake`.
+// WithIntakeName configures the agent to use the given hostname as intake.
+//
+// To use a fakeintake, see WithFakeintake.
+//
+// This option is overwritten by `WithFakeintake`.
 func WithIntakeHostname(hostname string) func(*Params) error {
 	return withIntakeHostname(pulumi.String(hostname))
 }
 
 // WithFakeintake installs the fake intake and configures the Agent to use it.
-// This option is incompatible with `WithIntakeHostname`.
+//
+// This option is overwritten by `WithIntakeHostname`.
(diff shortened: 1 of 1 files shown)


pducolin:
**❓ question**
Would you use it to test against an intake other than the fakeintake ? Would you use one real DD intake ?

pducolin (approved):

components/datadog/agentparams/params.go
```diff
@@ -218,12 +218,24 @@ container_lifecycle.logs_no_ssl: true
...
 sbom.logs_dd_url: %[1]s:80
-sbom.logs_no_ssl: true`, fakeintake.Host)
+sbom.logs_no_ssl: true`, hostname)
 		p.ExtraAgentConfig = append(p.ExtraAgentConfig, extraConfig)
 		return nil
 	}
 }
 
+// WithIntakeName configures the agent to use the given hostname as intake
+// This option is incompatible with `WithFakeintake`.
```
pducolin, line 228:
```suggestion
// This option is overwritten by `WithFakeintake`.
```

components/datadog/agentparams/params.go
```diff
@@ -218,12 +218,24 @@ container_lifecycle.logs_no_ssl: true
...
+func WithIntakeHostname(hostname string) func(*Params) error {
+	return withIntakeHostname(pulumi.String(hostname))
+}
+
+// WithFakeintake installs the fake intake and configures the Agent to use it.
+// This option is incompatible with `WithIntakeHostname`.
```
pducolin, line 234:
```suggestion
// This option is overwritten by `WithIntakeHostname`.
```

pgimalac:
> Would you use it to test against an intake other than the fakeintake ? Would you use one real DD intake ?

As discussed IRL this is used when you need a custom intake name when playing with DNS.
I guess it could also be used for a real DD intake, but the `site` setting handles that.

PR #410, opened 2023-10-25
[scenarios/aws/{kindvm,eks}] Add option to deploy standalone dogstatsd

davidor:
This PR adds the option to deploy standalone dogstatsd in the `aws/kind` scenario.
The dogstatsd clients already defined in `components/datadog/apps/dogstatsd` can be used to test it.

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -19,6 +19,7 @@ const (
 	DDInfraConfigNamespace     = "ddinfra"
 	DDAgentConfigNamespace     = "ddagent"
 	DDTestingWorkloadNamespace = "ddtestworkload"
+	DDDogstatsdNamespace       = "dddogstatsd"
 
 	// Infra namespace
 	DDInfraEnvironment                      = "env"
@@ -45,6 +46,10 @@ const (
 
 	// Testing workload namerNamespace
 	DDTestingWorkloadDeployParamName = "deploy"
+
+	// Dogstatsd namespace
+	DDDogstatsdDeployParamName        = "deploy"
+	DDDogstatsdFullImagePathParamName = "fullImagePath"
 )
 
 type CommonEnvironment struct {
@@ -56,6 +61,7 @@ type CommonEnvironment struct {
 	InfraConfig           *sdkconfig.Config
 	AgentConfig           *sdkconfig.Config
 	TestingWorkloadConfig *sdkconfig.Config
+	DogstatsdConfig       *sdkconfig.Config
 
 	username string
 }
@@ -66,6 +72,7 @@ func NewCommonEnvironment(ctx *pulumi.Context) (CommonEnvironment, error) {
 		InfraConfig:           sdkconfig.New(ctx, DDInfraConfigNamespace),
 		AgentConfig:           sdkconfig.New(ctx, DDAgentConfigNamespace),
 		TestingWorkloadConfig: sdkconfig.New(ctx, DDTestingWorkloadNamespace),
+		DogstatsdConfig:       sdkconfig.New(ctx, DDDogstatsdNamespace),
 		CommonNamer:           namer.NewNamer(ctx, ""),
 		providerRegistry:      newProviderRegistry(ctx),
 	}
@@ -271,4 +278,14 @@ type Environment interface {
 // Testing workload namespace
 func (e *CommonEnvironment) TestingWorkloadDeploy() bool {
 	return e.GetBoolWithDefault(e.TestingWorkloadConfig, DDTestingWorkloadDeployParamName, true)
+}
+
+// Dogstatsd namespace
+
+func (e *CommonEnvironment) DogstatsdDeploy() bool {
+	return e.GetBoolWithDefault(e.DogstatsdConfig, DDDogstatsdDeployParamName, true)
+}
+
+func (e *CommonEnvironment) DogstatsdFullImagePath() string {
+	return e.GetStringWithDefault(e.DogstatsdConfig, DDDogstatsdFullImagePathParamName, "gcr.io/datadoghq/dogstatsd")
 }
--- a/components/datadog/apps/dogstatsd/k8s.go
+++ b/components/datadog/apps/dogstatsd/k8s.go
@@ -1,6 +1,8 @@
 package dogstatsd
 
 import (
+	"fmt"
+
 	"github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/utils"
 
@@ -15,11 +17,11 @@ type K8sComponent struct {
 	pulumi.ResourceState
 }
 
-func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
+func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, statsdPort int, statsdSocket string, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
 	opts = append(opts, pulumi.Provider(kubeProvider), pulumi.Parent(kubeProvider), pulumi.DeletedWith(kubeProvider))
 
 	k8sComponent := &K8sComponent{}
-	if err := e.Ctx.RegisterComponentResource("dd:apps", "dogstatsd", k8sComponent, opts...); err != nil {
+	if err := e.Ctx.RegisterComponentResource("dd:apps", fmt.Sprintf("dogstatsd-%d", statsdPort), k8sComponent, opts...); err != nil {
 		return nil, err
 	}
 
@@ -36,7 +38,7 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
 
 	opts = append(opts, utils.PulumiDependsOn(ns))
 
-	if _, err := appsv1.NewDeployment(e.Ctx, "dogstatsd-uds", &appsv1.DeploymentArgs{
+	if _, err := appsv1.NewDeployment(e.Ctx, fmt.Sprintf("dogstatsd-uds-%d", statsdPort), &appsv1.DeploymentArgs{
 		Metadata: &metav1.ObjectMetaArgs{
 			Name:      pulumi.String("dogstatsd-uds"),
 			Namespace: pulumi.String(namespace),
@@ -65,7 +67,7 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
 							Env: &corev1.EnvVarArray{
 								&corev1.EnvVarArgs{
 									Name:  pulumi.String("STATSD_URL"),
-									Value: pulumi.StringPtr("unix:///var/run/datadog/dsd.socket"),
+									Value: pulumi.String("unix:///var/dsd.socket"),
 								},
 							},
 							Resources: &corev1.ResourceRequirementsArgs{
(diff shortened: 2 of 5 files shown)


components/datadog/dogstatsd-standalone/k8s.go
```diff
@@ -0,0 +1,258 @@
...
+							},
+							Env: &corev1.EnvVarArray{
+								&corev1.EnvVarArgs{
+									Name:  pulumi.String("DD_API_KEY"),
+									Value: e.DogstatsdAPIKey(),
+								},
+								&corev1.EnvVarArgs{
+									Name:  pulumi.String("DD_LOG_LEVEL"),
+									Value: pulumi.StringPtr("DEBUG"),
+								},
```
L3n41c, line 92: For [the agent, we don’t set DEBUG log level](https://github.com/DataDog/test-infra-definitions/blob/8158a5b2c60e9d31f827390168a916a2534c2ac1/components/datadog/agent/helm.go).
So, maybe we can keep the default log level for standalone dogstatsd as well.

components/datadog/dogstatsd-standalone/k8s.go
```diff
@@ -0,0 +1,258 @@
...
+									Value: pulumi.StringPtr(Socket),
+								},
+								&corev1.EnvVarArgs{
+									Name:  pulumi.String("DD_DOGSTATSD_TAG_CARDINALITY"),
+									Value: pulumi.StringPtr("orchestrator"),
+								},
+								&corev1.EnvVarArgs{
+									Name:  pulumi.String("DD_KUBELET_TLS_VERIFY"),
+									Value: pulumi.StringPtr("false"),
+								},
```
L3n41c, line 120: Is this necessary ?
[I couldn’t see `kubelet.tlsVerify: false` for the agent](https://github.com/DataDog/test-infra-definitions/blob/8158a5b2c60e9d31f827390168a916a2534c2ac1/components/datadog/agent/helm.go).

components/datadog/dogstatsd-standalone/k8s.go
```diff
@@ -0,0 +1,258 @@
...
+		RoleRef: v1.RoleRefArgs{
+			ApiGroup: pulumi.String("rbac.authorization.k8s.io"),
+			Kind:     pulumi.String("ClusterRole"),
+			Name:     pulumi.String("dogstatsd-standalone"),
+		},
+		Subjects: v1.SubjectArray{
+			&v1.SubjectArgs{
+				Kind:      pulumi.String("ServiceAccount"),
+				Name:      pulumi.String("dogstatsd-standalone"),
+				Namespace: pulumi.String("dogstatsd-standalone"),
```
L3n41c, line 236:
```suggestion
				Namespace: pulumi.String(namespace),
```

components/datadog/dogstatsd-standalone/k8s.go
L3n41c, line 1: The agent is configured to dual ship data to:
* [the fake intake](https://github.com/DataDog/test-infra-definitions/blob/8158a5b2c60e9d31f827390168a916a2534c2ac1/components/datadog/agent/helm.go#L322-L331) because it’s the source of data that the `datadog-agent` tests’ are asserting. And
* [`dddev`](https://github.com/DataDog/test-infra-definitions/blob/8158a5b2c60e9d31f827390168a916a2534c2ac1/components/datadog/agent/helm.go#L79-L88) because the fake intake lacks visualisation tooling. So, in case of failure, it’s way more convenient to be able to browse the data on the real datadog UI (with [this dashboard for ex.](https://dddev.datadoghq.com/dashboard/qcp-brm-ysc/e2e-tests-containers-k8s))

Can `dogstatsd` standalone also be configured to dual-ship data ?

common/config/environment.go
```diff
@@ -45,6 +46,11 @@ const (
 
 	// Testing workload namerNamespace
 	DDTestingWorkloadDeployParamName = "deploy"
+
+	// Dogstatsd namespace
+	DDDogstatsdDeployParamName        = "deploy"
+	DDDogstatsdAPIKeyParamName        = "apiKey"
```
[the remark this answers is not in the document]
davidor: Do you think that we should just reuse `ddagent:apiKey` or go for the solution that you suggested in the comment below?

components/datadog/dogstatsd-standalone/k8s.go
davidor: Fixed :+1:

components/datadog/dogstatsd-standalone/k8s.go
davidor: This is what we have in our k8s-testing-resources repo, but I also think that we should leave the default.
Changed :+1:

components/datadog/dogstatsd-standalone/k8s.go
davidor: It doesn't work for me with tls verify set to true, but I'm not sure why it works in the agent and not dogstatsd standalone.

components/datadog/apps/dogstatsd/k8s.go
```diff
@@ -139,7 +141,7 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
 								},
 								&corev1.EnvVarArgs{
 									Name:  pulumi.String("STATSD_URL"),
-									Value: pulumi.StringPtr("$(HOST_IP):8125"),
+									Value: pulumi.StringPtr(fmt.Sprintf("$(HOST_IP):%d", statsdPort)),
```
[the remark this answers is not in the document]
davidor: Fixed :+1:

components/datadog/apps/dogstatsd/k8s.go
```diff
@@ -65,7 +67,7 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
 							Env: &corev1.EnvVarArray{
 								&corev1.EnvVarArgs{
 									Name:  pulumi.String("STATSD_URL"),
-									Value: pulumi.StringPtr("unix:///var/run/datadog/dsd.socket"),
+									Value: pulumi.StringPtr(statsdSocket),
```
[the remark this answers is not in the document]
davidor: This is better. Changed :+1:

components/datadog/dogstatsd-standalone/k8s.go
davidor: Actually, when deploying on Kind we also set tlsVerify: false for the agent:
https://github.com/DataDog/test-infra-definitions/blob/8158a5b2c60e9d31f827390168a916a2534c2ac1/scenarios/aws/kindvm/run.go#L58

L3n41c: Ah ! Thanks for catching it !
So, for `dogstatsd-standalone`, would it be possible to also have `kubelet:tlsVerify: false` only for Kind and not for all kubernetes setups ?

common/config/environment.go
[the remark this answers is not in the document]
L3n41c: I think that just reusing `ddagent:apiKey` is simpler.
And I cannot see any limitation it would bring. (Why would we want the agent and dogstatsd standalone to target two different orgs ?)
Let’s go for the simplest solution.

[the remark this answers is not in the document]
davidor: Changed :+1:

common/config/environment.go
```diff
@@ -272,3 +280,17 @@ type Environment interface {
...
+
+// Dogstatsd namespace
+
+func (e *CommonEnvironment) DogstatsdDeploy() bool {
+	return e.GetBoolWithDefault(e.DogstatsdConfig, DDDogstatsdDeployParamName, false)
+}
+
+func (e *CommonEnvironment) DogstatsdAPIKey() pulumi.StringOutput {
+	return e.DogstatsdConfig.RequireSecret(DDDogstatsdAPIKeyParamName)
+}
```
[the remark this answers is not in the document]
davidor: In the end we decided to use `ddagent:apiKey` only.

components/datadog/dogstatsd-standalone/k8s.go
davidor: Yes, it can. I used the same ENVs as in the agent and it works.
I pushed this change in a new commit.

davidor:
@L3n41c I addressed all your comments.

common/config/environment.go
```diff
@@ -272,3 +279,13 @@ type Environment interface {
 func (e *CommonEnvironment) TestingWorkloadDeploy() bool {
 	return e.GetBoolWithDefault(e.TestingWorkloadConfig, DDTestingWorkloadDeployParamName, true)
 }
+
+// Dogstatsd namespace
+
+func (e *CommonEnvironment) DogstatsdDeploy() bool {
+	return e.GetBoolWithDefault(e.DogstatsdConfig, DDDogstatsdDeployParamName, false)
```
L3n41c, line 286: For all the other types of workload, the `deploy` option defaults to `true`.

Cons of having all `deploy` options defaulting to `true`:
* By default and without any explicit parameter from the user, we get the fullest setup.
  * It could take the more time to startup but as everything can be started in parallel, I’m not sure it’s that true.
  * It could be more expensive but as long as everything fits on the same number of k8s nodes, it doesn’t change anything.

Pros of having all `deploy` options defaulting to `true`:
* By default and without any explicit parameter from the user, they will test everything. So, it highly reduces the risk of having people committing things that haven’t been tested.

```suggestion
	return e.GetBoolWithDefault(e.DogstatsdConfig, DDDogstatsdDeployParamName, true)
```

components/datadog/apps/dogstatsd/k8s.go
```diff
@@ -65,7 +67,7 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
 							Env: &corev1.EnvVarArray{
 								&corev1.EnvVarArgs{
 									Name:  pulumi.String("STATSD_URL"),
-									Value: pulumi.StringPtr("unix:///var/run/datadog/dsd.socket"),
+									Value: pulumi.String("unix:///var/dsd.socket"),
```
L3n41c, line 70: Pulumi is magic 🪄
`pulumi.String(…)` and `pulumi.StringPtr(…)` are interchangeable !

components/datadog/dogstatsd-standalone/k8s.go
```diff
@@ -0,0 +1,273 @@
...
+		envVars = append(
+			envVars,
+			&corev1.EnvVarArgs{
+				Name:  pulumi.String("DD_ADDITIONAL_ENDPOINTS"),
+				Value: pulumi.Sprintf(`{"http://%s": ["FAKEAPIKEY"]}`, fakeIntake.Host),
+			},
+			&corev1.EnvVarArgs{
+				Name:  pulumi.String("DD_LOGS_CONFIG_ADDITIONAL_ENDPOINTS"),
+				Value: pulumi.Sprintf(`[{"host": "%s", "port": 80, "is_reliable": true, "usessl": false}]`, fakeIntake.Host),
+			},
```
L3n41c, line 100: This one shouldn’t be needed as `dogstatsd-standalone` won’t run a logs agent.

L3n41c (approved):

components/datadog/dogstatsd-standalone/k8s.go
davidor: Right, this was a copy-pasta oversight.

common/config/environment.go
davidor: Right, at first I set it to false because of the conflicts with the agent, but now it should be OK.
I think we can set it to to true. I don't think the dogstatsd-standalone deployment takes a lot of time compared to everything else that needs to be deployed.

L3n41c (approved):

PR #418, opened 2023-10-31
Add AWS account `TSE-Playground`

jeanbaptiste-pham:
What does this PR do?
---------------------

Create the basic setup to create resources in a the new AWS account, `tse-playground`

-------------------

No existing scenario will be impacted.
The new account will be able, in the end, to use all existing scenarii.

Motivation
----------

Solution Engineers often need to replicate customer environment to help solving customer issues.
This requires both skill and time. The SET team (team doing Tooling for Solution Engineers) would like to provide an environment where these env can be spin up easily, as you did. We would like to try to leverage all the work that you already did on the subject.

Additional Notes
----------------

- I tried the scenario `scenario=aws/eks`, but it's not working yet. It's failing when creating the VpcCni:
image

Would you have an idea of what I missed?

- I'm not very clear about the AppGate access requirement, should I request to have the AWS account configured to access it? (currently, I see no need for us to connect to the CI)


pducolin:
Hi @jeanbaptiste-pham !

Indeed an account requires networks that are accessible from your local machine. We use appgate to authorise the access with special rules that are available only to people who strictly require working with that account.

jeanbaptiste-pham:
Hey @pducolin!
Thanks a lot for the answer :-)
I asked Fabric to setup the network of this new aws account to be able to access AppGate as you do, in the hope to do something similar. They agreed, and we should hopefully get this by the end of this quarter 🤞 
I thought I'd then resume on this PR...

tasks/setup.py
```diff
@@ -12,7 +12,7 @@
 from .config import Config, get_full_profile_path, get_local_config
 from .tool import ask, info, is_linux, is_windows, warn
 
-available_aws_accounts = ["agent-sandbox", "sandbox", "agent-qa"]
+available_aws_accounts = ["agent-sandbox", "sandbox", "agent-qa", "playground"]
```
prabhjes82, line 15: tse-playground

vboulineau (approved):
I cannot check the validity of the default values as we don't own the tenant but everything is 👍

PR #428, opened 2023-11-03
[components/datadog/dogstatsd-standalone] Create image pull secret if needed

davidor:
What does this PR do?
---------------------

Creates the image pull secret when it's needed in the dogstats-standalone daemonset.

### DIFF
--- a/components/datadog/dogstatsd-standalone/k8s.go
+++ b/components/datadog/dogstatsd-standalone/k8s.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/utils"
+	"github.com/DataDog/test-infra-definitions/components/datadog/agent"
 	ddfakeintake "github.com/DataDog/test-infra-definitions/components/datadog/fakeintake"
 	"github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes"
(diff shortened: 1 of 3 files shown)


components/datadog/dogstatsd-standalone/k8s.go
```diff
@@ -52,6 +53,19 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
...
 	opts = append(opts, utils.PulumiDependsOn(ns))
 
+	var imagePullSecrets corev1.LocalObjectReferenceArray
+	if e.ImagePullRegistry() != "" {
+		imgPullSecret, err := agent.NewImagePullSecret(e, namespace, opts...)
+		if err != nil {
+			return nil, err
+		}
+
+		opts = append(opts, utils.PulumiDependsOn(imgPullSecret))
```
L3n41c, line 63: This isn’t needed because the `appsv1.DaemonSet` object below already has a direct reference to this `corev1.Secret`.
In pulumi, values are kind of [“futures” or “promises”](https://en.wikipedia.org/wiki/Futures_and_promises).
If a resource uses, as input, the output of another resource, then, there is already an implicit dependency between the two since the full creation of the later resource is needed to get the actual value of the parameter needed to create the former one.

```suggestion
```

davidor: Fixed :+1:

L3n41c (approved):

PR #454, opened 2023-11-16
Implement notification for the Linux platform

YoannGh:
What does this PR do?
---------------------

Use the `notify-send` utility to implement the notify function for Linux.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/tool.py
+++ b/tasks/tool.py
@@ -117,10 +117,11 @@ def get_image_description(ctx: Context, ami_id: str) -> Any:
 
 def notify(ctx, text):
     if is_linux():
-        notify_linux()
-    if is_windows():
+        notify_linux(ctx, text)
+    elif is_windows():
         notify_windows()
-    notify_macos(ctx, text)
+    else:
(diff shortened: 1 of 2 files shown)


pducolin (approved):
Nice 🎉

tasks/tool.py
```diff
@@ -132,9 +133,8 @@ def notify_macos(ctx, text):
     ctx.run(f"osascript -e '{CMD}' test-infra-definitions '{text}'")
 
 
-def notify_linux():
-    # TODO: Implement notification on linux. Would require linux computer (with desktop) to test
-    return
+def notify_linux(ctx, text):
+    ctx.run(f"notify-send 'test-infra-definitions' '{text}'")
```
pducolin, line 137: **❓ question**
Is `notify-send` a system binary ? If not, can you add it as a dependency to the readme please ?

YoannGh: Done in my last commit, although I think the `notify-send` binary comes preinstalled with most Linux distributions.

PR #460, opened 2023-11-20
Define fakeintake healthcheck, enforce fakeintake health during deployment

pgimalac:
What does this PR do?
---------------------
- enforce fakeintake health during deployment, have pulumi fail if fakeintake is not healthy
- define health check command of the fakeintake task in AWS

-------------------
Fakeintake

Motivation
----------
Test could start before the fakeintake was healthy, we want to be able to
1. fail the deployment if the fakeintake was not healthy after some time
2. as a bonus, see from the AWS interface whether a fakeintake is healthy

Additional Notes
----------------
Noticed that the definition of the fakeintake service was duplicated so factored that.

The health check in AWS uses `curl`, so this PR depends on https://github.com/DataDog/datadog-agent/pull/20940 which adds `curl` to the container.


pducolin (approved):
Nice !!

PR #450, opened 2023-11-16
Disable logs dual shipping towards fake intake

L3n41c:
What does this PR do?
---------------------

Disable the logs dual shipping towards fake intake.

Which scenarios this will impact?
-------------------

* `ECS`
* `K8s`

Motivation
----------

ECS tests are currently very flaky because the agent is endlessly restarting.
It is in fact killed by ECS because its health check fails.

The component whose health check is failing is always the same:
```
2023-11-10 15:42:10 UTC | CORE | INFO | (cmd/agent/api/internal/agent/agent.go:350 in getHealth) | Healthcheck failed on: [ad-servicelistening]
```

Here is the stack trace of the stuck goroutine:
```
goroutine 474 [chan send]:
github.com/DataDog/datadog-agent/pkg/logs/sources.(*LogSources).AddSource(0xc000ea3d40, 0xc000acfe30)
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/sources/sources.go:59 +0x350
github.com/DataDog/datadog-agent/pkg/logs/schedulers.(*sourceManager).AddSource(0xc0024a3fe0, 0xc000acfe30)
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/schedulers/source_manager.go:26 +0x2c
github.com/DataDog/datadog-agent/pkg/logs/schedulers/ad.(*Scheduler).Schedule(0xc0007f45b8, {0xc0036c9800, 0x2, 0x2})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/schedulers/ad/scheduler.go:78 +0x624
github.com/DataDog/datadog-agent/pkg/logs/internal/util/adlistener.(*ADListener).Schedule(0xc0025fc9c0, {0xc0036c9800, 0x2, 0x2})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/internal/util/adlistener/ad.go:61 +0x36
github.com/DataDog/datadog-agent/pkg/autodiscovery/scheduler.(*MetaScheduler).Schedule(0xc004fee150, {0xc0036c9800, 0x2, 0x2})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/scheduler/meta.go:81 +0x3a9
[...]

PR #463, opened 2023-11-22
[CONTINT-3520] Add an HTTPS endpoint to the fake intake

L3n41c:
What does this PR do?
---------------------

Add an HTTPS endpoint to the fake intake.

-------------------

* `ECS`
* `K8s`

Motivation
----------

ECS tests are currently very flaky because the agent is endlessly restarting.
It is in fact killed by ECS because its health check fails.

The component whose health check is failing is always the same:
```
2023-11-10 15:42:10 UTC | CORE | INFO | (cmd/agent/api/internal/agent/agent.go:350 in getHealth) | Healthcheck failed on: [ad-servicelistening]
```

Here is the stack trace of the stuck goroutine:
```
goroutine 474 [chan send]:
github.com/DataDog/datadog-agent/pkg/logs/sources.(*LogSources).AddSource(0xc000ea3d40, 0xc000acfe30)
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/sources/sources.go:59 +0x350
github.com/DataDog/datadog-agent/pkg/logs/schedulers.(*sourceManager).AddSource(0xc0024a3fe0, 0xc000acfe30)
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/schedulers/source_manager.go:26 +0x2c
github.com/DataDog/datadog-agent/pkg/logs/schedulers/ad.(*Scheduler).Schedule(0xc0007f45b8, {0xc0036c9800, 0x2, 0x2})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/schedulers/ad/scheduler.go:78 +0x624
github.com/DataDog/datadog-agent/pkg/logs/internal/util/adlistener.(*ADListener).Schedule(0xc0025fc9c0, {0xc0036c9800, 0x2, 0x2})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/logs/internal/util/adlistener/ad.go:61 +0x36
github.com/DataDog/datadog-agent/pkg/autodiscovery/scheduler.(*MetaScheduler).Schedule(0xc004fee150, {0xc0036c9800, 0x2, 0x2})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/scheduler/meta.go:81 +0x3a9
github.com/DataDog/datadog-agent/pkg/autodiscovery.(*AutoConfig).applyChanges(0xc000387f40, {{0xc0036c9800, 0x2, 0x2}, {0x0, 0x0, 0x0}})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/autoconfig.go:472 +0x2dd
github.com/DataDog/datadog-agent/pkg/autodiscovery.(*AutoConfig).processNewService(0xc000387f40, {0xb3d4060, 0xc00494db20}, {0xb3efac0, 0xc001008750})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/autoconfig.go:431 +0x485
github.com/DataDog/datadog-agent/pkg/autodiscovery.(*AutoConfig).checkTagFreshness(0xc000387f40, {0xb3d4060, 0xc00494db20})
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/autoconfig.go:140 +0x472
github.com/DataDog/datadog-agent/pkg/autodiscovery.(*AutoConfig).serviceListening(0xc000387f40)
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/autoconfig.go:118 +0x308
created by github.com/DataDog/datadog-agent/pkg/autodiscovery.NewAutoConfig in goroutine 447
        /home/lenaic/doc/devel/DataDog/datadog-agent/pkg/autodiscovery/autoconfig.go:71 +0x8d
```

So, the `ad-servicelistening` goroutine is blocked on a full `chan` because the logs component is stuck.

On the other side, without logs dual shipping towards fake intake, `agent status` shows:
```
==========
Logs Agent
==========
    Reliable: Sending compressed logs in HTTPS to agent-http-intake.logs.datadoghq.com on port 443
```
which is what is expected.

But with dual shipping setup (current `main` branch), `agent status` shows:
```
Logs Agent
==========
    Reliable: Sending uncompressed logs in SSL encrypted TCP to agent-intake.logs.datadoghq.com on port 10516
    Reliable: Sending uncompressed logs in SSL encrypted TCP to internal-lenaic-eks-fakeintake-1018237308.us-east-1.elb.amazonaws.com on port 80

    You are currently sending Logs to Datadog through TCP (either because logs_config.force_use_tcp or logs_config.socks5_proxy_address is set or the HTTP connectivity test has failed). To benefit from increased reliability and better network performances, we strongly encourage switching over to compressed HTTPS which is now the default protocol.
```

and the logs are full of:
```
2023-11-16 09:19:13 UTC | CORE | WARN | (pkg/logs/client/tcp/connection_manager.go:116 in NewConnection) | tls: first record does not look like a TLS handshake
```

Not only SSL is enabled on the extra endpoint whereas it shouldn’t but the main endpoint (which wasn’t expected to be impacted by the extra one) is now using the deprecated TCP protocol instead of HTTPS.

All this is because:
* It is not possible to have SSL enabled on the main endpoint (with HTTPS) and disabled on extra-endpoints (plain HTTP): 

https://github.com/DataDog/datadog-agent/blob/aaecb265b539bb2a998074ee0beda6d968dce453/comp/logs/agent/config/config.go#L176-L178

https://github.com/DataDog/datadog-agent/blob/aaecb265b539bb2a998074ee0beda6d968dce453/comp/logs/agent/config/config.go#L249-L251

* Because SSL is enabled on the main endpoint, it probes the extra endpoint in HTTPS.
* It fails because the extra endpoint is plain HTTP.
* It fallback to TCP (with SSL) instead of HTTPS not only on the extra endpoint, but also on the main one.

So, we need a way to target the fake intake with HTTPS.

With this PR, `agent status` shows:
```
==========
Logs Agent
==========
    Reliable: Sending compressed logs in HTTPS to agent-http-intake.logs.datadoghq.com on port 443
    Reliable: Sending compressed logs in HTTPS to internal-lenaic-eks-fakeintake-782409152.us-east-1.elb.amazonaws.com on port 443
```
which is exactly what we want.

Additional Notes
----------------

This PR supersedes #450.


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/DataDog/test-infra-definitions
 go 1.20
 
 require (
+	github.com/DataDog/datadog-agent/pkg/util/pointer v0.49.0
```
vboulineau, line 6: Should not be a deps.

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -79,6 +86,90 @@ func NewECSFargateInstance(e aws.Environment, option ...fakeintakeparams.Option)
...
+			},
+			RevocationConfiguration: acmpca.CertificateAuthorityRevocationConfigurationArgs{
+				CrlConfiguration: acmpca.CertificateAuthorityRevocationConfigurationCrlConfigurationArgs{
+					Enabled: pulumi.BoolPtr(false),
+				},
+				OcspConfiguration: acmpca.CertificateAuthorityRevocationConfigurationOcspConfigurationArgs{
+					Enabled: pulumi.Bool(false),
+				},
+			},
+			// UsageMode:                   pulumi.StringPtr("SHORT_LIVED_CERTIFICATE"),
```
vboulineau, line 106: I think we can remove the comment.

go.mod
L3n41c: Fixed: https://github.com/DataDog/test-infra-definitions/pull/463/commits/e10a97f9b0aa648e0796e5957e3d73b61c584c9a

components/datadog/apps/fakeintake/fargateFakeintakeService.go
L3n41c: Fixed: https://github.com/DataDog/test-infra-definitions/pull/463/commits/e10a97f9b0aa648e0796e5957e3d73b61c584c9a

vboulineau (approved):

pducolin:
😮 Wow !

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -159,7 +248,7 @@ func fargateLinuxContainerDefinition(imageURL string) *ecs.TaskDefinitionContain
 		Environment: ecs.TaskDefinitionKeyValuePairArray{
 			ecs.TaskDefinitionKeyValuePairArgs{
 				Name:  pulumi.StringPtr("GOMEMLIMIT"),
-				Value: pulumi.StringPtr("768MiB"),
+				Value: pulumi.StringPtr("1536MiB"),
```
pducolin, line 251: **💭 thought**
I was thinking that we might reduce OOM by filtering the stored payloads, either in space, including only interesting metric names, or in time, limiting retention to 5 minutes, which should be more than enough for the tests

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -79,6 +85,89 @@ func NewECSFargateInstance(e aws.Environment, option ...fakeintakeparams.Option)
...
+			},
+		}, opts...)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, err = clb.NewListener(e.Ctx, namer.ResourceName("lb-https"), &clb.ListenerArgs{
+			LoadBalancerArn: alb.LoadBalancer.Arn(),
+			Port:            pulumi.IntPtr(sslPort),
+			Protocol:        pulumi.StringPtr("HTTPS"),
```
pducolin, line 159: **👏 praise**
Fascinating, can we have a listener without a load balancer ?

pducolin: Looks like we cannot https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html

components/datadog/apps/fakeintake/fargateFakeintakeService.go
```diff
@@ -79,6 +85,89 @@ func NewECSFargateInstance(e aws.Environment, option ...fakeintakeparams.Option)
...
+			return nil, err
+		}
+
+		domainName := alb.LoadBalancer.DnsName().ApplyT(func(dnsName string) *string {
+			if len(dnsName) <= 64 {
+				return &dnsName
+			}
+			dummyDNSName := "fakeintake.datadoghq.com"
+			return &dummyDNSName
+		}).(pulumi.StringPtrOutput)
```
pducolin, line 143: **❓ question**
How does this work ? How is domain name used ? How dummy name works ?

components/datadog/apps/fakeintake/fargateFakeintakeService.go
L3n41c: I think that we already have the payload retention limited in the fake intake.
So, if the flow of data remains more or less constant, the memory consumption of the fake intake should already stabilize after a period of time equal to the retention period.

The reason why I had to bump the memory is because, before this PR, the fake intake contained only metrics payloads. But with this fix, it now also contains logs payloads.
And the extra data made it be OOM killed again.

So, I had to increase the memory to make it able to store logs.

components/datadog/apps/fakeintake/fargateFakeintakeService.go
L3n41c: A standalone listener makes no sense if not attached to a load-balancer.
But a load-balancer can have several listeners.
That’s why, on AWS side, they are distinct objects.

The `awsx` pulumi module provides some syntax sugar to create in one call, the loadbalancer, and its attached target group and listeners. But with the lower level `aws` module, that would have been three calls to create three AWS objects.
This is even visible in the output of pulumi:
```
   │  ├─ awsx:lb:ApplicationLoadBalancer                                aws-fakeintake-lb                                 created (7s)        
   │  │  ├─ aws:lb:TargetGroup                                          aws-fakeintake-lb                                 created (4s)        
   │  │  ├─ aws:lb:LoadBalancer                                         aws-fakeintake-lb                                 created (17s)      
   │  │  └─ aws:lb:Listener                                             aws-fakeintake-lb-0                               created (5s)        
```

To create an HTTPS listener we have no choice but splitting things because:

* In order to create the HTTPS LB listener, we need an SSL certificate.
* In order to create the SSL certificate, we need to know the DNS name to put in it.
* And we need to create the load balancer to gets its DNS name.

components/datadog/apps/fakeintake/fargateFakeintakeService.go
L3n41c: The `DomainName` field ends in the Subject CN field of the X509 certificate.
This was the legacy way of specifying for which domain a certificate is issued. That’s why it’s still mandatory.
But this field was suffering from some limitations like being unique and limited to 64 bytes.
That’s why it has now been superseded by the SAN field which allows to specify longer domains.

So, if the real DNS name fits inside the 64 bytes limit, let’s use it in the subject CN because it’s nicer.
But otherwise, what’s really important nowadays is to set it properly in the SAN field.

image

PR #464, opened 2023-11-22
[CONTINT-3356] Allow running different k8s version kind clusters

adel121:
What does this PR do?
---------------------

This PR uses the kubernetes version that is set as a config variable to be used when spinning a kubernetes cluster and defining E2E resouces. 

This is achieved by adding adding a version matrix that defines which version of `kind` to install and what node image to use when installing `kind` on the vm and creating the cluster. 

-------------------
- aws/kind

Motivation
----------
Currently, the user can specify the desired kubernetes version by passing the `ddinfra:kubernetesVersion`. 

However:
- when installing `kind` on the VM, we don't take into consideration which `kind` version should be installed depending on the desired kubernetes version. 
- we always create a cluster with k8s version `1.26` even if the user sets a different version.
- when creating the resources, we use the same `APIVersion` regardless of the kubernetes version. Some `APIVersion` are not supported in older versions of kubernetes. 

The main motivation to change this behaviour is to be able to run E2E on different kubernetes versions (specifically on `aws/kind`)

Additional Notes
----------------
- Kubernetes versions older than `1.19` have issues when used with kind. We added support for versions `1.19+` 
- Had to change `autoscalingv2` to `autoscalingv2beta2` in `nginx` and `redis` for `HPA` when kubernetes version is `1.22` or older.
- Had to change `policyv1` to `policyv1beta1` in `nginx` and `redis` for `PodDisruptionBudget` when kubernetes version is `1.20` or older
- Had to explicitly set the `Protocol: pulumi.String("TCP")` because older versions don't provide a default value
- Running versions of kubernetes earlier than `1.19` would require running on with `cgroupv1`, while currently the EC2 instances and AMI we are using support `cgroupv2`. This will be solved in a separate PR (for example by changing the AMI of the EC2 machine to support earlier versions of kubernetes)


components/kubernetes/versions.go
```diff
@@ -0,0 +1,56 @@
+package kubernetes
```
vboulineau, line 1: This is specific to kind, not Kubernetes, so the file should be `kind_versions`

components/kubernetes/versions.go
```diff
@@ -0,0 +1,56 @@
...
+type kindConfig struct {
+	kindVersion string
+	nodeImage   string
+}
+
+// Source: https://github.com/kubernetes-sigs/kind/releases
+var kubeToKindVersion = map[string]kindConfig{
+	"1.27": {
+		kindVersion: "v0.20.0",
+		nodeImage:   "kindest/node:v1.27.3",
```
vboulineau, line 12: I am not sure you need to repeat `kindest/node` everywhere, especially if we want to change registry.

components/kubernetes/kind.go
```diff
@@ -35,10 +34,17 @@ func NewKindCluster(vm *vm.UnixVM, clusterName, arch string) (*remote.Command, p
 	if err != nil {
 		return nil, pulumi.StringOutput{}, err
 	}
+
+	kindVersionConfig, found := kubeToKindVersion[KubernetesVersion]
+
+	if !found {
```
vboulineau, line 40: I don't think you need to limit this. If version is not found, use latest Kind version and generate the image with the requested string version

adel121: What if the user asks for kubernetes version `1.18` ? 

If I default to the latest Kind version, then I won't be able to spin a cluster with kubernetes version `1.18` because the latest version of kind doesn't support `1.19` version of kubernetes.

So I will either leave it fail to create the cluster, and the user will have to understand what's going on. Or, I can simply tell the user that this is not going to work from the beginning and show them what kube versions they can use. I thought the latter would be more convenient.

WDYT?

PR #465, opened 2023-11-22
[ecs-ec2] Increase ASG to 2 hosts

wdhif:
What does this PR do?
---------------------
Increase the default EC2 Autoscalling Group to 2 to account for ECS scenario workload.

-------------------
This impact the ECS scenario. 

Motivation
----------
Without this, the `create-ecs` fails by default because of a Redis port collision.

Additional Notes
----------------


wdhif:
Is replaced by https://github.com/DataDog/test-infra-definitions/pull/466

PR #470, opened 2023-11-23
fix: remove health check command which breaks updateEnv

pgimalac:
What does this PR do?
---------------------
Remove the health check command from the fakeintake task definition.

-------------------
Fakeintake

Motivation
----------
When performing an `UpdateEnv`, pulumi redeploys the fakeintake (which shouldn't have changed), so the one the agent is using doesn't exist anymore.
It is not clear why but removing the health command removes this issue.

Additional Notes
----------------


KevinFairise2 (approved):
The healthcheck will come back stronger as soon as we find what is going on 🤣

pducolin:
What was the content passed to the call to `UpdateEnv` ?

PR #474, opened 2023-11-23
Add scaffolding of HyperV VMs using local Command

vboulineau:
What does this PR do?
---------------------

Add scaffolding for spawning HyperV VMs on Windows using local command (not working, requires implementation).

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/resources/hyperv/environment.go
+++ b/resources/hyperv/environment.go
@@ -1,0 +1,65 @@
+package hyperv
+
+import (
+	config "github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/DataDog/test-infra-definitions/common/namer"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+const (
+	hvConfigNamespace = "hyperv"
+	hvNamerNamespace  = "hv"
+
+	// HyperV Infra (local)
+	DDInfraDefaultPublicKeyPath      = "hv/defaultPublicKeyPath"
+	DDInfraDefaultPrivateKeyPath     = "hv/defaultPrivateKeyPath"
+	DDInfraDefaultPrivateKeyPassword = "hv/defaultPrivateKeyPassword"
+)
+
+type Environment struct {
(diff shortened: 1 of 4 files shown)


resources/hyperv/environment.go
```diff
@@ -0,0 +1,65 @@
...
+	hvConfigNamespace = "hyperv"
+	hvNamerNamespace  = "hv"
+
+	// HyperV Infra (local)
+	DDInfraDefaultPublicKeyPath      = "hv/defaultPublicKeyPath"
+	DDInfraDefaultPrivateKeyPath     = "hv/defaultPrivateKeyPath"
+	DDInfraDefaultPrivateKeyPassword = "hv/defaultPrivateKeyPassword"
+)
+
+type Environment struct {
```
pducolin, line 20: **💬 suggestion**
Could we rename to `HyperVEnvironment` ? I was confused between resource environment and command environment, which are the env vars.

scenarios/hyperv/vm.go
```diff
@@ -0,0 +1,28 @@
...
+		return err
+	}
+
+	vm, err := hyperv.NewVM(env, hyperv.VMArgs{})
+	if err != nil {
+		return err
+	}
+
+	// From here forward to whatever you want with your VM, it's the same as any other VM
+	_, err = agent.NewInstaller(vm)
```
pducolin, line 22: **👏 praise**
😍

pducolin (approved):
Minor suggestion on `hyperv.Environment` name

resources/hyperv/environment.go
vboulineau: No because the package is `hyperv` already, and as `govet` says, that stutters.
(Note that we use the same pattern everywhere)

pducolin: True, we do have the `aws.Environment`. Could we add an explicit check for the interface implementation ? It helps me with discoverability.

PR #482, opened 2023-11-29
Allow list of tags for each vmset

usamasaqib:
What does this PR do?
---------------------
This PR allows users to specify list of tags for VMSets. This allows a user to provide unique list of identifier to differentiate vmsets.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/microVMs/microvms/domain.go
+++ b/scenarios/aws/microVMs/microvms/domain.go
@@ -40,8 +40,8 @@ type Domain struct {
 }
 
 func generateDomainIdentifier(vcpu, memory int, vmsetTags, tag, arch string) string {
-	domainPrefix := fmt.Sprintf("%s-ddvm", vmsetTags)
-	return fmt.Sprintf("%s-%s-%s-%d-%d", domainPrefix, arch, tag, vcpu, memory)
+	setTagsWithDelim := fmt.Sprintf("%s-ddvm", vmsetTags)
+	return fmt.Sprintf("%s-%s-%s-%d-%d", arch, tag, setTagsWithDelim, vcpu, memory)
 }
 func generateNewUnicastMac(e config.CommonEnvironment, domainID string) (pulumi.StringOutput, error) {
 	r := utils.NewRandomGenerator(e, domainID)
--- a/scenarios/aws/microVMs/microvms/domain.go
+++ b/scenarios/aws/microVMs/microvms/domain.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
+	"strings"
 
 	"github.com/pulumi/pulumi-libvirt/sdk/go/libvirt"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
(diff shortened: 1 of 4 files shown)


scenarios/aws/microVMs/microvms/domain.go
```diff
@@ -38,8 +39,9 @@ type Domain struct {
 	lvDomain    *libvirt.Domain
 }
 
-func generateDomainIdentifier(vcpu, memory int, vmsetName, tag, arch string) string {
-	return fmt.Sprintf("ddvm-%s-%s-%s-%d-%d", vmsetName, arch, tag, vcpu, memory)
+func generateDomainIdentifier(vcpu, memory int, vmsetTags, tag, arch string) string {
+	setTagsWithDelim := fmt.Sprintf("%s-ddvm", vmsetTags)
+	return fmt.Sprintf("%s-%s-%s-%d-%d", arch, tag, setTagsWithDelim, vcpu, memory)
```
val06, line 44: what is the usage of `domainID`? is it used for filenames and potentially can exceed the `MAX_PATH` (255 chars)?

val06 (changes requested):
cred

scenarios/aws/microVMs/microvms/domain.go
usamasaqib: `domainID` is used for naming the VMs and generating their resource URN. The URN has the 255 character limit and yes it can potentially exceed this limit. 
The problem is that we overload the VM name in the datadog-agent CI for matching a VM name as reported by libvirt to its corresponding VMset. This is why we have to incorporate the tag information in the `domainID`. The rest of the information, such as the `arch` and `tag` are required for generating a unique URN

val06: if we end up generating a longer than 255 string in `generateDomainIdentifier` we end up with an error later in the callstack?

should we consider adding "truncation" logic in this function to generate only valid names? (maybe performing some transformation on `setTagsWithDelim` as I assume this can be the main cause for exceeding the limit)

WDYT?

usamasaqib: The problem is that at this stage we do not know if we will be above the 255 character limit. The reason for this is that the underlying `test-infra-definitions` framework also adds strings to the resource URNs.

Secondly, truncation cannot work even if it was possible. The `domainID` is generated using user provided `tags`. Atleast in the datadog-agent CI we expect to find these tags in the domain name.

scenarios/aws/microVMs/microvms/libvirt_fs.go
```diff
@@ -34,9 +34,9 @@ func fsPathToLibvirtResource(path string) string {
 
 // the vmset name deduplicates volume resource name for the same VMs launched in different vmsets
 // the architecture deduplicates volume resource name for the same VMs launched with different archs.
-func getNamer(ctx *pulumi.Context, vmsetName, arch string) func(string) namer.Namer {
+func getNamer(ctx *pulumi.Context, arch string, vmsetTags []string) func(string) namer.Namer {
```
brycekahle, line 37: Can you use the `VMSet.ID` field here instead?

val06 (approved):

scenarios/aws/microVMs/microvms/libvirt_fs.go
```diff
@@ -130,7 +130,7 @@ func NewLibvirtFSDistroRecipe(ctx *pulumi.Context, vmset *vmconfig.VMSet, pools
 		pools:         pools,
 		volumes:       volumes,
 		baseVolumeMap: baseVolumeMap,
-		fsNamer:       libvirtResourceNamer(ctx, vmset.Name),
+		fsNamer:       libvirtResourceNamer(ctx, vmset.Tags...),
```
brycekahle, line 133: should this be `vmset.ID`?

usamasaqib: It can be used here. But since, there is no need to use the arch in the namer, I would rather avoid it.

brycekahle: Why is `arch` not necessary here?

scenarios/aws/microVMs/microvms/domain.go
brycekahle, line 44: Why not fold the above `fmt.Sprintf` into this?

scenarios/aws/microVMs/microvms/libvirt_fs.go
usamasaqib: Kind of incidental, but the reason is that the resources this namer is used to generate all use the `command.Runner` provided by the underlying test-infra framework. This appends to the URN some string generated from the name of the ec2 VM, which contains the `arch`

brycekahle (approved):

PR #239, opened 2023-07-10
AET-185 Use AMI with docker installed on Docker VM scenario

KevinFairise2:
What does this PR do?
---------------------
Change DockerVM scenario to use an AMI with docker installed.
Docker-compose is installed automatically since it it not integrated in the ECS AMI

Docker installation on VM is now optional and can be done with `WithDockerInstall` option on DockerManager.

Which scenarios this will impact?
-------------------
It impacts DockerVM scenario

Motivation
----------

Using an AMI where Docker is already installed is faster to run 

Additional Notes
----------------

PR #484, opened 2023-11-29
Introduce Components and migrate Agent, VM and Kube Clusters to it

vboulineau:
What does this PR do?
---------------------

Introduce Components and migrate Agent, VM and Kube Clusters to it.
Implementation following PoC from #244.

-------------------

Mostly everything is impacted by the changes.

Motivation
----------

Additional Notes
----------------


pducolin:
Can you split changes in more atomic commits ? For example isolating renaming to one single commit, and tinier refactoring to separate commits. It takes a longer time to review this in one single commit.

vboulineau:
> Can you split changes in more atomic commits ? For example isolating renaming to one single commit, and tinier refactoring to separate commits. It takes a longer time to review this in one single commit.

I understand reviewing this PR is unfortunately going to be a lengthy exercice, but due to how the code was structured, it's difficult to make meaningful atomic changes as each small change will have repercussions in a lot of other places (basically all the VM stuff, which is 80% of the PR), so it would have created "fake" changes with repo in broken state.
There were probably some opportunities that I missed during development, but that ship has sailed unfortunately.

I could isolate the remaining 20% by putting some files in different (broken) commits, but I'm not sure it's going to ease the reviewing that much.

scenarios/aws/ec2/vm.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		// Create the EC2 instance
+		instance, err := ec2.NewInstance(e, name, instanceArgs, pulumi.Parent(c))
+		if err != nil {
+			return err
+		}
+
+		// Create connection
+		conn, err := remote.MakeConnection(instance.PrivateIp, amiInfo.defaultUser, e.DefaultPrivateKeyPath(), e.DefaultPrivateKeyPassword(), "")
```
pducolin, line 47: **❓ question**
Is `Make*` a go pattern ? I am more used to `New*` or `Create*`

scenarios/aws/fakeintake/params.go
```diff
@@ -28,15 +26,6 @@ func WithoutLoadBalancer() Option {
 	}
 }
 
-// WithName sets the name of the fakeintake.
```
pducolin, line 31: **❓ question**
Why was it removed ? We need it for the fallback test

common/config/environment.go
```diff
@@ -24,10 +24,9 @@ const (
 	// Infra namespace
 	DDInfraEnvironment                      = "env"
 	DDInfraKubernetesVersion                = "kubernetesVersion"
-	DDInfraOSFamily                         = "osFamily"
-	DDInfraOSArchitecture                   = "osArchitecture"
+	DDInfraOSDescriptor                     = "osDescriptor"
```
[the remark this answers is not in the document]
vboulineau: Yes it has `<osFamily>:<osVersion>:<osArch>` format with defaulting that allows to easily pass any combination.

common/utils/pulumi.go
```diff
@@ -7,3 +7,18 @@ import (
...
+	if len(opts) == 0 {
+		return current
+	}
+
+	addedOptions := make([]T, len(current)+len(opts))
+	for _, array := range [][]T{current, opts} {
+		for _, opt := range array {
+			addedOptions = append(addedOptions, opt)
+		}
+	}
```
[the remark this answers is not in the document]
pgimalac: Am I missing something or would the following work ?
```suggestion
	addedOptions := make([]T, 0, len(current)+len(opts))
	addedOptions = append(addedOptions, current...)
	addedOptions = append(addedOptions, opts...)
```

components/datadog/agentparams/version.go
```diff
@@ -0,0 +1,8 @@
+package agentparams
+
+type PackageVersion struct {
+	Major       string
```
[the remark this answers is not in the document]
vboulineau: I don't think we should restrict in the model, 6 is still officially supported

components/os/linux_descriptors.go
```diff
@@ -0,0 +1,30 @@
...
+
+	DebianDefault = Debian12
+	Debian12      = NewDescriptor(Debian, "12")
+
+	AmazonLinuxDefault = AmazonLinux2023
+	AmazonLinux2023    = NewDescriptor(AmazonLinux, "2023")
+	AmazonLinux2       = NewDescriptor(AmazonLinux, "2")
+
+	AmazonLinuxECSDefault = AmazonLinuxECS2023
+	AmazonLinuxECS2023    = NewDescriptor(AmazonLinuxECS, "2023")
```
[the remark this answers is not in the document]
vboulineau: Yes correct

components/os/macos_servicemanagers.go
```diff
@@ -0,0 +1,27 @@
...
+type macOSServiceManager struct {
+	e      config.CommonEnvironment
+	runner *command.Runner
+}
+
+func newMacOSServiceManager(e config.CommonEnvironment, runner *command.Runner) ServiceManager {
+	return &macOSServiceManager{e: e, runner: runner}
+}
+
+func (s *macOSServiceManager) EnsureRunning(serviceName string, triggers pulumi.ArrayInput, opts ...pulumi.ResourceOption) (*remote.Command, error) {
```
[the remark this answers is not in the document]
vboulineau: It may `Start` or `Restart`, I tried to find a name that express the desired state (as that's we want to do with Pulumi). I may need to think of a better naming but I don't think `Restart` expresses that.

components/remote/os.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/DataDog/test-infra-definitions/components/command"
+	"github.com/DataDog/test-infra-definitions/components/os"
+
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+func MakeHost(e config.CommonEnvironment, conn remote.ConnectionOutput, osDesc os.Descriptor, osUser string, readyFunc command.ReadyFunc, host *Host) error {
```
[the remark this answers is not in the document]
vboulineau: No, it's a function that fills a `Host` object from some generic input to avoid mistakes and repetitive code in all VM creation. We do not create anything in there. `InitHost` would work as well, but perhaps the intent is a bit less clear

resources/aws/ec2/vm.go
```diff
@@ -2,36 +2,61 @@ package ec2
...
 		},
 		InstanceInitiatedShutdownBehavior: pulumi.String(e.DefaultShutdownBehavior()),
-	}, e.WithProviders(config.ProviderAWS))
+	}, utils.MergeOptions(opts, e.WithProviders(config.ProviderAWS))...)
 	return instance, err
 }
+
+func defaultInstanceArgs(e aws.Environment, args *InstanceArgs) {
+	if args.InstanceType == "" {
+		args.InstanceType = e.DefaultInstanceType()
```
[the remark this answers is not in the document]
vboulineau: The environment provides the values read from either ConfigMap or default values, no changes on that side.

scenarios/aws/ec2/vm.go
vboulineau: It's not a Go pattern, I do see it used in some libraries but that's it. I could be `NewConnection` but I aligned with `MakeHost`.

scenarios/aws/fakeintake/params.go
vboulineau: Because `Name` is now standard parameter for all components given at creation.

common/utils/pulumi.go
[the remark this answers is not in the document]
vboulineau: That's indeed the case

components/datadog/agent/host.go
```diff
@@ -0,0 +1,194 @@
...
+		pulumiAgentString = pulumi.Sprintf("api_key: %v\n%v", env.AgentAPIKey(), pulumiAgentString)
+	}
+
+	copyCmd, err := h.targetOS.FileManager().CopyInlineFile(pulumiAgentString, configFullPath, true, opts...)
+	if err != nil {
+		return nil, pulumiAgentString, err
+	}
+
+	return copyCmd, pulumiAgentString, nil
+}
```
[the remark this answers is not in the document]
vboulineau: I prefer it as well, but note that it was previous code that just got moved. I still made the changes.

components/os/windows_descriptors.go
```diff
@@ -0,0 +1,9 @@
+//nolint:revive
```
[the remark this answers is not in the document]
vboulineau: It was because it did not like `WindowsServer_2022` names (basically `_`), but I changed to full camel case.

components/command/runner.go
```diff
@@ -140,6 +140,6 @@ func NewLocalRunner(e config.CommonEnvironment, args LocalRunnerArgs) *LocalRunn
 }
 
 func (r *LocalRunner) Command(name string, args *Args, opts ...pulumi.ResourceOption) (*local.Command, error) {
-	depends := append(opts, r.e.WithProviders(config.ProviderCommand))
-	return local.NewCommand(r.e.Ctx, r.namer.ResourceName("cmd", name), args.toLocalCommandArgs(r.config, r.osCommand), depends...)
+	opts = utils.MergeOptions[pulumi.ResourceOption](opts, r.e.WithProviders(config.ProviderCommand))
```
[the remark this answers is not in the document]
pducolin: Thank you !

components/command/utils.go
```diff
@@ -14,3 +16,12 @@ func WaitForCloudInit(runner *Runner) (*remote.Command, error) {
...
 }
+
+func WaitUntilSuccess(runner *Runner) (*remote.Command, error) {
+	return runner.Command(
+		"wait-until-success",
+		&Args{
+			// echo works in shell and powershell
+			Create: pulumi.String("echo \"OK\""),
+		})
+}
```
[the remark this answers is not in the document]
pducolin: Maybe `WaitForEcho` ?

components/command/windowsOSCommand.go
```diff
@@ -75,3 +79,18 @@ func (fs windowsOSCommand) BuildCommandString(
...
+	}
+	if strings.HasPrefix(path, "//") || strings.HasPrefix(path, `\\`) {
+		return true
+	} else if strings.Index(path, ":/") == 1 {
+		return true
+	} else if strings.Index(path, `:\`) == 1 {
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
pducolin: Can be a follow up yes

components/component.go
```diff
@@ -0,0 +1,138 @@
...
+	RegisterOutputs(ctx *pulumi.Context, self pulumi.ComponentResource) error
+}
+
+type Component struct {
+	name       string // Name is set to the name of Pulumi component, it allows to name dependencies easily.
+	outputs    pulumi.Map
+	exportName string
+}
+
+func (c *Component) init(name, exportName string) {
```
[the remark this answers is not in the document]
pducolin: **🥜 nitpick**
For readability I would move it after `NewComponent`

components/component.go
```diff
@@ -0,0 +1,136 @@
...
+	c.name = name
+	c.outputs = make(pulumi.Map)
+	c.exportName = exportName
+}
+
+func (c *Component) getOutputs() pulumi.Map { //nolint:unused
```
pducolin, line 59:
```suggestion
func (c *Component) getOutputs() pulumi.Map { //nolint:unused | Actually used through the interface
```

components/component.go
```diff
@@ -0,0 +1,136 @@
...
+	return c.outputs
+}
+
+func (c *Component) getExportName() string { //nolint:unused
```
pducolin, line 63:
```suggestion
func (c *Component) getExportName() string { //nolint:unused | Actually used through the interface
```

components/component.go
```diff
@@ -0,0 +1,138 @@
...
+	return c.exportName
+}
```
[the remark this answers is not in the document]
pducolin: Have you tried explicitly telling that `Component` implements `component` using

```golang
var _ component = &Component{}
```

?

components/datadog/agent/host.go
```diff
@@ -0,0 +1,194 @@
...
+
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type HostAgentOutput struct {
+	components.JSONImporter
+}
+
+// HostAgent is an installer for the Agent on a remote host
```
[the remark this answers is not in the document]
pducolin: When I think `remote` I think out of my machine, I can see how a local VM is still remote, ok

components/os/const.go
```diff
@@ -0,0 +1,101 @@
+package os
```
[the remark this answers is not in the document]
pducolin: The `NewArchitectureFromString` tricked me, could rename to `Get*`

components/os/const.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+type Architecture string
+
+const (
+	AMD64Arch = Architecture("x86_64")
+	ARM64Arch = Architecture("arm64")
+)
+
+func NewArchitectureFromString(archStr string) Architecture {
```
pducolin, line 15:
```suggestion
func GetArchitectureFromString(archStr string) Architecture {
```

components/os/const.go
```diff
@@ -0,0 +1,130 @@
...
+	RockyLinux     Flavor = iota
+
+	// Windows
+	WindowsServer Flavor = 500
+
+	// MacOS
+	MacosOS Flavor = 1000
+)
+
+func NewFlavorFromString(flavorStr string) Flavor {
```
pducolin, line 60:
```suggestion
func GetFlavorFromString(flavorStr string) Flavor {
```

components/os/descriptor.go
```diff
@@ -0,0 +1,50 @@
...
+	return NewDescriptorWithArch(f, version, AMD64Arch)
+}
+
+func NewDescriptorWithArch(f Flavor, version string, arch Architecture) Descriptor {
+	return Descriptor{
+		family:       f.Type(),
+		Flavor:       f,
+		Version:      version,
+		Architecture: arch,
+	}
```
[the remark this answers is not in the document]
pducolin: Never mind, I thought you were building a string out of params, but you build params out of string in `NewDescriptorFromString`

components/os/linux_servicemanagers.go
```diff
@@ -0,0 +1,43 @@
...
+// 	return &sysVServiceManager{e: e, runner: runner}
+// }
+
+// func (s *sysVServiceManager) EnsureRunning(serviceName string, triggers pulumi.ArrayInput, opts ...pulumi.ResourceOption) (*remote.Command, error) {
+// 	return s.runner.Command(s.e.CommonNamer.ResourceName("running", serviceName), &command.Args{
+// 		Sudo:     true,
+// 		Create:   pulumi.String("service restart " + serviceName),
+// 		Triggers: triggers,
+// 	}, opts...)
+// }
```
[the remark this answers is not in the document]
pducolin: If it doesn't break anything, let's remove it

components/remote/os.go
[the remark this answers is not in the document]
pducolin: Reading it I thought it was creating a host. It was weird that it was taking a host pointer and filled it. Maybe adding a description is enough.

pducolin, line 12:
```suggestion
// MakeHost helps fill a host with commonly required params
func MakeHost(e config.CommonEnvironment, conn remote.ConnectionOutput, osDesc os.Descriptor, osUser string, readyFunc command.ReadyFunc, host *Host) error {
```

scenarios/aws/ec2/vm.go
pducolin: The difference with `MakeHost` is that it fills an existing instance, while here you create a new one, which makes me think about `New*` for go

pducolin (approved):
Other than minor descriptions and comments, all good !

components/command/utils.go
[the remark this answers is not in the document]
vboulineau: I think `echo` is irrelevant here, we just wait for connection to success and we cannot do that without a command, so use a dummy command but it could any command really.
I will rename to `WaitForSuccessfulConnection`

components/component.go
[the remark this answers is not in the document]
vboulineau: You cannot do that because `component` requests the implementation of `pulumi.ComponentResource` interface as well, which `Component` does not.

components/component.go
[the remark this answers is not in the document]
vboulineau: `NewComponent` does not create a `Component` but an instance of _something that implements_ `component`, it has no relation to `Component` in itself. That's why the file order is like this.

components/os/const.go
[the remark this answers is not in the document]
vboulineau: I don't think the semantic of `Get` is well suited to the operation, so I've just removed the `New`

components/docker/docker.go
```diff
@@ -0,0 +1,170 @@
...
+		manager.opts = utils.MergeOptions(manager.opts, utils.PulumiDependsOn(installCmd))
+	}
+
+	composeCmd, err := manager.installCompose()
+	if err != nil {
+		return nil, nil, err
+	}
+	manager.opts = utils.MergeOptions(manager.opts, utils.PulumiDependsOn(composeCmd))
+
+	return manager, nil, nil
```
pducolin, line 54: **💬 suggestion**
I think the second one used to be the `installCmd`. If we always return nil then we can remove it from the return type

components/docker/docker.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+type Manager struct {
+	namer namer.Namer
+	host  *remoteComp.Host
+
+	opts []pulumi.ResourceOption
+}
+
+func NewManager(e config.CommonEnvironment, host *remoteComp.Host, installDocker bool, opts ...pulumi.ResourceOption) (*Manager, pulumi.Resource, error) {
```
pducolin, line 30: **❓ question**
So manager is no more lazy, right ?

components/kubernetes/kind.go
```diff
@@ -22,76 +24,81 @@ const (
...
-		commonEnvironment.CommonNamer.ResourceName("kind-install"),
-		&command.Args{
-			Create: pulumi.String(fmt.Sprintf(`curl -Lo ./kind "https://kind.sigs.k8s.io/dl/%s/kind-linux-%s" && sudo install kind /usr/local/bin/kind`, kindVersionConfig.kindVersion, arch)),
-		},
-		utils.PulumiDependsOn(cmd, curlCommand),
-	)
-	if err != nil {
-		return nil, pulumi.StringOutput{}, err
-	}
+		_, dockerInstallCmd, err := docker.NewManager(env, vm, true, opts...)
```
pducolin, line 40: **💬 suggestion**
You might want to return also the `installCompose` cmd, or a list of resources that `kindInstall` should depend on

components/os/os.go
```diff
@@ -1,40 +1,80 @@
...
+// os is a generic implementation of OS interface
+type os struct {
+	descriptor     Descriptor
+	runner         *command.Runner
+	fileManager    *command.FileManager
+	packageManager PackageManager
+	serviceManager ServiceManager
+}
+
+func (o os) Descriptor() Descriptor {
```
pducolin, line 43: **💬 suggestion**
```suggestion
var _ OS = &os{}

func (o os) Descriptor() Descriptor {
```

components/os/os.go
```diff
@@ -1,40 +1,80 @@
...
-	GetAgentConfigFolder() string
-	GetSSHUser() string
-	GetAgentInstallCmd(AgentVersion) (string, error)
-	GetRunAgentCmd(parameters string) string
-	GetType() Type
-	CreatePackageManager(runner *command.Runner) (command.PackageManager, error)
-	CheckIsAbsPath(string) bool
+	Descriptor() Descriptor
+
+	Runner() *command.Runner
```
pducolin, line 28: **💭 thought**
Should `Runner` be injected, rather than being a member of an instance ?

scenarios/aws/ec2/vm_run.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func VMRunWithDocker(ctx *pulumi.Context) error {
+	env, err := aws.NewEnvironment(ctx)
+	if err != nil {
+		return err
+	}
+
+	osDesc := os.DescriptorFromString(env.InfraOSDescriptor())
+	vm, err := NewVM(env, "vm", WithAMI(env.InfraOSImageID(), osDesc, osDesc.Architecture))
```
pducolin, line 61: **🔨 warning**
We should limit to use with Amazon ECS, and avoid installing Docker. This is a breaking change compared to what we have today.

pducolin (changes requested):

components/docker/docker.go
vboulineau: It should be `composeCmd` indeed, fixed.

components/docker/docker.go
vboulineau: That is correct, if you instantiate a `DockerManager` without calling it, it will still install Docker. Not sure it's a valid use case though.

components/kubernetes/kind.go
vboulineau: That's the beauty of components, it's not necessary. As they're all child of the `Cluster` component, external users only need to depend on `Cluster` and it will automatically wait on all children.

scenarios/aws/ec2/vm_run.go
vboulineau: I see that it was changed indeed https://github.com/DataDog/test-infra-definitions/pull/239.

I don't think we should limit test cases, especially for manual testing. We can handle in code to flip the boolean if the OS is `AmazonLinuxECS` but I see no valid reason to artificially limit Docker.

components/docker/docker.go
pducolin: Reading it from `datadog-agent` we should have `NewManagerWithoutDockerInstall` and `NewManagerWithDockerInstall`. Can be done in a follow up

scenarios/aws/ec2/vm_run.go
pducolin: I don't see usage of Docker out of `Amazon ECS`. For Windows I don't think the docker installer works today. So I don't see why we should not use the one with Docker out of the box. Maybe for testing on ne versions of Docker ?

components/kubernetes/kind.go
pducolin: What I meant is you were not returning the `docker-compose` cmd, but it is fixed now

pducolin (commented):
I think we should limit Docker Agent to the efficient path (Amazon ECS), or change it but in a separate PR

components/os/os.go
vboulineau: I am not sure I understand this one. `Runner` is already injected in the `os` implementation of the interface, the interface just mandates to have a way to return the `Runner`.

components/docker/docker.go
vboulineau: Isn't it the goal of the boolean? I think that we should move the `Docker` install to be conditional on presence of `docker` CLI, it would improve the experience without being tied to a specific OS.
It's for follow-up though.

scenarios/aws/ec2/vm_run.go
vboulineau: We have many use cases to test Docker with specific distribution, Kernel version, or cgroup version (v1 vs v2), to reproduce some customer cases or to manually test features (like cgroupv2 support recently).

components/docker/docker.go
pducolin: The boolean makes it harder to read the purpose, while implementing the feature. In other words, using it from `datadog-agent` it is more explicit to use 

```go
docker.NewManagerWithoutDockerInstall()
``` 

rather than

```go
docker.NewManager(false)
```

components/os/os.go
pducolin: If injected, I expect a getter on the injector rather than on the injected. But I can see we might not have access to the injector and needing it on the injected interface.

scenarios/aws/ec2/vm_run.go
pducolin: I see, we should make it more explicit that the recommended way is using it on Amazon ECS to have docker pre-installed. With this breaking change we are removing a limitation, we should replace it with explicit guidelines.

pducolin (approved):

pducolin (approved):

pducolin (approved):

scenarios/aws/microVMs/microvms/run.go
```diff
@@ -118,11 +111,11 @@ func newMetalInstance(instanceEnv *InstanceEnvironment, name, arch string, m con
...
 
-	awsInstance, err := ec2vm.NewEC2VMWithEnv(*awsEnv,
-		ec2params.WithImageName(ami, os.Architecture(arch), ec2os.UbuntuOS),
-		ec2params.WithInstanceType(instanceType),
-		ec2params.WithName(name),
-		ec2params.WithUserData(metalUserData),
+	awsInstance, err := ec2Scn.NewVM(*awsEnv, name,
+		ec2Scn.WithAMI(ami, os.UbuntuDefault, os.Architecture(arch)),
+		ec2Scn.WithInstanceType(instanceType),
+		ec2Scn.WithAMI(ami, os.UbuntuDefault, os.Architecture(arch)),
```
usamasaqib, line 117: Provided twice?

usamasaqib (approved):

PR #493, opened 2023-12-04
Rerun destroy with refresh on failure

KevinFairise2:
What does this PR do?
---------------------

Rerun the pulumi destroy command with refresh in case of failure, to prevent issue when the managed resources are already destroyed

-------------------

Motivation
----------

Additional Notes
----------------


vboulineau (approved):
LGTM but linter failure.

pducolin (approved):

PR #500, opened 2023-12-05
[CONTINT-3407] Add dogstatsd workloads in dockervm scenario

adel121:
What does this PR do?
---------------------

Adds dogstatsd workload to dockervm scenario:
- Dummy dogstatsd apps (one sending metrics on UDP and one on UDS)

-------------------
aws/dockervm

Motivation
----------
Be able to automate dogstatsd standalone and full agent on docker  QA tasks

Additional Notes
----------------
In order to deploy the scenario with dogstatsd standsalone, you only need to pass `-c ddagent:fullImagePath="datadog/dogstatsd:latest"` to the `pulumi up` command


components/datadog/apps/dogstatsd/docker-compose.yaml
```diff
@@ -0,0 +1,22 @@
...
+    container_name: metric-sender-uds
+    network_mode: host
+    pid: "host"
+    volumes:
+      - /var/run/datadog:/var/run/datadog:ro
+    command: /bin/bash -ec 'apt update; apt install -y netcat; while true; do echo -n "custom.metric.uds:1|c" | nc -U -u -w1 /var/run/datadog/dsd.socket; sleep 10; done'
+    restart: "no" 
+    depends_on:
+      - agent
+  dogstats-udp:
```
adel121, line 14: For some reason, the metrics send over UDP are not found on the DD backend.

The docker-compose definition of the app seems correct to me, I can't figure out if the issue is related to dogstatsd or to some missing configuration. 

`agent config` and `agent dogstatsd-stats` show that the metric is received over UDP, but still can't see it on the DD app.

components/datadog/agent/docker.go
```diff
@@ -98,12 +98,19 @@ func dockerAgentComposeContent(agentImagePath string, apiKey pulumi.StringInput,
...
+						"DD_API_KEY":                           apiKeyResolved,
+						"DD_PROCESS_AGENT_ENABLED":             true,
+						"DD_DOGSTATSD_NON_LOCAL_TRAFFIC":       true,
+						"DD_DOGSTATSD_ORIGIN_DETECTION":        true,
+						"DD_DOGSTATSD_TAG_CARDINALITY":         "high",
+						"DD_DOGSTATSD_SOCKET":                  "/var/run/datadog/dsd.socket",
+						"DD_DOGSTATSD_ORIGIN_DETECTION_CLIENT": true,
 					},
+					Pid:   "host",
+					Ports: []string{"8125:8125/udp"},
```
pducolin, line 113: **💭 thought**
Wonder if this should be an option, and if we should change this function's signature to a `DockerComposeManifestService` to merge the default one with. The env variables you add should be passed through `envVars`, using `WithAgentServiceEnvVars`, rather than having them as default

scenarios/aws/dockerVM/run.go
```diff
@@ -48,6 +49,12 @@ func Run(ctx *pulumi.Context) error {
 			}
 		}
 		agentOptions = append(agentOptions, dockeragentparams.WithAgentServiceEnvVariable("DD_LOG_LEVEL", pulumi.String("debug")))
```
pducolin, line 51: **💬 suggestion**
You can probably remove the log debug level here.

components/datadog/agent/docker.go
adel121: I moved the env vars related to dsd into the docker-compose of the dsd apps

pducolin (approved):

PR #501, opened 2023-12-05
[docker] Add support for fakeintake and logs in agent docker

pducolin:
What does this PR do?
---------------------

Refactoring docker Daemon to inject an UnixVM, instead of creating one inside it, and support some of the options available in the Agent VM installer.

This is a refactoring to move docker agent support closer to VM agent installer.

Best reviewed commit by commit.

-------------------

docker. ecs, eks code changed but with no effect

Motivation
----------

Some time ago @nplanel asked us to allow accessing the VM in a docker scenario: when a docker compose shares a folder with the host, we want to allow operating on the host.

At the same time we received few requests to support fakeintake in a docker scenario.

To allow all of the above, I moved the VM creation out of the `Daemon`. We can now call

```golang
daemon, err := Daemon(vm, options...)
```

to install a dockerized Agent on a UnixVm with docker installed. This is closer to what we do to install the agent on a VM

```golang
installer, err := agent.NewInstaller(vm, agentParameters...)
``` 

This will not change the generic usage in tests `datadog-agent`: the implementation change will still be wrapped inside a `DockerStackDef`. Changes in `datadog-agent` are in https://github.com/DataDog/datadog-agent/pull/21324 

Additional Notes
----------------


components/command/docker.go
```diff
@@ -17,6 +17,18 @@ const (
 	defaultTimeout = 300
 )
 
+type DockerComposeManifest struct {
+	Version  string                                  `yaml:"version"`
+	Services map[string]DockerComposeManifestService `yaml:"services"`
+}
+
+type DockerComposeManifestService struct {
```
adel121, line 25: I think it can be useful to add `Ports` and `Pid` properties here.

components/datadog/agent/docker.go
```diff
@@ -2,31 +2,122 @@ package agent
...
+	}
+
+	return &ClientData{Connection: vmData.Connection}, nil
+}
+
+func (d *Daemon) VM() vm.VM {
+	return d.vm.VM
+}
+
+func dockerAgentComposeContent(agentImagePath string, apiKey pulumi.StringInput, envVars pulumi.Map) pulumi.StringOutput {
```
vboulineau, line 87: I am not sure of the added value of the structured type as it's mixed with untyped content, but LGTM.

components/datadog/dockeragentparams/params.go
```diff
@@ -0,0 +1,188 @@
...
+		return nil
+	}
+}
+
+// WithIntake configures the agent to use the given hostname as intake.
+//
+// To use a fakeintake, see WithFakeintake.
+//
+// This option is overwritten by `WithFakeintake`.
+func WithIntake(hostname string, skipSSL bool) func(*Params) error {
```
vboulineau, line 96: `skipSSL` is not very clear, it would be more something like `tlsVerify`. But honestly you can forget about the parameter entirely, we'll ~never have truely valid certificate.

pducolin: In containers test currently we use double shipping to real DD and fakeintake by default. I imagine it's going to be the case for docker based tests, and it could be the case for VM based tests, as it makes it easier to investigate or test cross intake. Will change to `tlsVerify` as suggested !

components/datadog/agent/docker.go
pducolin: What do you mean by mixed up with untyped content ?

components/datadog/dockeragentparams/params.go
vboulineau: Ok so it is for DD target. I think we should rework a bit. Unfortunately the endpoint configuration is the Agent is really messy, but currently you hardcode `http` in some cases while you check if target is `https` for logs?

I don't think we want to support mixed http/https on the same endpoint.

Also you're checking `https` for `DD_LOGS_CONFIG_FORCE_USE_HTTP` but `DD_LOGS_CONFIG_FORCE_USE_HTTP` is to switch from TCP to HTTP/HTTPS. AFAIK logs with HTTP for main endpoint but not for additional endpoints? @L3n41c to confirm as you looked at it recently.

pducolin: Ah actually we have a `WithFaketintake` and `WithExtraIntake`, we can keep the ssl validation hidden indeed.

pducolin: Changed !

adel121 (approved):

PR #508, opened 2023-12-08
Use last version of install_script on all cases

chouetz:
What does this PR do?
---------------------
Remove the usage of `install_script.sh` which should not be used anymore, I guess

-------------------
All scenarios where the agent is installed

Motivation
----------
Maintenance

Additional Notes
----------------


pducolin:
cc @vboulineau you'll need to update it in your Component PR, adding a comment there with the suggestion

PR #509, opened 2023-12-08
Use latest version of the agent to monitor fakeintake

KevinFairise2:
What does this PR do?
---------------------

Change image used for the agent monitoring fake intake to latest agent image
Since the agent deployed next to the fakeintake is supposed to help to monitor the fakeintake it is better to rely on a stable version of the agent

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/ecsFargate.go
+++ b/components/datadog/agent/ecsFargate.go
@@ -10,7 +10,6 @@ import (
 
 func ECSFargateLinuxContainerDefinition(e config.CommonEnvironment, image string, apiKeySSMParamName pulumi.StringInput, fakeintake *ddfakeintake.ConnectionExporter, logConfig ecs.TaskDefinitionLogConfigurationPtrInput) *ecs.TaskDefinitionContainerDefinitionArgs {
 	var agentImage string
-
 	if image == "" {
 		agentImage = DockerAgentFullImagePath(&e, "public.ecr.aws/datadog/agent")
 	} else {
--- a/components/datadog/agent/ecsFargate.go
+++ b/components/datadog/agent/ecsFargate.go
@@ -8,54 +8,20 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-func ECSFargateLinuxContainerDefinition(e config.CommonEnvironment, apiKeySSMParamName pulumi.StringInput, fakeintake *ddfakeintake.ConnectionExporter, logConfig ecs.TaskDefinitionLogConfigurationPtrInput) *ecs.TaskDefinitionContainerDefinitionArgs {
+func ECSFargateLinuxContainerDefinition(e config.CommonEnvironment, image string, apiKeySSMParamName pulumi.StringInput, fakeintake *ddfakeintake.ConnectionExporter, logConfig ecs.TaskDefinitionLogConfigurationPtrInput) *ecs.TaskDefinitionContainerDefinitionArgs {
 
-	return &ecs.TaskDefinitionContainerDefinitionArgs{
-		Cpu:       pulumi.IntPtr(0),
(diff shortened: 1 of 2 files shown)


components/datadog/agent/ecsFargate.go
```diff
@@ -48,3 +49,45 @@ func ECSFargateLinuxContainerDefinition(e config.CommonEnvironment, apiKeySSMPar
 		VolumesFrom:      ecs.TaskDefinitionVolumeFromArray{},
 	}
 }
+
+func ECSFargateLinuxContainerDefinitionLatest(apiKeySSMParamName pulumi.StringInput, fakeintake *ddfakeintake.ConnectionExporter, logConfig ecs.TaskDefinitionLogConfigurationPtrInput) *ecs.TaskDefinitionContainerDefinitionArgs {
```
vboulineau, line 53: Let's just take `image` as parameter and fallback to `pulumi.Sprintf(DockerAgentFullImagePath(&e, "public.ecr.aws/datadog/agent"))` if parameter is empty.

KevinFairise2: I added an `image` parameter to the `ECSFargateLinuxContainerDefinition`. It may be better to do it with Optional function parameter so we can omit the `image` argument when calling the function

vboulineau: I am not sure we want to pass optional in internal APIs. It adds a lot of overhead for low value and we cannot pass `Pulumi.ResourceOption` anymore. I think that passing `struct` args around is the most concise and fits well with Pulumi patterns.

components/datadog/agent/ecsFargate.go
```diff
@@ -8,11 +8,20 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-func ECSFargateLinuxContainerDefinition(e config.CommonEnvironment, apiKeySSMParamName pulumi.StringInput, fakeintake *ddfakeintake.ConnectionExporter, logConfig ecs.TaskDefinitionLogConfigurationPtrInput) *ecs.TaskDefinitionContainerDefinitionArgs {
+func ECSFargateLinuxContainerDefinition(e config.CommonEnvironment, image string, apiKeySSMParamName pulumi.StringInput, fakeintake *ddfakeintake.ConnectionExporter, logConfig ecs.TaskDefinitionLogConfigurationPtrInput) *ecs.TaskDefinitionContainerDefinitionArgs {
+
```
vboulineau, line 12:
```suggestion
```

components/datadog/agent/ecsFargate.go
```diff
@@ -8,11 +8,20 @@ import (
...
+	var agentImage string
+
```
vboulineau, line 14:
```suggestion
```

vboulineau (approved):

components/datadog/agent/ecsFargate.go
```diff
@@ -8,11 +8,18 @@ import (
...
+		agentImage = DockerAgentFullImagePath(&e, "public.ecr.aws/datadog/agent")
+	} else {
+		agentImage = image
+	}
+
 	return &ecs.TaskDefinitionContainerDefinitionArgs{
 		Cpu:       pulumi.IntPtr(0),
 		Name:      pulumi.String("datadog-agent"),
-		Image:     pulumi.Sprintf(DockerAgentFullImagePath(&e, "public.ecr.aws/datadog/agent")),
+		Image:     pulumi.Sprintf(agentImage),
```
L3n41c, line 22: The `image` `string` parameter isn’t a `printf` pattern.

```suggestion
		Image:     pulumi.String(agentImage),
```

L3n41c: #519

PR #518, opened 2023-12-12
Make Windows installation work with WithPipelineID

KevinFairise2:
What does this PR do?
---------------------

Modify windows install to work with WithPipelineID

-------------------

`create-vm` 

Motivation
----------

Additional Notes
----------------


components/os/windows.go
```diff
@@ -170,6 +186,43 @@ func (f *agentURLFinder) findVersion(fullVersion string) (string, error) {
...
+	}
+
+	t := s3ListBucketResult{}
+	if err = xml.Unmarshal(body, &t); err != nil {
+		return "", err
+	}
+	fmt.Println("FOUND: ", t)
+	if len(t.Contents) == 0 {
+		return "", fmt.Errorf("no agent found for pipeline %v", pipelineID)
+	}
```
julien-lebot, line 211: I think we should use the [S3 SDK](https://pkg.go.dev/github.com/aws/aws-sdk-go/service/s3#hdr-Download_Manager) instead of parsing the result by hand

KevinFairise2: I hesitated, this one did not require to add a new dependency. But the code will be clearer using the SDK for sure

julien-lebot: For sure it's simple for now, but it puts the burden of maintaining compatibility with the format returned by S3 on us.
In the future S3 could break this API (I don't know if they have any agreement that they won't change this format, and I don't see any versioning in their HTTP API) but they would update their SDK for sure and it will be a simple matter of bumping the version for us.

KevinFairise2: Yes you are absolutely right, did not even think about changes to the API. I updated the code to use Go SDK :)

components/os/windows.go
```diff
@@ -170,6 +179,30 @@ func (f *agentURLFinder) findVersion(fullVersion string) (string, error) {
...
+	result, err := s3Client.ListObjectsV2(context.Background(), &s3.ListObjectsV2Input{
+		Bucket: aws.String("dd-agent-mstesting"),
+		Prefix: aws.String(fmt.Sprintf("pipelines/A7/%v", pipelineID)),
+	})
+
+	if err != nil {
+		return "", err
+	}
+
+	return "https://s3.amazonaws.com/dd-agent-mstesting/" + *result.Contents[0].Key, nil
```
julien-lebot, line 203: nit: I would check `len(result.Contents) > 0`

julien-lebot (approved):
LGTM, just one nit

components/os/windows.go
KevinFairise2: Oh yes, I had it and removed it when switching to Go SDK 🙈

components/os/windows.go
```diff
@@ -170,6 +179,34 @@ func (f *agentURLFinder) findVersion(fullVersion string) (string, error) {
 	return url, nil
 }
 
+func getAgentURLFromPipelineID(pipeline string) (string, error) {
+	// FIXME: remove pipeline- from the pipelineID we do not want it for Windows
```
pducolin, line 183: **❓ question**
Can you open a JIRA card to fix this ?

pducolin (approved):

components/os/windows.go
KevinFairise2: Sure will create it

PR #520, opened 2023-12-12
enable mergequeue on repo

adel121:
What does this PR do?
---------------------

Enable mergequeue on the repo

-------------------
None

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/repository.datadog.yml
+++ b/repository.datadog.yml
@@ -1,0 +1,5 @@
+---
+schema-version: v1
+kind: mergequeue
+team: ci-interfaces
+gitlab_check_enable: false


KevinFairise2 (commented):
I think the merge queue is already enabled even without this config file. You can already use it by writing `/merge` or with the Doggo extension merge button.

Unless you want specific configuration for the merge queue, I think we do not need to add this file. If we really want to commit the config file we should set `team:` to the team owning the repository, it should be `agent-platform` here I think
Also, I am wondering why you set `gitlab_check_enable: false`?

PR #524, opened 2023-12-13
Default deploy fakeintake without LB on VM stack

Pythyu:
What does this PR do?
---------------------

Set the `LoadBalancerEnabled` to `false` by default. Adding `WithLoadBalancer` method to enable the loadbalancer.

-------------------

Motivation
----------

Additional Notes
----------------

Modifying existing tests to work with these changes on [PR#21535](https://github.com/DataDog/datadog-agent/pull/21535)

### DIFF
--- a/scenarios/aws/fakeintake/fakeintakeparams/params.go
+++ b/scenarios/aws/fakeintake/fakeintakeparams/params.go
@@ -13,17 +13,17 @@ type Option = func(*Params) error
 // NewParams returns a new instance of Fakeintake Params
 func NewParams(options ...Option) (*Params, error) {
 	params := &Params{
-		LoadBalancerEnabled: true,
+		LoadBalancerEnabled: false,
 		Name:                "fakeintake",
 		ImageURL:            "public.ecr.aws/datadog/fakeintake:latest",
 	}
 	return common.ApplyOption(params, options)
 }
 
-// WithoutLoadBalancer disable load balancer in front of the fakeintake
-func WithoutLoadBalancer() Option {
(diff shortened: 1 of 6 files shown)


scenarios/aws/fakeintake/fakeintakeparams/params.go
```diff
@@ -13,17 +13,17 @@ type Option = func(*Params) error
 // NewParams returns a new instance of Fakeintake Params
 func NewParams(options ...Option) (*Params, error) {
 	params := &Params{
-		LoadBalancerEnabled: true,
+		LoadBalancerEnabled: false,
```
pducolin, line 16: **🔨 warning**
This will make LB for fakeintake disabled by default for all scenarios, while we would like to skip fakeintake only for VM scenarios. 

We could keep `test-infra` as is, and always apply `WithoutLoadBalancer` on stack definitions that wrap VM environments

Pythyu: Since we have the current limit of 32 loadbalancer, I thought safer to disable it by default all together to assure to only use it when when needed. What do you think ? Otherwise I'll just disable it from the VM wrapper

Pythyu: Already being tested in [PR#21535](https://github.com/DataDog/datadog-agent/pull/21535)

Pythyu: Thanks ! 👀 I'll change that

PR #534, opened 2023-12-21
feat(vm): add a new option to select instance type

chouetz:
What does this PR do?
---------------------
Minimal implementation to be able to select different instance type with the cli

-------------------
AWS EC2

Motivation
----------
Add more flexibility in the cli

Additional Notes
----------------
* We could cover azure VM
* I tested to start a `t3.small`, would deserve more tests (maybe automated)
* If I launch `inv create-vm -t t3.small` then `inv create-vm` it is considered as "no-change" whereas I guess it should


chouetz:
> The option already exists without any change: `ddinfra:aws/defaultInstanceType` see [`main`/resources/aws/environment.go#L21-L22](https://github.com/DataDog/test-infra-definitions/blob/main/resources/aws/environment.go?rgh-link-date=2023-12-27T13%3A14%3A40Z#L21-L22)

Hi @vboulineau when you say `without any change` do you mean we should dismiss this change or should I adapt it? It would probably require some dedicated documentation, wdyt?

vboulineau:
You can add the bits in `tasks/vm.py` but you don't need to add a dedicated parameter for it.

Personally I'd rather have an option on the `invoke` tasks that can set any parameter and a generated doc for all existing parameters, but it's not the current way of doing.

chouetz:
> You can add the bits in `tasks/vm.py` but you don't need to add a dedicated parameter for it.
> 
> Personally I'd rather have an option on the `invoke` tasks that can set any parameter and a generated doc for all existing parameters, but it's not the current way of doing.

we could also use a key-value config file to have only one option?
Thanks in advance for your review @vboulineau

vboulineau:
> > You can add the bits in `tasks/vm.py` but you don't need to add a dedicated parameter for it.
> > Personally I'd rather have an option on the `invoke` tasks that can set any parameter and a generated doc for all existing parameters, but it's not the current way of doing.
> 
> we could also use a key-value config file to have only one option? Thanks in advance for your review @vboulineau

We could but it was not the path chosen as it was not user friendly enough.

vboulineau (approved):

PR #536, opened 2023-12-22
Dual-ship `contimage`, `contlcycle` and `sbom` payloads to the fakeintake

L3n41c:
What does this PR do?
---------------------

Configure the agent on Kubernetes to dual-ship `container image`, `container lifecycle events` and `SBOM` payloads to the fake intake.

-------------------

* `aws/eks`
* `aws/kind`

Motivation
----------

Allow to write e2e tests for the
* `containerimage`
* `contlcycle`
* `sbom`
checks.

Additional Notes
----------------

The fakeintake client has been updated to support those new payloads in DataDog/datadog-agent#21742.

### DIFF
--- a/components/datadog/agent/helm.go
+++ b/components/datadog/agent/helm.go
@@ -197,6 +197,34 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 			"prometheusScrape": pulumi.Map{
 				"enabled": pulumi.Bool(true),
 			},
+			"sbom": pulumi.Map{
+				"containerImage": pulumi.Map{
+					"enabled": pulumi.Bool(true),
+				},
+			},
+			// The fake intake keeps payloads only for a hardcoded period of 15 minutes.
+			// https://github.com/DataDog/datadog-agent/blob/34922393ce47261da9835d7bf62fb5e090e5fa55/test/fakeintake/server/server.go#L81
+			// So, we need `container_image` and `sbom` checks to resubmit their payloads more frequently than that.
+			"confd": pulumi.Map{
+				"container_image.yaml": pulumi.String(utils.JSONMustMarshal(map[string]interface{}{
+					"ad_identifiers": []string{"_container_image"},
+					"init_config":    map[string]interface{}{},
+					"instances": []map[string]interface{}{
+						{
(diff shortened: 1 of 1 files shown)


components/datadog/agent/helm.go
```diff
@@ -197,6 +197,34 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 			"prometheusScrape": pulumi.Map{
 				"enabled": pulumi.Bool(true),
 			},
+			"sbom": pulumi.Map{
+				"containerImage": pulumi.Map{
+					"enabled": pulumi.Bool(true),
+				},
+			},
+			// The fake intake keeps payloads only for a hardcoded period of 15 minutes.
+			// https://github.com/DataDog/datadog-agent/blob/34922393ce47261da9835d7bf62fb5e090e5fa55/test/fakeintake/server/server.go#L81
```
KevinFairise2, line 206: We have a function https://github.com/DataDog/datadog-agent/blob/34922393ce47261da9835d7bf62fb5e090e5fa55/test/fakeintake/server/server.go#L163C8-L171 to manually set the retention time on the fakeintake. It is not yet exposed, but some lines could be easily changed to be able to set the retention time when running the fakeintake (with environment variables or a flag).
If you think it would be better to update retention time than to submit payload more frequently, I could do the change

L3n41c: I think it doesn’t change a lot of things.

Even if we can configure the fake intake retention time, we will always need to have one in order to not have the memory consumption of the fake intake increasing forever.
And as soon as the payloads may expire in the fake intake, we need to resubmit them in order to have the tests be guaranteed to find them.

The only thing we could do would be to set the fake intake retention time to make it consistent with the *default value* of `periodic_refresh_seconds` so that we don’t need to override it explicitly like done here.

But that would only make things less explicit.

The best would be to have the fake intake retention time be an output of the Pulumi resource and have it used to dynamically set the `periodic_refresh_seconds` parameters of the `sbom` and `container_image` checks.
But I felt it would be overkill for now.

KevinFairise2: Yes I agree with you, probably too much work for what it is worth!

KevinFairise2 (approved):

PR #543, opened 2023-12-29
[CONTINT-3593] Support creating kata-enabled AKS cluster

AliDatadog:
What does this PR do?
---------------------
This PR adds support of kata-enabled AKS clusters. It creates a new nodepool for kata runtimes.

-------------------
az/aks

Motivation
----------
We would like to add e2e tests for kata enabled runtime

Additional Notes
----------------

### DIFF
--- a/resources/azure/environment.go
+++ b/resources/azure/environment.go
@@ -22,8 +22,7 @@ const (
 	DDInfraDefaultPublicKeyPath            = "az/defaultPublicKeyPath"
 	DDInfraDefaultPrivateKeyPath           = "az/defaultPrivateKeyPath"
 	DDInfraDefaultPrivateKeyPassword       = "az/defaultPrivateKeyPassword"
-	DDInfraAksDeployKata                   = "az/aks/deployKata"
-	DDInfraAksLinuxKataInstanceType        = "az/aks/linuxKataInstanceType"
+	DDInfraAksLinuxKataNodeGroup           = "az/aks/linuxKataNodeGroup"
 )
 
 type Environment struct {
@@ -100,12 +99,7 @@ func (e *Environment) GetCommonEnvironment() *config.CommonEnvironment {
 	return e.CommonEnvironment
 }
 
-// DeployKata Whether to deploy a kata node pool
-func (e *Environment) DeployKata() bool {
-	return e.GetBoolWithDefault(e.InfraConfig, DDInfraAksDeployKata, e.envDefault.ddInfra.aks.defaultDeployKata)
-}
-
-// KataInstanceType returns the instance type that should be used for kata nodes
-func (e *Environment) KataInstanceType() string {
-	return e.GetStringWithDefault(e.InfraConfig, DDInfraAksLinuxKataInstanceType, e.envDefault.ddInfra.aks.defaultKataInstanceType)
+// LinuxKataNodeGroup Whether to deploy a kata node pool
+func (e *Environment) LinuxKataNodeGroup() bool {
+	return e.GetBoolWithDefault(e.InfraConfig, DDInfraAksLinuxKataNodeGroup, e.envDefault.ddInfra.aks.linuxKataNodeGroup)
 }
--- a/resources/azure/environment.go
+++ b/resources/azure/environment.go
@@ -22,9 +22,8 @@ const (
 	DDInfraDefaultPublicKeyPath            = "az/defaultPublicKeyPath"
 	DDInfraDefaultPrivateKeyPath           = "az/defaultPrivateKeyPath"
 	DDInfraDefaultPrivateKeyPassword       = "az/defaultPrivateKeyPassword"
-	DDInfraDefaultDeployKata               = "az/defaultDeployKata"
-	DDInfraDefaultKataRuntime              = "az/defaultKataRuntime"
-	DDInfraDefaultKataOsSku                = "az/defaultKataOsSku"
+	DDInfraAksDeployKata                   = "az/aks/deployKata"
+	DDInfraAksLinuxKataInstanceType        = "az/aks/linuxKataInstanceType"
 )
 
 type Environment struct {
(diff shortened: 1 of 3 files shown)


common/config/environment.go
```diff
@@ -29,6 +29,7 @@ const (
 	DDInfraDeployFakeintakeWithLoadBalancer = "deployFakeintakeWithLoadBalancer"
 	DDInfraOSAmiID                          = "osAmiId"
 	DDInfraExtraResourcesTags               = "extraResourcesTags"
+	DDInfraContainerRuntime                 = "containerRuntime"
```
AliDatadog, line 32: Put in common configuration as it's possible that GKE and EKS add support for kata. Minikube also supports it. It can also be ran without kubernetes.

clamoriniere: how do you see the value? for example `containerd_runc` vs `containerd_kata` ? or should we have 2 variables: `ContainerRuntime` and `ContainerRuntimeFlavor` ?

resources/azure/environment.go
```diff
@@ -97,3 +98,25 @@ func (e *Environment) DefaultPrivateKeyPassword() string {
...
 }
+
+// ContainerRuntime Determines the type of workload a node can run.
+// Note that KataMshvVmIsolation is supported only on Mariner Sku.
+// Allowed values: KataCcIsolation, KataMshvVmIsolation, OCIContainer, WasmWasi.
+// Default : OCIContainer.
+func (e *Environment) ContainerRuntime() string {
+	env := e.CommonEnvironment.ContainerRuntime()
+	switch env {
+	case "kata":
```
clamoriniere, line 109: IMO you can define a `const string` for the value `kata`

common/config/environment.go
AliDatadog: I think those two should be seperate concerns. This field could potentially be used outside of kubernetes, if we ever decide to implement some e2e tests using the containerd/docker socket.

resources/azure/environment.go
AliDatadog: Done

AliDatadog:
After looking at what is done in other scenarios, I think we can potentially just add another nodepool for kata instead of modifying the system one

clamoriniere (approved):
Nice work @AliDatadog 👏

resources/azure/environmentDefaults.go
```diff
@@ -43,8 +46,11 @@ func sandboxDefault() environmentDefault {
 			defaultVNet:            "/subscriptions/8c56d827-5f07-45ce-8f2b-6c5001db5c6f/resourceGroups/datadog-agent-testing/providers/Microsoft.Network/virtualNetworks/default-vnet",
 			defaultSubnet:          "/subscriptions/8c56d827-5f07-45ce-8f2b-6c5001db5c6f/resourceGroups/datadog-agent-testing/providers/Microsoft.Network/virtualNetworks/default-vnet/subnets/default-subnet",
 			defaultSecurityGroup:   "/subscriptions/8c56d827-5f07-45ce-8f2b-6c5001db5c6f/resourceGroups/datadog-agent-testing/providers/Microsoft.Network/networkSecurityGroups/default",
-			defaultInstanceType:    "Standard_B4ms",
-			defaultARMInstanceType: "Standard_D4ps_v5",
+			defaultInstanceType:    "Standard_D4s_v3",  // Allows nested virtualization for kata runtimes
+			defaultARMInstanceType: "Standard_D4ps_v5", // No azure arm instance supports nested virtualization
+			defaultKataOsSku:       "Mariner",
```
vboulineau, line 51: Generally speaking, these options are AKS specific. They should in their own namespace (see `eks` on AWS): `az/aks/linuxKataNodeGroup`

The doc says it only supports `AzureLinux`: https://learn.microsoft.com/en-us/azure/aks/use-pod-sandboxing#deploy-to-an-existing-cluster. Also I'm not sure we need to input an option for that.

resources/azure/environmentDefaults.go
```diff
@@ -43,8 +46,11 @@ func sandboxDefault() environmentDefault {
...
+			defaultDeployKata:      true,
+			defaultKataRuntime:     "KataMshvVmIsolation", // Azure also supports KataCcIsolation
```
vboulineau, line 53: `KataCcIsolation` requires a whole setup with OIDC and Identity mapping. None of that being implemented, I would not support this option.

resources/azure/environmentDefaults.go
AliDatadog: I think Mariner also works. If I remember correctly I tried with the default setting and it said something about only `Mariner` being supported.

resources/azure/environmentDefaults.go
AliDatadog: Hardcoded it in the last commit

resources/azure/environmentDefaults.go
AliDatadog: I think both Mariner and AzureLinux are supported. I removed this option and hardcoded AzureLinux.

resources/azure/environmentDefaults.go
```diff
@@ -21,6 +21,12 @@ type ddInfra struct {
 	defaultSecurityGroup   string
 	defaultInstanceType    string
 	defaultARMInstanceType string
+	aks                    ddInfraAks
+}
+
+type ddInfraAks struct {
+	defaultDeployKata       bool
```
vboulineau, line 28:
```suggestion
	linuxKataNodegroup       bool
```

resources/azure/environmentDefaults.go
```diff
@@ -43,8 +49,12 @@ func sandboxDefault() environmentDefault {
 			defaultVNet:            "/subscriptions/8c56d827-5f07-45ce-8f2b-6c5001db5c6f/resourceGroups/datadog-agent-testing/providers/Microsoft.Network/virtualNetworks/default-vnet",
 			defaultSubnet:          "/subscriptions/8c56d827-5f07-45ce-8f2b-6c5001db5c6f/resourceGroups/datadog-agent-testing/providers/Microsoft.Network/virtualNetworks/default-vnet/subnets/default-subnet",
 			defaultSecurityGroup:   "/subscriptions/8c56d827-5f07-45ce-8f2b-6c5001db5c6f/resourceGroups/datadog-agent-testing/providers/Microsoft.Network/networkSecurityGroups/default",
-			defaultInstanceType:    "Standard_B4ms",
-			defaultARMInstanceType: "Standard_D4ps_v5",
+			defaultInstanceType:    "Standard_D4ps_v5",
```
vboulineau, line 52:
```suggestion
			defaultInstanceType:    "Standard_D4s_v5",
```

resources/azure/environmentDefaults.go
```diff
@@ -43,8 +49,12 @@ func sandboxDefault() environmentDefault {
...
+			defaultARMInstanceType: "Standard_D4ps_v5", // No azure arm instance supports nested virtualization
+			aks: ddInfraAks{
+				defaultKataInstanceType: "Standard_D4s_v3", // Allows nested virtualization for kata runtimes
```
vboulineau, line 55: I think you can remove that as well and set a default that supports nested virtualization (done above)

AliDatadog: Done in the last commit

vboulineau (approved):

PR #545, opened 2023-12-29
[CONTINT-3593] Add dummy workload to the aks scenario

AliDatadog:
What does this PR do?
---------------------
This PR adds a dummy workload to the aks scenario.

-------------------
az/aks

Motivation
----------
This is a first step towards implementing kata e2e tests.

Additional Notes
----------------


components/datadog/agent/helm.go
```diff
@@ -197,6 +197,9 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 			"prometheusScrape": pulumi.Map{
 				"enabled": pulumi.Bool(true),
 			},
+			"kubelet": pulumi.Map{
+				"tlsVerify": pulumi.Bool(false),
+			},
```
clamoriniere, line 202: can we have this setting optionnel? 
IMO we should disable `tlsVerify` only if the k8s destribution doesn't have have a proper certificat, but by default consider that the certificat is valid.
Also it seems that for AKS this setting is not anymore needed see: https://docs.datadoghq.com/containers/kubernetes/distributions?tab=helm#aks-kubelet-certificate

L3n41c: I agree that `kubelet.tlsVerify: false` shouldn’t be the default.
It should be needed only for Kind.

common/utils/kubernetes.go
```diff
@@ -2,17 +2,24 @@ package utils
...
 	"encoding/json"
-
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"gopkg.in/yaml.v3"
 )
 
 func KubeconfigToJSON(kubeConfig pulumi.Output) pulumi.StringInput {
-	return kubeConfig.ApplyT(func(config interface{}) (string, error) {
-		jsonConfig, err := json.Marshal(config)
+	return kubeConfig.ApplyT(func(config string) (string, error) {
```
L3n41c, line 10: This changes the signature of the `KubeconfigToJSON` function.
Won’t it break all the [callers](https://github.com/DataDog/test-infra-definitions/blob/c45b6947c391d5d70eac6fde60e1711b88925e0c/scenarios/aws/eks/run.go#L169) that are not updated in this PR ?

components/datadog/agent/helm.go
AliDatadog: Thank you for the info. I found a way to add environment-specific fields in the helm chart. It's done in this commit: 60cdcefbccd5e45c555fcf82c508456661f8a041.

scenarios/azure/aks/run.go
```diff
@@ -13,11 +23,82 @@ func Run(ctx *pulumi.Context) error {
...
 
-	_, kubeConfig, err := aks.NewCluster(env, "aks", nil)
+	cluster, kubeConfig, err := aks.NewCluster(env, "aks", nil)
 	if err != nil {
 		return err
 	}
 
+	// Export cluster’s properties
 	ctx.Export("kubeconfig", kubeConfig)
+	ctx.Export("kube-cluster-name", cluster.Name.ApplyT(func(value string) (string, error) { return value, nil }))
```
L3n41c, line 33: Why do we need an `ApplyT` here ?
The function looks like identity.

```suggestion
	ctx.Export("kube-cluster-name", cluster.Name)
```

common/utils/kubernetes.go
AliDatadog: I think the function name wasn't changed. The only change should be that we unmarshal as yaml first.

vboulineau: Your change assumes that `kubeConfig` is a YAML string.
In this case, you should change the signature to take in a ` kubeConfigYAML pulumi.StringOutput` and perhaps change the name of the function to `KubeconfigYAMLToJSON` as it does not do exactly what it says.

scenarios/azure/aks/run.go
```diff
@@ -13,11 +23,98 @@ func Run(ctx *pulumi.Context) error {
...
 	}
 
-	_, kubeConfig, err := aks.NewCluster(env, "aks", nil)
+	cluster, kubeConfig, err := aks.NewCluster(env, "aks", nil)
 	if err != nil {
 		return err
 	}
 
+	// Export cluster’s properties
 	ctx.Export("kubeconfig", kubeConfig)
```
vboulineau, line 32: I'd suggest to pu these modifications after https://github.com/DataDog/test-infra-definitions/pull/484 to directly implement the Kubernetes component for AKS.

AliDatadog: Do you have an estimation of when it should be merged ?

common/utils/kubernetes.go
AliDatadog: I did it in the last commit

scenarios/azure/aks/run.go
```diff
@@ -13,11 +25,105 @@ func Run(ctx *pulumi.Context) error {
...
 	}
 
-	_, kubeConfig, err := aks.NewCluster(env, "aks", nil)
+	clusterComp, err := components.NewComponent(*env.CommonEnvironment, env.Namer.ResourceName("aks"), func(comp *kubeComp.Cluster) error {
+		cluster, kubeConfig, err := aks.NewCluster(env, "aks", nil)
+		if err != nil {
+			return err
+		}
+
+		ctx.Export("kubeconfig", kubeConfig)
```
vboulineau, line 34: Not necessary anymore, you will export `clusterComp` later

scenarios/azure/aks/run.go
```diff
@@ -13,11 +25,105 @@ func Run(ctx *pulumi.Context) error {
...
+		// Filling Kubernetes component from EKS cluster
+		comp.ClusterName = cluster.Name
+		comp.KubeConfig = kubeConfig
+
+		// Building Kubernetes provider
+		aksKubeProvider, err := kubernetes.NewProvider(env.Ctx, env.Namer.ResourceName("k8s-provider"), &kubernetes.ProviderArgs{
```
vboulineau, line 40: The provider needs to be store inside `comp`.

AliDatadog: I didn't see it for other scenarios but I added it now

go.mod
```diff
@@ -35,7 +35,7 @@ require (
 require (
 	dario.cat/mergo v1.0.0 // indirect
 	github.com/Microsoft/go-winio v0.6.1 // indirect
-	github.com/ProtonMail/go-crypto v1.0.0 // indirect
+	github.com/ProtonMail/go-crypto v0.0.0-20230828082145-3c4c8a2d2371 // indirect
```
vboulineau, line 38: I guess it should not be present

AliDatadog: I bumped it again to v1.0.0 👍

vboulineau (approved):

PR #549, opened 2024-01-05
feat(logging): fakeintake is an agent option in VM scenario. 

chouetz:
As a consequence, --no-install-agent and --use-fakeintake are currently incompatible

What does this PR do?
---------------------
Add a warning when using incompatible options

-------------------
VM

Motivation
----------
Sanity

Additional Notes
----------------


KevinFairise2 (approved):
Thanks!

pducolin:
This should be possible though, maybe will fix later

PR #550, opened 2024-01-05
feat(vm): clean the known hosts to prevent possible collision

chouetz:
What does this PR do?
---------------------
Remove entries in known_hosts related to the "to-be-destroyed-VM" IP

-------------------
VM

Motivation
----------
I sometime faced already existing IPs in my known_hosts. I believe it's mostly harmless to remove them, I put it enabled by default, can be disabled if needed

Additional Notes
----------------


KevinFairise2 (approved):
Happens to me all the time, thanks!

PR #551, opened 2024-01-05
add orch options to helm chart for e2e

fisherevans:


components/datadog/agent/helm.go
```diff
@@ -114,6 +115,9 @@ func NewHelmInstallation(e config.CommonEnvironment, args HelmInstallationArgs,
 	values := buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAgentImagePath, clusterAgentImageTag, randomClusterAgentToken.Result)
 	values.configureImagePullSecret(imgPullSecret)
 	values.configureFakeintake(args.Fakeintake)
+	if args.EnableOrchestrator {
```
vboulineau, line 118: Orchestrator is enabled by default, what you want is a custom config.

components/datadog/agent/helm.go
```diff
@@ -21,11 +21,12 @@ const (
...
-	Namespace     string
-	ValuesYAML    pulumi.AssetOrArchiveArrayInput
-	Fakeintake    *ddfakeintake.ConnectionExporter
-	DeployWindows bool
+	KubeProvider       *kubernetes.Provider
+	Namespace          string
+	ValuesYAML         pulumi.AssetOrArchiveArrayInput
+	Fakeintake         *ddfakeintake.ConnectionExporter
+	DeployWindows      bool
+	EnableOrchestrator bool
```
vboulineau, line 29: We don't want to add all kind of possible options here. The `ValuesYAML` is here to allow setting the custom configuration you want.

fisherevans: Ah! Thank you. I misunderstood how that worked

PR #580, opened 2024-01-25
Add instance profile and configure creds helper to authorize pulling docker images from ECR

KevinFairise2:
What does this PR do?
---------------------

Configure docker scenario to add an Instance Profile that allows pulling docker image from QA env ECR. Also configures the aws-ecr-credentials-helper

Related PR on `test-infra-cleaner`: https://github.com/DataDog/test-infra-cleaner/pull/16
Related PR on `datadog-agent`: https://github.com/DataDog/datadog-agent/pull/22373

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/docker/docker.go
+++ b/components/docker/docker.go
@@ -155,7 +155,24 @@ func (d *Manager) install() (*remote.Command, error) {
 		return nil, err
 	}
 
-	return groupCmd, err
+	ecrCredsHelperInstall, err := d.host.OS.PackageManager().Ensure("amazon-ecr-credential-helper", utils.MergeOptions(d.opts, utils.PulumiDependsOn(groupCmd))...)
+	if err != nil {
+		return nil, err
+	}
+
+	ecrCredsHelperConfig, err := d.host.OS.Runner().Command(
+		d.namer.ResourceName("ecr-config"),
+		&command.Args{
+			Create: pulumi.Sprintf("mkdir -p ~/.docker && echo '{\"credsStore\": \"ecr-login\"}' > ~/.docker/config.json"),
+			Sudo:   false,
+		},
+		utils.MergeOptions(d.opts, utils.PulumiDependsOn(ecrCredsHelperInstall))...,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return ecrCredsHelperConfig, err
 }
 
 func (d *Manager) installCompose() (*remote.Command, error) {
--- a/components/docker/docker.go
+++ b/components/docker/docker.go
@@ -155,24 +155,7 @@ func (d *Manager) install() (*remote.Command, error) {
 		return nil, err
 	}
 
-	ecrCredsHelperInstall, err := d.host.OS.PackageManager().Ensure("amazon-ecr-credential-helper", utils.MergeOptions(d.opts, utils.PulumiDependsOn(groupCmd))...)
-	if err != nil {
-		return nil, err
-	}
-
-	ecrCredsHelperConfig, err := d.host.OS.Runner().Command(
-		d.namer.ResourceName("ecr-config"),
-		&command.Args{
-			Create: pulumi.Sprintf("mkdir -p ~/.docker && echo '{\"credsStore\": \"ecr-login\"}' > ~/.docker/config.json"),
-			Sudo:   false,
-		},
-		utils.MergeOptions(d.opts, utils.PulumiDependsOn(ecrCredsHelperInstall))...,
-	)
-	if err != nil {
-		return nil, err
-	}
-
-	return ecrCredsHelperConfig, err
+	return groupCmd, err
 }
 
 func (d *Manager) installCompose() (*remote.Command, error) {
--- a/resources/aws/environment.go
+++ b/resources/aws/environment.go
@@ -15,18 +15,18 @@ const (
 	awsRegionParamName = "region"
 
 	// AWS Infra
-	DDInfraDefaultVPCIDParamName                 = "aws/defaultVPCID"
-	DDInfraDefaultSubnetsParamName               = "aws/defaultSubnets"
-	DDInfraDefaultSecurityGroupsParamName        = "aws/defaultSecurityGroups"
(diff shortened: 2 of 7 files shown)


components/docker/docker.go
```diff
@@ -155,7 +155,24 @@ func (d *Manager) install() (*remote.Command, error) {
 		return nil, err
 	}
 
-	return groupCmd, err
+	ecrCredsHelperInstall, err := d.host.OS.PackageManager().Ensure("amazon-ecr-credential-helper", utils.MergeOptions(d.opts, utils.PulumiDependsOn(groupCmd))...)
```
vboulineau, line 158: Given that it's limited to AWS scenarios, it should not be in the common component but in `scenario/aws`.

resources/aws/environment.go
```diff
@@ -15,17 +15,18 @@ const (
...
-	DDinfraDefaultPublicKeyPath            = "aws/defaultPublicKeyPath"
-	DDInfraDefaultPrivateKeyPath           = "aws/defaultPrivateKeyPath"
-	DDInfraDefaultPrivateKeyPassword       = "aws/defaultPrivateKeyPassword"
-	DDInfraDefaultInstanceStorageSize      = "aws/defaultInstanceStorageSize"
-	DDInfraDefaultShutdownBehavior         = "aws/defaultShutdownBehavior"
+	DDInfraDefaultVPCIDParamName                 = "aws/defaultVPCID"
+	DDInfraDefaultSubnetsParamName               = "aws/defaultSubnets"
+	DDInfraDefaultSecurityGroupsParamName        = "aws/defaultSecurityGroups"
+	DDInfraDefaultInstanceTypeParamName          = "aws/defaultInstanceType"
+	DDInfraDefaultDockerInstanceProfileParamName = "aws/defaultInstanceProfile"
```
vboulineau, line 22: I don't think it should be specific to Docker. The need to specify an instanceProfile can be generic.

resources/aws/environmentDefaults.go
```diff
@@ -75,13 +76,14 @@ func sandboxDefault() environmentDefault {
...
-			defaultSecurityGroups:      []string{"sg-46506837", "sg-7fedd80a", "sg-0e952e295ab41e748"},
-			defaultInstanceType:        "t3.medium",
-			defaultARMInstanceType:     "t4g.medium",
-			defaultInstanceStorageSize: 200,
-			defaultShutdownBehavior:    "stop",
+			defaultVPCID:                 "vpc-d1aac1a8",
+			defaultSubnets:               []string{"subnet-b89e00e2", "subnet-8ee8b1c6", "subnet-3f5db45b"},
+			defaultSecurityGroups:        []string{"sg-46506837", "sg-7fedd80a", "sg-0e952e295ab41e748"},
+			defaultInstanceType:          "t3.medium",
+			defaultDockerInstanceProfile: "agent-ecr-read",
```
vboulineau, line 83: As I did not see a PR on `infra-cleaner`, I assume it's deleted already.
Also I don't think we need a role per use case. It could be better to have a default `vmInstanceRole` as we have a default `ecsInstanceRole` below.

scenarios/aws/ec2/vm_run.go
```diff
@@ -75,16 +75,16 @@ func VMRunWithDocker(ctx *pulumi.Context) error {
 
 	// If no OS is provided, we default to AmazonLinuxECS as it ships with Docker pre-installed
 	osDesc := os.DescriptorFromString(env.InfraOSDescriptor(), os.AmazonLinuxECS)
-	vm, err := NewVM(env, "vm", WithAMI(env.InfraOSImageID(), osDesc, osDesc.Architecture))
+	vm, err := NewVM(env, "vm", WithAMI(env.InfraOSImageID(), osDesc, osDesc.Architecture), WithInstanceProfile(env.DefaultDockerInstanceProfile()))
```
vboulineau, line 78: It should be defaulted in `resources/aws/ec2/vm.go` instead as I don't think we need to only use it for Docker case.

resources/aws/environmentDefaults.go
KevinFairise2: If we are fine with VM outside of Docker scenario created with this role, I'll change it

components/docker/docker.go
KevinFairise2: Oh yes, you're right. Moved.

scenarios/aws/ec2/vm_run.go
```diff
@@ -83,8 +85,23 @@ func VMRunWithDocker(ctx *pulumi.Context) error {
...
 
-	_, installDocker := osWithDockerProvided[vm.OS.Descriptor().Flavor]
-	manager, _, err := docker.NewManager(*env.CommonEnvironment, vm, installDocker)
+	_, isDockerInstalled := osWithDockerProvided[vm.OS.Descriptor().Flavor]
+	manager, setupDockerCmd, err := docker.NewManager(*env.CommonEnvironment, vm, !isDockerInstalled)
+	if err != nil {
+		return err
+	}
+
+	ecrCredsHelperInstall, err := vm.OS.PackageManager().Ensure("amazon-ecr-credential-helper", utils.PulumiDependsOn(setupDockerCmd))
```
vboulineau, line 94: You will need to do this in `datadog-agent` as well I believe, so perhaps it should be put in a dedicated function.
Also we should perhaps revert the dependencies, you can install the helper before Docker, and set it as `DependsOn()` when creating `docker.NewManager`, so that any subsequent call does not need to depend on `ecrConfigCommand`.

vboulineau (approved):

scenarios/aws/ec2/vm.go
```diff
@@ -53,11 +57,34 @@ func NewVM(e aws.Environment, name string, params ...VMOption) (*remote.Host, er
 	})
 }
 
+func InstallEcrCredentialsHelper(e aws.Environment, vm *remote.Host) (*goremote.Command, error) {
```
vboulineau, line 60: nit

```suggestion
func InstallECRCredentialsHelper(e aws.Environment, vm *remote.Host) (*goremote.Command, error) {
```

PR #581, opened 2024-01-25
docker agent compose up running in privileged, required by system-probe

nplanel:
What does this PR do?
---------------------

Run docker agent in privileged mode, it's required to run system-probe properly
mount /sys/kernel/tracing from the host to the container as well

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/docker.go
+++ b/components/datadog/agent/docker.go
@@ -1,6 +1,8 @@
 package agent
 
 import (
+	"strings"
+
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	"gopkg.in/yaml.v3"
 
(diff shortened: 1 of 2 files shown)


components/datadog/agent/docker.go
```diff
@@ -65,13 +65,15 @@ func dockerAgentComposeManifest(agentImagePath string, apiKey pulumi.StringInput
 			Version: "3.9",
 			Services: map[string]docker.ComposeManifestService{
 				"agent": {
+					Privileged:    true,
```
pducolin, line 68: **💬 suggestion**
As the default agent config does not require to run with privileged rights, I suggest we have this as an option for when `system-probe` is enabled

pducolin (changes requested):

components/datadog/agent/docker.go
```diff
@@ -58,20 +60,30 @@ func NewDockerAgent(e config.CommonEnvironment, vm *remoteComp.Host, manager *do
 }
 
 func dockerAgentComposeManifest(agentImagePath string, apiKey pulumi.StringInput, envVars pulumi.Map) docker.ComposeInlineManifest {
+	runInPrivileged := false
+	for k, _ := range envVars {
+		if strings.HasPrefix(k, "DD_SYSTEM_PROBE_") {
```
pducolin, line 65: **👏 praise**
🎉

pducolin (approved):

PR #582, opened 2024-01-25
components/datadog/agent: add WithPulumiResourceOptions option for agent creation.

knusbaum:
What does this PR do?
---------------------

This commit adds the `WithPulumiResourceOptions` option to the `agentparams` package, allowing users to add
resource options such as dependencies to the agent installation.

-------------------

This will only affect scenarios which start using the `WithPulumiResourceOptions` scenario. Other scenarios will see no difference.

Motivation
----------

Sometimes installing other things along side the agent (e.g. docker) can interfere with each other due to their use of `apt` (and possibly other resources). It is useful to create a dependency link between them so that they don't block each other.

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/host.go
+++ b/components/datadog/agent/host.go
@@ -37,7 +37,7 @@ func (h *HostAgent) Export(ctx *pulumi.Context, out *HostAgentOutput) error {
 }
 
 // NewHostAgent creates a new instance of a on-host Agent
-func NewHostAgent(e *config.CommonEnvironment, host *remoteComp.Host, dep pulumi.Resource, options ...agentparams.Option) (*HostAgent, error) {
+func NewHostAgent(e *config.CommonEnvironment, host *remoteComp.Host, options ...agentparams.Option) (*HostAgent, error) {
 	hostInstallComp, err := components.NewComponent(*e, host.Name(), func(comp *HostAgent) error {
 		comp.namer = e.CommonNamer.WithPrefix(comp.Name())
 		comp.host = host
@@ -48,12 +48,8 @@ func NewHostAgent(e *config.CommonEnvironment, host *remoteComp.Host, dep pulumi
 			return err
 		}
 
-		opts := []pulumi.ResourceOption{pulumi.Parent(comp)}
-		if dep != nil {
-			opts = append(opts, utils.PulumiDependsOn(dep))
-		}
-
-		err = comp.installAgent(e, params, opts...)
+		deps := append(params.PulumiDependsOn, pulumi.Parent(comp))
+		err = comp.installAgent(e, params, deps...)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


brycekahle (approved):

components/datadog/agent/host.go
```diff
@@ -77,6 +82,55 @@ func (h *HostAgent) installAgent(env *config.CommonEnvironment, params *agentpar
...
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	_, err = h.host.OS.Runner().Command(
+		h.namer.ResourceName("var/run/datadog perm"),
+		&command.Args{
+			Create: pulumi.String("chown dd-agent:dd-agent /var/run/datadog"),
```
vboulineau, line 99: You can do everything in a single command post-install

components/datadog/agent/host.go
```diff
@@ -77,6 +82,55 @@ func (h *HostAgent) installAgent(env *config.CommonEnvironment, params *agentpar
...
+			Sudo:   true,
+		},
+		utils.PulumiDependsOn(varddcmd),
+		utils.PulumiDependsOn(installCmd),
+	)
+	if err != nil {
+		return err
+	}
+
+	dockerGroup, err := h.host.OS.Runner().Command(
```
vboulineau, line 109: Those two commands should not be here as it's not related to Agent host install.

components/datadog/agent/host.go
```diff
@@ -37,7 +37,7 @@ func (h *HostAgent) Export(ctx *pulumi.Context, out *HostAgentOutput) error {
 }
 
 // NewHostAgent creates a new instance of a on-host Agent
-func NewHostAgent(e *config.CommonEnvironment, host *remoteComp.Host, options ...agentparams.Option) (*HostAgent, error) {
+func NewHostAgent(e *config.CommonEnvironment, host *remoteComp.Host, dep pulumi.Resource, options ...agentparams.Option) (*HostAgent, error) {
```
vboulineau, line 40: You should not add parameter here but add a `WithPulumiResourceOptions` in `agentparams` (similar to https://github.com/DataDog/test-infra-definitions/blob/main/components/datadog/dockeragentparams/params.go#L81-L86 - but with proper name).

components/datadog/agentparams/params.go
```diff
@@ -187,6 +188,13 @@ func WithTelemetry() func(*Params) error {
 	}
 }
 
+func WithPulumiDependsOn(resources ...pulumi.ResourceOption) func(*Params) error {
```
vboulineau, line 191:
```suggestion
func WithPulumiResourceOptions(resources ...pulumi.ResourceOption) func(*Params) error {
```

knusbaum: Fixed. Renamed the config member as well for consistency.

vboulineau (approved):

PR #598, opened 2024-02-01
Use -Force when creating Windows folder to avoid ResourceExists error

Kaderinho:
What does this PR do?
---------------------

Use `-Force` flag when creating a folder on Windows

-------------------

We have issues when we use `WithIntegration` for the same integration in different tests on Windows. For instance, a first test creates the folder, and when the second test creates the folder again it fails with a `ResourceExists` error.

Motivation
----------

Additional Notes
----------------


pducolin (approved):

pducolin:
Thank you !

PR #604, opened 2024-02-07
Overwrite API key if it already exists

KevinFairise2:
What does this PR do?
---------------------

We expect the API key to be created by Pulumi. In some cases if the deletion failed we will get an error saying the parameter already exists. We prevent this error from happening by allowing pulumi to overwrite the parameter.

-------------------

Motivation
----------

Additional Notes
----------------


pducolin (approved):

scenarios/aws/ecs/run.go
```diff
@@ -102,9 +102,10 @@ func Run(ctx *pulumi.Context) error {
 			}
 		}
 		apiKeyParam, err = ssm.NewParameter(ctx, awsEnv.Namer.ResourceName("agent-apikey"), &ssm.ParameterArgs{
```
nplanel, line 104: nit:  namer.ResourceName("agent", "apikey")

PR #611, opened 2024-02-08
[os] descriptor from string defaults to default descriptor

pducolin:
What does this PR do?
---------------------

Default descriptor to a valid descriptor instead of to an os flavor

-------------------

docker and vm

Motivation
----------

The docker run function is currently broken, as it panics if a valid os descriptor is not defined on the config map

Additional Notes
----------------


adel121 (approved):

L3n41c (approved):

vboulineau:
@pducolin This PR has broken usage of `inv` docker command. If you input `::arm64`, it will always use Ubuntu regardless of the default flavor set in `scenarios/aws/ec2/vm_run.go`

PR #614, opened 2024-02-12
[host] increase agent installer curl retry

pducolin:
What does this PR do?
---------------------

Increase `curl` retries to fetch the installer script from 3 to 10 

-------------------

Linux Host

Motivation
----------

We are seeing flakiness doe to an `SSL_ERROR_ZERO_RETURN` on curl on tests using this resource

Additional Notes
----------------


Pythyu (approved):
🚀

PR #619, opened 2024-02-14
Use separate network bridge for each VM set.

usamasaqib:
What does this PR do?
---------------------
This PR uses a separate network bridge for each VM set.
It also set the MTU size of the network bridge to 9000.
[This PR](https://github.com/DataDog/ami-builder/pull/157) in ami-builder sets the corresponding MTUs of the VM interfaces to match this.

-------------------
`aws/microvms` only

Motivation
----------
We noticed the network read path to the NFS server experiencing latency causing problem with tests in KMT.
One possible culprit in this path is the shared network bridge between all the VMs and the host. This PR tries to alleviate pressure on the bridge by a assigning each vmset a different one.

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/microVMs/microvms/libvirt.go
+++ b/scenarios/aws/microVMs/microvms/libvirt.go
@@ -81,7 +81,8 @@ type VMCollection struct {
 	vmsets   []vmconfig.VMSet
 	pools    map[vmconfig.PoolType]LibvirtPool
 	fs       map[vmconfig.VMSetID]*LibvirtFilesystem
-	domains  []*Domain
+	domains  map[vmconfig.VMSetID][]*Domain
+	subnets  map[vmconfig.VMSetID]string
 	LibvirtProvider
 }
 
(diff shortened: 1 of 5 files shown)


scenarios/aws/microVMs/microvms/libvirt.go
```diff
@@ -177,39 +178,59 @@ func (vm *VMCollection) SetupCollectionDomainConfigurations(depends []pulumi.Res
 			}
 		}
 
-		vm.domains = append(vm.domains, domains...)
+		if _, ok := vm.domains[set.ID]; ok {
+			vm.domains[set.ID] = append(vm.domains[set.ID], domains...)
+		} else {
+			vm.domains[set.ID] = domains
+		}
```
gjulianm, line 185: nit: I think this could be simplified

```suggestion
		vm.domains[set.ID] = append(vm.domains[set.ID], domains...)
```

scenarios/aws/microVMs/microvms/libvirt.go
```diff
@@ -177,39 +178,59 @@ func (vm *VMCollection) SetupCollectionDomainConfigurations(depends []pulumi.Res
...
-		dhcpEntries = append(dhcpEntries, d.dhcpEntry)
-
+	dhcpEntries := make(map[vmconfig.VMSetID][]interface{})
+	for setID, dls := range vm.domains {
+		for _, domain := range dls {
+			if _, ok := dhcpEntries[setID]; ok {
+				dhcpEntries[setID] = append(dhcpEntries[setID], domain.dhcpEntry)
+			} else {
+				dhcpEntries[setID] = []interface{}{domain.dhcpEntry}
+			}
```
gjulianm, line 199: Same as the other one:

```suggestion
			dhcpEntries[setID] = append(dhcpEntries[setID], domain.dhcpEntry)
```

gjulianm (approved):
just some minor comments

gjulianm (approved):

PR #620, opened 2024-02-14
[CONTINT-3783] Add tracegen app to eks and kind scenarios

AliDatadog:
What does this PR do?
---------------------
This PR adds a workload sending traces from both cgroupv1 and cgroupv2 nodes in the aws/eks scenario

-------------------
aws/eks, aws/kind

Motivation
----------
We would like to add e2e tests to make sure container id is retrieved for apm traces on both cgroupv1 and cgroupv2. 

Additional Notes
----------------
Example: https://dddev.datadoghq.com/apm/traces?query=%40_top_level%3A1%20kube_cluster_name%3Aali-benabdallah-aws-eks%20kube_namespace%3Aworkload-tracegen-cgroupv2%20kube_deployment%3Atracegen-udp&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=true&messageDisplay=inline&query_translation_version=v0&sort=desc&spanType=service-entry&traceQuery=&view=spans&start=1707755328234&end=1707928128234&paused=false

### DIFF
--- a/components/datadog/apps/tracegen/k8s.go
+++ b/components/datadog/apps/tracegen/k8s.go
@@ -17,21 +17,7 @@ type K8sComponent struct {
 	pulumi.ResourceState
 }
 
-func nodeSelector(nodeGroups pulumi.StringArrayInput) pulumi.StringMapInput {
-	return nodeGroups.ToStringArrayOutput().ApplyT(
-		func(arr []string) pulumi.StringMapInput {
-			for _, ng := range arr {
-				if ng != "" {
-					return pulumi.StringMap{
-						"eks.amazonaws.com/nodegroup": pulumi.String(ng),
-					}
-				}
-			}
-			return nil
-		}).(pulumi.StringMapInput)
-}
-
-func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, traceAgentURL string, nodeGroups pulumi.StringArrayInput, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
+func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, traceAgentURL string, nodeGroup pulumi.StringInput, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
 	opts = append(opts, pulumi.Provider(kubeProvider), pulumi.Parent(kubeProvider), pulumi.DeletedWith(kubeProvider))
 
 	k8sComponent := &K8sComponent{}
@@ -76,7 +62,9 @@ func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provi
 					},
 				},
 				Spec: &corev1.PodSpecArgs{
-					NodeSelector: nodeSel,
+					NodeSelector: pulumi.StringMap{
+						"eks.amazonaws.com/nodegroup": nodeGroup,
+					},
 					Containers: corev1.ContainerArray{
 						&corev1.ContainerArgs{
 							Name:  pulumi.String("tracegen"),
--- a/components/datadog/apps/tracegen/k8s.go
+++ b/components/datadog/apps/tracegen/k8s.go
@@ -1,0 +1,197 @@
+package tracegen
+
+import (
+	"fmt"
+
+	"github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+
+	"github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes"
+	appsv1 "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/apps/v1"
+	corev1 "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/core/v1"
+	metav1 "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/meta/v1"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type K8sComponent struct {
+	pulumi.ResourceState
+}
+
+func nodeGroupAffinity(nodeGroups pulumi.StringArray) corev1.AffinityPtrInput {
+	if len(nodeGroups) == 0 {
+		return nil
+	}
+	return &corev1.AffinityArgs{
+		NodeAffinity: &corev1.NodeAffinityArgs{
+			RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelectorArgs{
+				NodeSelectorTerms: corev1.NodeSelectorTermArray{
+					&corev1.NodeSelectorTermArgs{
+						MatchExpressions: corev1.NodeSelectorRequirementArray{
+							&corev1.NodeSelectorRequirementArgs{
+								Key:      pulumi.String("eks.amazonaws.com/nodegroup"),
+								Operator: pulumi.String("In"),
+								Values:   nodeGroups,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, traceAgentURL string, nodeGroups pulumi.StringArray, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
+	opts = append(opts, pulumi.Provider(kubeProvider), pulumi.Parent(kubeProvider), pulumi.DeletedWith(kubeProvider))
+
+	k8sComponent := &K8sComponent{}
+	if err := e.Ctx.RegisterComponentResource("dd:apps", fmt.Sprintf("%s/tracegen", namespace), k8sComponent, opts...); err != nil {
+		return nil, err
+	}
+
+	opts = append(opts, pulumi.Parent(k8sComponent))
+
+	ns, err := corev1.NewNamespace(e.Ctx, namespace, &corev1.NamespaceArgs{
+		Metadata: metav1.ObjectMetaArgs{
+			Name: pulumi.String(namespace),
+		},
+	}, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	affinity := nodeGroupAffinity(nodeGroups)
+
+	opts = append(opts, utils.PulumiDependsOn(ns))
+
+	if _, err := appsv1.NewDeployment(e.Ctx, fmt.Sprintf("%s/tracegen-uds", namespace), &appsv1.DeploymentArgs{
+		Metadata: &metav1.ObjectMetaArgs{
+			Name:      pulumi.String("tracegen-uds"),
+			Namespace: pulumi.String(namespace),
+			Labels: pulumi.StringMap{
+				"app": pulumi.String("tracegen-uds"),
+			},
+		},
+		Spec: &appsv1.DeploymentSpecArgs{
+			Replicas: pulumi.Int(1),
+			Selector: &metav1.LabelSelectorArgs{
+				MatchLabels: pulumi.StringMap{
+					"app": pulumi.String("tracegen-uds"),
+				},
+			},
+			Template: &corev1.PodTemplateSpecArgs{
+				Metadata: &metav1.ObjectMetaArgs{
+					Labels: pulumi.StringMap{
+						"app": pulumi.String("tracegen-uds"),
+					},
+				},
+				Spec: &corev1.PodSpecArgs{
+					Affinity: affinity,
+					Containers: corev1.ContainerArray{
+						&corev1.ContainerArgs{
+							Name:  pulumi.String("tracegen"),
+							Image: pulumi.String("ghcr.io/datadog/apps-tracegen:main"),
+							Env: &corev1.EnvVarArray{
+								&corev1.EnvVarArgs{
+									Name:  pulumi.String("DD_TRACE_AGENT_URL"),
+									Value: pulumi.String(traceAgentURL),
+								},
+							},
+							Resources: &corev1.ResourceRequirementsArgs{
+								Limits: pulumi.StringMap{
+									"cpu":    pulumi.String("100m"),
+									"memory": pulumi.String("32Mi"),
+								},
+								Requests: pulumi.StringMap{
+									"cpu":    pulumi.String("10m"),
+									"memory": pulumi.String("32Mi"),
+								},
+							},
+							VolumeMounts: &corev1.VolumeMountArray{
+								&corev1.VolumeMountArgs{
+									Name:      pulumi.String("apmsocketpath"),
+									MountPath: pulumi.String("/var/run/datadog"),
+								},
+							},
+						},
+					},
+					Volumes: corev1.VolumeArray{
+						&corev1.VolumeArgs{
+							Name: pulumi.String("apmsocketpath"),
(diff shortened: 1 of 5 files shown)


components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,197 @@
...
+						},
+					},
+				},
+			},
+		},
+	}, opts...); err != nil {
+		return nil, err
+	}
+
+	if _, err := appsv1.NewDeployment(e.Ctx, fmt.Sprintf("%s/tracegen-udp", namespace), &appsv1.DeploymentArgs{
```
ahmed-mez, line 132: So there is no traces on udp, it's either UDS or TCP.

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,197 @@
...
+									},
+								},
+								&corev1.EnvVarArgs{
+									Name: pulumi.String("DD_ENTITY_ID"),
+									ValueFrom: &corev1.EnvVarSourceArgs{
+										FieldRef: &corev1.ObjectFieldSelectorArgs{
+											FieldPath: pulumi.String("metadata.uid"),
+										},
+									},
+								},
```
ahmed-mez, line 175: ✂️ This is irrelevant for the tracer.

ahmed-mez (commented):
Thanks - just left some minor comments on tracegen

resources/aws/eks/nodeGroups.go
```diff
@@ -48,7 +48,7 @@ func newManagedNodeGroup(e aws.Environment, name string, cluster *eks.Cluster, n
 		NodeGroupNamePrefix: e.CommonNamer.DisplayName(37, pulumi.String(name), pulumi.String("ng")),
 		ScalingConfig: awsEks.NodeGroupScalingConfigArgs{
 			DesiredSize: pulumi.Int(1),
-			MaxSize:     pulumi.Int(1),
+			MaxSize:     pulumi.Int(2),
```
L3n41c, line 51: Did you fill the cluster up to the point where some pods couldn’t be scheduled ?
That’s surprising given the low resource requests set on the `tracegen` pods.

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,197 @@
...
+							Resources: &corev1.ResourceRequirementsArgs{
+								Limits: pulumi.StringMap{
+									"cpu":    pulumi.String("10m"),
+									"memory": pulumi.String("32Mi"),
+								},
+								Requests: pulumi.StringMap{
+									"cpu":    pulumi.String("2m"),
+									"memory": pulumi.String("32Mi"),
+								},
+							},
```
L3n41c, line 186: For the `tracegen-uds` workload, the CPU request and limit were set to 10 and 100 millicores.
I would have expected similar figures for both `tracegen` deployments.
Also, we can monitor a deployment to check how much it really needs.

scenarios/aws/eks/run.go
```diff
@@ -132,29 +135,59 @@ func Run(ctx *pulumi.Context) error {
...
 			nodeGroups = append(nodeGroups, ng)
+			// The default AMI used for Amazon Linux 2 is using cgroupv1
+			ng.NodeGroup.NodeGroupName().ApplyT(func(s pulumi.String) interface{} {
+				ngMutex.Lock()
+				defer ngMutex.Unlock()
+				cgroupV1Ng = append(cgroupV1Ng, s)
+				wg.Done()
+				return nil
+			})
 		}
```
L3n41c, line 158: Pulumi must be able to execute the full `Run()` function before starting the resources instantiation. We shouldn’t delay some resources declaration until some others are instantiated.
We should directly pass the `ng.NodeGroup.NodeGroupName()` to the `tracegen.K8sAppDefinition` function.

L3n41c (changes requested):

scenarios/aws/eks/run.go
AliDatadog: Fixed in the last commit

resources/aws/eks/nodeGroups.go
AliDatadog: Yes I did with the default `inv create-eks` command.

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,197 @@
...
+								Operator: pulumi.String("In"),
+								Values:   nodeGroups,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
```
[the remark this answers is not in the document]
AliDatadog: Is it possible to specify multiple possible nodegroups with the nodeSelector ? Or maybe we should use only one ?

components/datadog/apps/tracegen/k8s.go
AliDatadog: Removed !
Also I think it could be useful to keep it as the admission controller adds it and we should check that the expected behavior isn't affected.

components/datadog/apps/tracegen/k8s.go
AliDatadog: Alright thanks for the info, I just updated it

components/datadog/apps/tracegen/k8s.go
AliDatadog: I copy pasted what was set for dogstatsd. I suppose with the network stack we have more contention ?

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,181 @@
...
+					return pulumi.StringMap{
+						"eks.amazonaws.com/nodegroup": pulumi.String(ng),
+					}
+				}
+			}
+			return nil
+		}).(pulumi.StringMapInput)
+}
+
+func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, traceAgentURL string, nodeGroups pulumi.StringArrayInput, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
```
L3n41c, line 34:
```suggestion
func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, traceAgentURL string, nodeGroup pulumi.StringInput, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
```

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,181 @@
...
+				if ng != "" {
+					return pulumi.StringMap{
+						"eks.amazonaws.com/nodegroup": pulumi.String(ng),
+					}
+				}
+			}
+			return nil
+		}).(pulumi.StringMapInput)
+}
+
```
L3n41c, line 33:
```suggestion
```

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,165 @@
...
+				},
+			},
+			Template: &corev1.PodTemplateSpecArgs{
+				Metadata: &metav1.ObjectMetaArgs{
+					Labels: pulumi.StringMap{
+						"app": pulumi.String("tracegen-tcp"),
+					},
+				},
+				Spec: &corev1.PodSpecArgs{
+					Containers: corev1.ContainerArray{
```
L3n41c, line 131: Shouldn’t we also have a `NodeSelector` for the second deployment ?

```suggestion
				Spec: &corev1.PodSpecArgs{
					NodeSelector: pulumi.StringMap{
						"eks.amazonaws.com/nodegroup": nodeGroup,
					},
					Containers: corev1.ContainerArray{
```

resources/aws/eks/nodeGroups.go
L3n41c: I think that it’s because `inv create-eks` disables too many node groups (all but one).
I prefer to enable an other node group by default rather than increasing the size of this node group because it would better improve the test coverage.

components/datadog/apps/tracegen/k8s.go
AliDatadog: I lowered them to 2-10m as my tests showed between 1 and 2mcores of usage

resources/aws/eks/nodeGroups.go
AliDatadog: It makes sense, I propose to enable the bottlerocket one by default

components/datadog/apps/tracegen/k8s.go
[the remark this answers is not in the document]
L3n41c: For simplicity, we can arbitrarily chose one. Anyway, our deployment will have a single replica.

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,168 @@
...
+						"eks.amazonaws.com/nodegroup": nodeGroup,
+					},
+					Containers: corev1.ContainerArray{
+						&corev1.ContainerArgs{
+							Name:  pulumi.String("tracegen-uds"),
+							Image: pulumi.String("ghcr.io/datadog/apps-tracegen:main"),
+							Env: &corev1.EnvVarArray{
+								&corev1.EnvVarArgs{
+									Name:  pulumi.String("DD_TRACE_AGENT_URL"),
+									Value: pulumi.String(traceAgentURL),
```
L3n41c, line 73: I would be fine with a hard-code of the value here.

```suggestion
									Value: pulumi.String("/var/run/datadog/apm.socket"),
```

Having a `traceAgentURL` parameter in the `K8sAppDefinition` function gives the feeling that we can change the path of the `trace-agent` socket.
But in practice, it’s false. If we pass `/run/apm.socket` for ex., it won’t work because the `apmsocketpath` `HostPath` is hardcoded to `/var/run/datadog` line 98.

So, we would need to make `HostPath` line 98 use the dirname of `traceAgentURL`.
This sounds overkill to me.
That’s why I’d rather hardcode the socket path here and remove the `traceAgentURL` parameter from the `K8sAppDefinition`.

Also, it looked weird to be able to customize the UDS path in UDS case, but not the TCP port in the TCP case.

AliDatadog: I hardcode the path in the last commit 👍

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,168 @@
...
+			Template: &corev1.PodTemplateSpecArgs{
+				Metadata: &metav1.ObjectMetaArgs{
+					Labels: pulumi.StringMap{
+						"app": pulumi.String("tracegen-uds"),
+					},
+				},
+				Spec: &corev1.PodSpecArgs{
+					NodeSelector: pulumi.StringMap{
+						"eks.amazonaws.com/nodegroup": nodeGroup,
+					},
```
L3n41c, line 65: This looks like an EKS specificity inside a file that should be generic for all Kubernetes distributions.
When we’ll support AKS or GKE for ex., we’ll need to use `cloud.google.com/gke-nodepool` instead of `eks.amazonaws.com/nodegroup`.
So, maybe it makes more sense to have not only the node selector value, but also the key passed as parameter of this function.
WDYT ?

```suggestion
					NodeSelector: nodeSelector,
```

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,168 @@
...
+	corev1 "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/core/v1"
+	metav1 "github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/meta/v1"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type K8sComponent struct {
+	pulumi.ResourceState
+}
+
+func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, nodeGroup pulumi.StringInput, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
```
L3n41c, line 20:
```suggestion
func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, nodeSelector pulumi.StringMapInput, opts ...pulumi.ResourceOption) (*K8sComponent, error) {
```

components/datadog/apps/tracegen/k8s.go
```diff
@@ -0,0 +1,168 @@
...
+			Template: &corev1.PodTemplateSpecArgs{
+				Metadata: &metav1.ObjectMetaArgs{
+					Labels: pulumi.StringMap{
+						"app": pulumi.String("tracegen-tcp"),
+					},
+				},
+				Spec: &corev1.PodSpecArgs{
+					NodeSelector: pulumi.StringMap{
+						"eks.amazonaws.com/nodegroup": nodeGroup,
+					},
```
L3n41c, line 133:
```suggestion
					NodeSelector: nodeSelector,
```

scenarios/aws/eks/run.go
```diff
@@ -266,6 +274,14 @@ func Run(ctx *pulumi.Context) error {
 				return err
 			}
 
+			if _, err := tracegen.K8sAppDefinition(*awsEnv.CommonEnvironment, eksKubeProvider, "workload-tracegen-cgroupv1", cgroupV1Ng); err != nil {
```
L3n41c, line 277:
```suggestion
			if _, err := tracegen.K8sAppDefinition(*awsEnv.CommonEnvironment, eksKubeProvider, "workload-tracegen-cgroupv1", pulumi.StringMap{"eks.amazonaws.com/nodegroup": cgroupV1Ng}); err != nil {
```

scenarios/aws/eks/run.go
```diff
@@ -266,6 +274,14 @@ func Run(ctx *pulumi.Context) error {
...
+				return err
+			}
+
+			if _, err := tracegen.K8sAppDefinition(*awsEnv.CommonEnvironment, eksKubeProvider, "workload-tracegen-cgroupv2", cgroupV2Ng); err != nil {
```
L3n41c, line 281:
```suggestion
			if _, err := tracegen.K8sAppDefinition(*awsEnv.CommonEnvironment, eksKubeProvider, "workload-tracegen-cgroupv2", pulumi.StringMap{"eks.amazonaws.com/nodegroup": cgroupV2Ng}); err != nil {
```

scenarios/aws/eks/run.go
L3n41c, line 1: Even if on Kind we won’t have both cgroupv1 and cgroupv2 cohabiting in the same cluster, I think it would make sense to deploy the `tracegen` app also on the `kind` scenario so that we can test the `trace-agent` on kind (with a single cgroup version).

L3n41c (commented):
The current `tracegen.K8sAppDefinition(…)` function currently contains a `nodeSelector` definition that makes it EKS-specific.
I think it would be better to move this EKS-specific part to an EKS-specific file and to deploy the `tracegen` app also on Kind.
Even if we won’t be able to validate both versions of cgroup there, there is a value in having a test for the `trace-agent`.

components/datadog/apps/tracegen/k8s.go
AliDatadog: Indeed it makes sense ! Will correct it now

scenarios/aws/eks/run.go
AliDatadog: Indeed I wanted to leave it for another PR but I just added it now

L3n41c (approved):
👍🏻

PR #622, opened 2024-02-15
[ECS] enable NPM in the agent

nplanel:
What does this PR do?
---------------------

Enable NPM in the agent for ECS deployment

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/ecs.go
+++ b/components/datadog/agent/ecs.go
@@ -116,6 +116,10 @@ func ecsLinuxAgentSingleContainerDefinition(e config.CommonEnvironment, apiKeySS
 				Name:  pulumi.StringPtr("DD_PROCESS_AGENT_ENABLED"),
 				Value: pulumi.StringPtr("true"),
 			},
+			ecs.TaskDefinitionKeyValuePairArgs{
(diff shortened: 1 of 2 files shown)


components/datadog/agent/ecs.go
```diff
@@ -116,6 +116,10 @@ func ecsLinuxAgentSingleContainerDefinition(e config.CommonEnvironment, apiKeySS
 				Name:  pulumi.StringPtr("DD_PROCESS_AGENT_ENABLED"),
 				Value: pulumi.StringPtr("true"),
 			},
+			ecs.TaskDefinitionKeyValuePairArgs{
+				Name:  pulumi.StringPtr("DD_SYSTEM_PROBE_NETWORK_ENABLED"),
```
pducolin, line 120: **❓ question**
Should this be an option ? Not sure if we want to enable it by default

nplanel: At the moment we have no option passing through for ECS, but I agreed would be better to have the same options than others environments.

nplanel: I added ecsagentparams setting (for example) environment variables to the agent

pducolin (approved):
Thank you !

PR #629, opened 2024-02-15
Add checks for encrypted keys and pulumi creds to setup.debug

clarkb7:
What does this PR do?
---------------------
Add more checks to `inv setup.debug`
- Ensures pulumi is logged in
- if `privateKeyPath` is configured, ensures it is not encrypted or a proper `privateKeyPassword` is provided

Fail `inv.setup` if `pulumi` cannot be executed, e.g. not on PATH

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/config.py
+++ b/tasks/config.py
@@ -16,6 +16,7 @@ class Config(BaseModel, extra=Extra.forbid):
             keyPairName: Optional[str]
             publicKeyPath: Optional[str]
             privateKeyPath: Optional[str] = None
+            privateKeyPassword: Optional[str] = None
             account: Optional[str]
             teamTag: Optional[str]
 
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -30,31 +30,7 @@ def setup(
     if pulumi_up_to_date:
         info(f"Pulumi is up to date: {pulumi_version}")
     else:
-        info("🤖 Install Pulumi")
-        if is_windows():
-            ctx.run("winget install pulumi")
-        elif is_linux():
-            ctx.run("curl -fsSL https://get.pulumi.com | sh")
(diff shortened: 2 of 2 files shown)


tasks/setup.py
```diff
@@ -31,16 +32,34 @@ def setup(
...
+                    str(x)
+                    for x in [
+                        Path().home().joinpath(".pulumi", "bin"),
+                        Path().home().joinpath("AppData", "Local", "pulumi", "bin"),
+                        'C:\\Program Files (x86)\\Pulumi\\bin',
+                    ]
+                ]
+                os.environ["PATH"] = ';'.join([os.environ["PATH"]] + paths)
+            else:
+                path = Path().home().joinpath(".pulumi", "bin")
```
vboulineau, line 56: Why would that be necessary?

tasks/config.py
```diff
@@ -16,6 +16,7 @@ class Aws(BaseModel, extra=Extra.forbid):
             keyPairName: Optional[str]
             publicKeyPath: Optional[str]
             privateKeyPath: Optional[str] = None
+            privateKeyPassword: Optional[str] = None
```
vboulineau, line 19: If we store the private key password in this file, we should explicitly generate a warning in `setup.py` to use a key dedicated to test accounts.

tasks/setup.py
clarkb7: If the above lines install pulumi for he first time, then it's not on the path for the process because it was only just installed. This means that the `pulumi plugin install` and `pulumi login --local` commands fail. This was hidden before because `os.system` continues on error.

vboulineau (approved):

PR #632, opened 2024-02-16
Create interface for parameter cross cloud providers

KevinFairise2:
What does this PR do?
---------------------
Provide way to have parameter that exists on all provider with different values for each provider.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/resources/aws/environment.go
+++ b/resources/aws/environment.go
@@ -66,6 +66,8 @@ func (e *Environment) GetInternalRegistry() string {
 	return "669783387624.dkr.ecr.us-east-1.amazonaws.com"
 }
 
+var _ config.CloudProviderEnvironment = (*Environment)(nil)
+
 func WithCommonEnvironment(e *config.CommonEnvironment) func(*Environment) {
 	return func(awsEnv *Environment) {
 		awsEnv.CommonEnvironment = e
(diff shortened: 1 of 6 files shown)


resources/aws/environment.go
```diff
@@ -62,9 +62,16 @@ type Environment struct {
 	randomSubnets pulumi.StringArrayOutput
 }
 
+type CloudProviderEnvironment struct{}
+
+func (e *CloudProviderEnvironment) GetInternalRegistry() string {
+	return "669783387624.dkr.ecr.us-east-1.amazonaws.com"
+}
+
```
pducolin, line 70: **💬 suggestion**
I'd rather have `aws.Environment` implementing the interface

```suggestion
var _ CloudProviderEnvironment = &Environment{}

func (e *Environment) GetInternalRegistry() string {
	return "669783387624.dkr.ecr.us-east-1.amazonaws.com"
}
```

KevinFairise2: I am not sure understand how I should use it if I do this. Because I wanted to have 
an instance of `CloudProviderEnvironment` in `CommonConfig`.
Then I would have `aws.Environment` has an attribute for `CommonConfig` and `CommonConfig` that has an attribute pointing to `aws.Environment

KevinFairise2: Modified the code!

usamasaqib (approved):

resources/aws/environment.go
```diff
@@ -62,9 +62,16 @@ type Environment struct {
 	randomSubnets pulumi.StringArrayOutput
 }
 
+func (e *Environment) GetInternalRegistry() string {
```
vboulineau, line 65: The function should be below with other parameters and it should use a parameter set in environment defaults as it's not the same between accounts.

KevinFairise2: For now we always use the QA env registry even for tests run on other account. But it is likely that we will have parameters depending on the account. Updated the PR to pave the way for this use case

vboulineau (approved):

PR #634, opened 2024-02-19
Add pre-commit hooks corresponding to GitHub workflows

gjulianm:
What does this PR do?
---------------------

Adds pre-commit config matching the linters found in workflows

-------------------

None, only tooling

Motivation
----------

Detect linter failures before without requiring CI runs.

Additional Notes
----------------

Due to how pre-commit works, we cannot install the modules from `requirements.txt` in the virtualenv, which means that `pyright` does not detect them and therefore fails (see https://github.com/pre-commit/pre-commit/issues/730 for a discussion on this problem). The solution is to revert to using the pyright program installed in the system, without the internal virtualenv that pre-commit creates.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,6 @@
 
 [tool.black]
 line-length = 120
-py36 = false
 skip-string-normalization = true
 
 [tool.isort]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,16 +3,22 @@ repos:
     rev: v1.52.2
     hooks:
     - id: golangci-lint
-  - repo: https://github.com/RobertCraigie/pyright-python
-    rev: v1.1.350
+  - repo: local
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -5,7 +5,6 @@
 
 [tool.black]
 line-length = 120
-py36 = false
```
pducolin, line 8: **❓ question**
Mostly for internal doc, why was this option removed ?

pducolin (approved):
Thank you so much !

pyproject.toml
gjulianm: Running black with this option prints a warning saying this option is unused. It seems it was deprecated in 2019 (https://github.com/psf/black/pull/724). From what I've been reading, black does automatic detection based on language features if the minimum Python version is not specified, so there's no need to disable explicitly python3.6.

PR #647, opened 2024-02-26
Create local kind cluster install

KevinFairise2:
What does this PR do?
---------------------

This PR add two new components:
- A component to install a local kind cluster, assuming that `Kind` is installed locally
- A component to deploy a Fakeintake in a local kind cluster, The port mapping is made to have the fakeintake accessible from the port `30080` of the local host

The following PR will create a local kind provider on datadog-agent: https://github.com/DataDog/datadog-agent/pull/23153

The PR also creates another API to be able to set the environment variables of all the Agents deployed with helm. Otherwise this is impossible since the env variables used to configure the fakeintake override the ones provided with `WithHelmValues`

-------------------

Motivation
----------

The main goal of this PR is to provide things for developers to iterate faster on their e2e tests.
Running Kubernetes tests on a remote host can be long and hard to debug, even when using `kind`

Additional Notes
----------------


components/datadog/fakeintake/component.go
```diff
@@ -9,7 +9,8 @@ import (
 type FakeintakeOutput struct { // nolint:revive, We want to keep the name as <Component>Output
 	components.JSONImporter
 
-	URL string `json:"url"`
+	ClientURL string `json:"clientURL"`
+	URL       string `json:"url"`
```
KevinFairise2, line 13: Ideally when the URL used internally by Pulumi and the on used by the fakeintake client in the test is different we'd like to export only the URL used by the fakeintake client.
In this case we'd like Pulumi to use `fakeintake.default.svc.cluster.local` but we'd like `localhost:30080` to be exported as this the URL that must be used when initializing the fakeintake client.

Open to any suggestions

components/kubernetes/kind-cluster.yaml
```diff
@@ -2,10 +2,15 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
+  extraPortMappings:
+  - containerPort: 30080
+    hostPort: 30080
```
KevinFairise2, line 7: Maybe we could benefit from having it configurable. But for now it uses a port that should not be already bound on the host. And I could not see why we'd want to change it

components/datadog/fakeintake/component.go
pducolin: Interesting, what is the URL you use `fakeintake.default.svc.cluster.local` to configure the agent and `ClientURL` for the fakeintake client? Let's discuss about this

vboulineau (commented):
Code overall LGTM, small comments.

However sharing concern about the use case. This helper assumes host machine is ~Unix + `docker (or podman)` + `kind` installed.

For manual tests, I am not sure doing all this to wrap `kind create cluster` is worth it, however it could argued that re-using manifests defined here to install some apps can be useful. In that case I'd then favor a solution that allows to re-use current in-context cluster instead (gives a lot of flexibility in the manual testing).

For E2E tests, it requires a slightly different provisioning code (hopefully environment and components are designed to have minimal changes) and it breaks the original promise: whether you run it from your laptop or from the CI, the result will be _exactly_ the same.

I also have doubt on the iterate faster. Iterating on Kubernetes E2E usually mean provisioning once with then a lot of updates (but you're not going to update the cluster), the added provisioning time is quite short.

Once provisioned, the difference to the `kubectl` user is expected to be ~0, so it'd be interesting to have a better understanding of the issues faced.

components/datadog/fakeintake/component.go
KevinFairise2: I will check if I can have something working using `<host_ip>:<node_port>` so I would not need to have 2 different URL.
Otherwise I will go with the `InternalURL`
Thanks!

components/datadog/agent/kubernetes_helm.go
```diff
@@ -393,7 +407,7 @@ func (values HelmValues) configureFakeintake(e config.CommonEnvironment, fakeint
 		},
 		pulumi.Map{
 			"name":  pulumi.String("DD_LOGS_CONFIG_ADDITIONAL_ENDPOINTS"),
-			"value": pulumi.Sprintf(`[{"host": "%s"}]`, fakeintake.Host),
+			"value": pulumi.Sprintf(`[{"host": "%s", "port": %v, "use_ssl": "%t"}]`, fakeintake.Host, fakeintake.Port, fakeintake.Scheme == "https"),
```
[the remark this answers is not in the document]
KevinFairise2: I think the port SHOULD always be set

KevinFairise2:
> For manual tests, I am not sure doing all this to wrap `kind create cluster` is worth it, however it could argued that re-using manifests defined here to install some apps can be useful. In that case I'd then favor a solution that allows to re-use current in-context cluster instead (gives a lot of flexibility in the manual testing).

Manual testing is not the first goal of this PR, the idea was more to create a provider that provision local resources that can be used when running e2e tests locally.

>it breaks the original promise: whether you run it from your laptop or from the CI, the result will be exactly the same.
Yes indeed the environment is not exactly the same as in the CI, that's one of the drawback but I tried to make the local environment as close as the remote one as possible.

>I also have doubt on the iterate faster. Iterating on Kubernetes E2E usually mean provisioning once with then a lot of updates (but you're not going to update the cluster), the added provisioning time is quite short.
It's a feedback we often have from users creating tests. Remote resources can be long to create, it often leaks some resources and running the test again can be painful (need to delete resources on AWS) and DevMode is not really helpful because the infra is not cleaned up properly between test runs

components/datadog/agent/kubernetes_helm.go
```diff
@@ -361,6 +364,17 @@ func (values HelmValues) configureImagePullSecret(secret *corev1.Secret) {
 	}
 }
 
+func (values HelmValues) configureEnvVars(envVars map[string]string) {
```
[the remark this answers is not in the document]
KevinFairise2: If you are talking about providing the `envDict` through the `ValuesYAML` parameter, I tried but this is not working. Because if you use a fakeintake, [here](https://github.com/DataDog/test-infra-definitions/blob/main/components/datadog/agent/kubernetes_helm.go#L132) we modify `values` with the environment variables needed to configure the fakeintake. When we pass `ValuesYAML` and `values` to `helm.NewInstallation` the env variables you define in `ValuesYAML` are not merged with the one in `values`

components/datadog/fakeintake/kubernetes.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const localPort = 30080
+
+func NewLocalKubernetesFakeintake(e config.CommonEnvironment, resourceName string, kubeProvider *kubernetes.Provider) (*Fakeintake, error) {
+	return components.NewComponent(e, resourceName, func(comp *Fakeintake) error {
+
+		_, err := v1.NewDeployment(e.Ctx, resourceName, &v1.DeploymentArgs{
+			Metadata: &metav1.ObjectMetaArgs{
+				Name: pulumi.String(resourceName),
```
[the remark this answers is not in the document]
KevinFairise2: Will do!

components/kubernetes/kind.go
```diff
@@ -108,3 +110,70 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
...
+			return err
+		}
+
+		kindVersionConfig, err := getKindVersionConfig(kubeVersion)
+		if err != nil {
+			return err
+		}
+
+		runner := command.NewLocalRunner(env, command.LocalRunnerArgs{
+			User:      user.Username,
```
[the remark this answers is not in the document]
KevinFairise2: I cannot remember why I added this but yes looks like you're right

components/datadog/agent/kubernetes_helm.go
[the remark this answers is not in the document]
vboulineau: I meant using using the `envDict` variable in the Helm chart: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/values.yaml#L531-L533.

The merge issues arises because `lists` merge is override, it's the behaviour in Helm (which is not used in Pulumi but replicated [here](https://github.com/pulumi/pulumi-kubernetes/blob/78ab38748b9101d98890d86af44f5addb6c852dd/provider/pkg/provider/helm_release.go#L1323))

Using `envDict` allow to define env vars in a map, which support merging, with `Values` overriding `ValuesYAML` if duplicate keys. If it's not working, there's something to be investigated.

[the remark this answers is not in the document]
KevinFairise2: Oh yes, I am not very familiar with the Agent helm chart but that looks like something I could use! I will try to confirm it works!

components/kubernetes/kind.go
```diff
@@ -108,3 +110,70 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
...
+		clusterConfig, err := runner.Command("kind-config", &command.Args{
+			Create: pulumi.Sprintf("cat - | tee %s > /dev/null", clusterConfigFilePath),
+			Delete: pulumi.Sprintf("rm -f %s", clusterConfigFilePath),
+			Stdin:  pulumi.String(kindClusterConfig),
+		}, opts...)
+		if err != nil {
+			return err
+		}
+
+		nodeImage := fmt.Sprintf("%s:%s", kindNodeImageRegistry, kindVersionConfig.nodeImageVersion)
```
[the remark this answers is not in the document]
KevinFairise2: Actually, I think this is not that easy to create an interface shared by the two runners because the `Command` method uses two different stuct `remote.Command` and `local.Command` which does not share an interface. I think we could achieve this by wrapping `local.Command` and `remote.Command` in a struct that implements a `Command` interface but that looks like a big refactor

components/datadog/fakeintake/component.go
KevinFairise2: Updated the code to use the host ip to communicate with the fakeintake. It should ease things if we want to have other local providers. And we no longer need two different URL

KevinFairise2:
Updated the PR to take into account the suggestions. I also no longer deploy the fakeintake in Kubernetes, but rather deploy it locally in Docker, it should make things easier to use it with other local providers

components/datadog/agent/kubernetes_helm.go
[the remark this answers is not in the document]
KevinFairise2: Removed these as it is not needed with `envDict`

components/datadog/fakeintake/docker.go
```diff
@@ -0,0 +1,52 @@
...
+					Internal: pulumi.Int(80),
+					External: pulumi.Int(hostPort),
+				},
+			},
+		}, e.WithProviders(config.ProviderDocker))
+		if err != nil {
+			return err
+		}
+
+		localIP := getLocalIP()
```
pducolin, line 31: **❓ question**
Ain't docker containers always listening on `localhost:<hostPort>`? You might skip this part

components/datadog/fakeintake/docker.go
```diff
@@ -0,0 +1,52 @@
...
+		comp.Port = hostPort
+		comp.Scheme = "http"
+		comp.URL = pulumi.Sprintf("%s://%s:%d", comp.Scheme, comp.Host, comp.Port)
+
+		return nil
+	})
+}
+
+func getLocalIP() net.IP {
+	conn, err := net.Dial("udp", "255.255.255.255:80") // initiate a connection to get the local address, the url does not need to exist
```
pducolin, line 43: **❓ question**
You should dial up on `hostPort` rather than on port `80`. If I understand it correctly, you are mapping the fakeintake container port 80 to the host port `30080`

Then we use it to connect to the fakeintake component URL from the test runner, that is the host where the fakeintake container instance is running.

I think we ca safely return `127.0.0.1:30080`

components/datadog/fakeintake/docker.go
KevinFairise2: Yes, if you want to reach the fakeintake from your computer you can do that. But I want a single URL that can be used from the your computer and from a `kind` (or `docker` image) running on your computer. Using `localhost:<hostPort>` in the `kind` cluster will not work. But using the host local IP does work.

components/datadog/fakeintake/docker.go
KevinFairise2: Actually that's not needed. This one is just a trick to get your local private IP. This is explained probably better [here](https://gosamples.dev/local-ip-address/)
But the idea is to initiate a connection (we do not even care that the connection can actually be created). Once the connection is initiated you can retrieve the local IP from the connection object.

KevinFairise2: > I think we ca safely return 127.0.0.1:30080
I would not be able to access the fake intake from within the `kind` cluster with that IP

pducolin:
```suggestion
	// Open a connection to an external valid URL to  
	// get the local address from the connection instance
	// The URL does not need to exist
	conn, err := net.Dial("udp", "255.255.255.255:80") 
```

pducolin: What is confusing is that `255.255. 255.255` is the broadcast address: `Dial` is sending a broadcast UDP request to all the network. In our private subnets there is probably no service on port `80`. This message is still reaching n hosts, depending on the subnet mask

I'd rather use a fake URL or `8.8.8.8:80`

KevinFairise2: Wanted something that make it clear that we do not need the IP to be reachable. Using the broadcast address was maybe not a good idea. I will move forward with `8.8.8.8:80`.
Thanks!

components/datadog/fakeintake/docker.go
```diff
@@ -0,0 +1,55 @@
...
+
+		return nil
+	})
+}
+
+func getLocalIP() net.IP {
+	// Open a connection to an external valid URL to  
+	// get the local address from the connection instance
+	// The URL does not need to exist
+	conn, err := net.Dial("udp", "255.255.255.255:80")
```
KevinFairise2, line 46:
```suggestion
	conn, err := net.Dial("udp", "8.8.8.8:80") 
```

pducolin (approved):

components/datadog/fakeintake/docker.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+}
+
+func getLocalIP() net.IP {
+	// Open a connection to an external valid URL to
+	// get the local address from the connection instance
+	// The URL does not need to exist
+	conn, err := net.Dial("udp", "8.8.8.8:80")
+	if err != nil {
+		log.Fatal(err)
```
pducolin, line 48: **❓ question**
Should you rather return the error, instead of panicking?

PR #661, opened 2024-03-04
fix fakeintake url on docker scenario

AliDatadog:
What does this PR do?
---------------------
This PR modifies the fakeintake url that is used in the docker scenario. This is because the fakeintake now uses https.

-------------------
aws/dockervm

Motivation
----------
Fix e2e test.

Additional Notes
----------------

### DIFF
--- a/components/datadog/dockeragentparams/params.go
+++ b/components/datadog/dockeragentparams/params.go
@@ -130,10 +130,10 @@ func withIntakeHostname(hostname pulumi.StringInput, shouldSkipSSLCertificateVal
 	}).(pulumi.BoolOutput)
 	return func(p *Params) error {
 		envVars := pulumi.Map{
-			"DD_DD_URL":                                 pulumi.Sprintf("http://%s:80", hostname),
+			"DD_DD_URL":                                 pulumi.Sprintf("https://%s", hostname),
 			"DD_LOGS_CONFIG_DD_URL":                     pulumi.Sprintf("%s:80", hostname),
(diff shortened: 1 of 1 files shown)


components/datadog/dockeragentparams/params.go
```diff
@@ -130,10 +130,10 @@ func withIntakeHostname(hostname pulumi.StringInput, shouldSkipSSLCertificateVal
 	}).(pulumi.BoolOutput)
 	return func(p *Params) error {
 		envVars := pulumi.Map{
-			"DD_DD_URL":                                 pulumi.Sprintf("http://%s:80", hostname),
+			"DD_DD_URL":                                 pulumi.Sprintf("https://%s", hostname),
 			"DD_LOGS_CONFIG_DD_URL":                     pulumi.Sprintf("%s:80", hostname),
```
L3n41c, line 134: Should the log intake be also migrated to http**s** ?

```suggestion
			"DD_LOGS_CONFIG_DD_URL":                     pulumi.Sprintf("%s:443", hostname),
```

AliDatadog: Is it mandatory to specify the port ?

L3n41c (approved):
The change looks good to me as long as we’re sure that this `withIntakeHostname` function is only used in tests that are using the load-balancer in front of the fake intake.
Indeed, it’s the load-balancer that does the TLS termination.
As the LB certificate should be valid, dropping the `*_SSL_VALIDATION` parameters should be fine.

PR #669, opened 2024-03-06
add graceful shutdown with poison pill to tracegen

knusbaum:
What does this PR do?
---------------------

This PR makes the tracegen send one final trace on shutdown, which can signal to consumers in the pipeline that no more traces for this run are coming.

-------------------

This will affect the APM scenarios

Motivation
----------

Sometimes tests get data from previous tests, when the data hasn't made its way fully from the tracegen, through the agent, to the intake before the next test starts.

Additional Notes
----------------

### DIFF
--- a/components/datadog/apps/tracegen/images/tracegen/main.go
+++ b/components/datadog/apps/tracegen/images/tracegen/main.go
@@ -26,8 +26,7 @@ func reportStats(done chan struct{}) {
 		select {
 		case <-done:
 			return
-		default:
-			time.Sleep(5 * time.Second)
+		case <-time.After(5 * time.Second):
 			tc := tracecount.Swap(0)
 			sc := spancount.Swap(0)
 			fmt.Printf("Finished %d traces/s, %d spans/second.\n", tc/5, sc/5)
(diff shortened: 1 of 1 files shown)


components/datadog/apps/tracegen/images/tracegen/main.go
```diff
@@ -99,6 +98,7 @@ func main() {
 		<-sigs
 		close(done)
 		fmt.Println("Exiting tracegen.")
+		time.Sleep(10 * time.Second) // alow time for other goroutines to shut down.
```
ahmed-mez, line 101: How about adding a sync group or is it overkill? 😅

ahmed-mez (approved):

components/datadog/apps/tracegen/images/tracegen/main.go
knusbaum: Yeah, could do, but I *do* want some sort of timeout so it doesn't indefinitely run.
Not sure the sync will help here much, since the program will shut down as soon as the main function exits, so this sleep won't actually delay the shutdown, just act as a safety to kill the process if anything deadlocks.

LMK.

PR #677, opened 2024-03-08
[apps] use common signature for k8s apps

pducolin:
What does this PR do?
---------------------

* Return `pulumi.ComponentResource` from k8s deployment app definitions

-------------------

eks, kind

Motivation
----------

We are adding custom workloads to Kind provider in datadog-agent side, we need a common type for the deployment functions to allow adding custom apps to the deployment through a `WithWorkloadApp` in https://github.com/DataDog/datadog-agent/pull/23632

Additional Notes
----------------

### DIFF
--- a/components/kubernetes/component.go
+++ b/components/kubernetes/component.go
@@ -28,7 +28,8 @@ func (c *Cluster) Export(ctx *pulumi.Context, out *ClusterOutput) error {
 	return components.Export(ctx, c, out)
 }
 
-type WorkloadComponent struct {
+// Workload is a Component that represents a Kubernetes workload
+type Workload struct {
 	pulumi.ResourceState
 	components.Component
 }
--- a/components/kubernetes/component.go
+++ b/components/kubernetes/component.go
@@ -27,3 +27,8 @@ type Cluster struct {
 func (c *Cluster) Export(ctx *pulumi.Context, out *ClusterOutput) error {
 	return components.Export(ctx, c, out)
 }
+
+type WorkloadComponent struct {
+	pulumi.ResourceState
+	components.Component
+}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/golangci/golangci-lint
-    rev: v1.52.2
+    rev: v1.55.2
     hooks:
     - id: golangci-lint
   - repo: local
--- a/components/datadog/apps/cpustress/k8s.go
+++ b/components/datadog/apps/cpustress/k8s.go
@@ -12,10 +12,10 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, opts ...pulumi.ResourceOption) (*componentskube.WorkloadComponent, error) {
+func K8sAppDefinition(e config.CommonEnvironment, kubeProvider *kubernetes.Provider, namespace string, opts ...pulumi.ResourceOption) (*componentskube.Workload, error) {
 	opts = append(opts, pulumi.Provider(kubeProvider), pulumi.Parent(kubeProvider), pulumi.DeletedWith(kubeProvider))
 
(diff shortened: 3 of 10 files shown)


components/datadog/apps/common/common.go
```diff
@@ -0,0 +1,23 @@
+package commonapp
+
+import (
+	"github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/pulumi/pulumi-kubernetes/sdk/v4/go/kubernetes"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+)
+
+type K8SDeployer interface {
+	DeployToKubeProvider(e config.CommonEnvironment, kubeProvider *kubernetes.Provider) (pulumi.ComponentResource, error)
```
KevinFairise2, line 10: I feel like the `namespace` and `pulumiOptions` could be in the signature of this function instead of being attributes of the `App` struct. Because we could have one single App that we want to deploy to different `namespace`.
This comment can be ignored, but I just found it more natural to select the namespace and the pulumiOptions when we deploy the App rather than when we create it

KevinFairise2 (approved):
One minor comment but otherwise I really like it 🎉

components/datadog/apps/common/common.go
pducolin: I initially had `namespace` in the `DeployToKubeProvider` and then moved it to the app instance. I agree the namespace is part of the deployment rather than the app definition, will change code on datadog-agent side to have an `WithAppWorkload` that takes an app and a namespace. It is harder to pass `pulumiOptions` as they depend on resources defined in the run function, I think in that case it is better to define a custom run function

vboulineau:
I think we can leverage the components more and avoid the interface + common implem + implem if we pass a functor `func (env, cluster) (kubernetes.Workload, error)` in `datadog-agent`.

Then component `kubernetes.Workload` would export nothing ATM (basically, just naming, would be an empty component). All `New` would take in `env, kubernetes.Cluster` (it should not take a provider anymore since we've introduce `Cluster` component).

pducolin:
@vboulineau I changed the commit history according to today's discussion, keeping the change that return a common component and dropping all the other changes, will adapt the PR on `datadog-agent side accordingly

pducolin:
> env, kubernetes.Cluster (it should not take a provider anymore since we've introduce Cluster component).

@vboulineau  this change would require further changes in the eks run function, that is not using the `kubernetes.Component`. I can work on it, but I would rather have a follow up to move from provider to component

vboulineau:
> > env, kubernetes.Cluster (it should not take a provider anymore since we've introduce Cluster component).
> 
> @vboulineau this change would require further changes in the eks run function, that is not using the `kubernetes.Component`. I can work on it, but I would rather have a follow up to move from provider to component

No issue on that. However you did not introduce a `Workload` component for the return type?

components/kubernetes/component.go
```diff
@@ -27,3 +27,8 @@ type Cluster struct {
 func (c *Cluster) Export(ctx *pulumi.Context, out *ClusterOutput) error {
 	return components.Export(ctx, c, out)
 }
+
+type WorkloadComponent struct {
```
vboulineau, line 31: nit: It's already a component `WorkloadComponent` -> `Workload`

vboulineau (approved):

components/kubernetes/component.go
pducolin: Added `Component` to make it more explicit from the caller that this is a component. We could move the file to a `components` folder and package and there it would be `componentsk8s.Workload`, but in `kubernetes` package it is not clear from the caller that this is a component, without looking into the struct implementation.

pducolin: I looked for guidance in https://go.dev/doc/effective_go, they use [both conventions ](https://go.dev/doc/effective_go#embedding)

From Go standard library I see that adding the name of the embedded struct is a practice adopted, for example in [bufio.Reader](https://cs.opensource.google/go/go/+/master:src/bufio/bufio.go;l=32-39)

`Component` is a relevant information for caller, it determines the nature of the struct: a `WorkloadComponent` is a component representing a workload. `Workload` would be too generic.

ahmed-mez (approved):

PR #680, opened 2024-03-11
update env vars

seoamy:
What does this PR do?
---------------------

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/docker.go
+++ b/components/datadog/agent/docker.go
@@ -86,8 +86,8 @@ func dockerAgentComposeManifest(agentImagePath string, apiKey pulumi.StringInput
 						"/sys/kernel/tracing:/sys/kernel/tracing",
 					},
 					Environment: map[string]any{
-						"DD_API_KEY":               apiKeyResolved,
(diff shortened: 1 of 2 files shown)


components/datadog/agent/docker.go
```diff
@@ -86,8 +86,8 @@ func dockerAgentComposeManifest(agentImagePath string, apiKey pulumi.StringInput
 						"/sys/kernel/tracing:/sys/kernel/tracing",
 					},
 					Environment: map[string]any{
-						"DD_API_KEY":               apiKeyResolved,
-						"DD_PROCESS_AGENT_ENABLED": true,
+						"DD_API_KEY": apiKeyResolved,
+						"DD_PROCESS_CONFIG_PROCESS_COLLECTION_ENABLED": true,
```
pducolin, line 90: **💭 thought**
Wonder if we should add a comment as per this variable is compatible with agent 7.35+. I don't expect folks to use the framework with such an old version though.

pducolin (approved):

PR #681, opened 2024-03-12
fix(install-script): Fix curl retry to reduce flakiness

KevinFairise2:
What does this PR do?
---------------------

Using `--retry` with curl only retries errors that `curl` consider as transient. Only timeouts and 5xx errors.
We want to retry more error than that to avoid flakes. For example the `SSL_ERROR_ZERO_RETURN` error

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/host_linuxos.go
+++ b/components/datadog/agent/host_linuxos.go
@@ -35,7 +35,7 @@ func (am *agentLinuxManager) getInstallCommand(version agentparams.PackageVersio
 		commandLine := strings.Join(testEnvVars, " ")
 
 		return fmt.Sprintf(
-			`curl --retry 10 -fsSL https://s3.amazonaws.com/dd-agent/scripts/%v  -o install-script.sh && for i in 1 2 3; do DD_API_KEY=%%s %v bash install-script.sh  && break || sleep 2; done`,
+			`for i in 1 2 3 4 5; do curl -L https://s3.amazonaws.com/dd-agent/scripts/%v -o install-script.sh && break || sleep $((2**$i)); done &&  for i in 1 2 3; do DD_API_KEY=%%s %v DD_INSTALL_ONLY=true bash install-script.sh  && break || sleep $((2**$i)); done`,
 			"install_script_agent7.sh",
 			commandLine), nil
 	}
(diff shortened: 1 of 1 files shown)


pgimalac (approved):
Not super readable but LGTM
Hopefully will fix our issue !

components/datadog/agent/host_linuxos.go
```diff
@@ -51,7 +51,7 @@ func (am *agentLinuxManager) getInstallCommand(version agentparams.PackageVersio
 	}
 
 	return fmt.Sprintf(
-		`curl --retry 10 -fsSL https://s3.amazonaws.com/dd-agent/scripts/%v -o install-script.sh && for i in 1 2 3; do DD_API_KEY=%%s %v DD_INSTALL_ONLY=true bash install-script.sh  && break || sleep 2; done`,
+		`for i in 1 2 3 4 5; do curl -L https://s3.amazonaws.com/dd-agent/scripts/%v -o install-script.sh && break || sleep $((2**$i)); done &&  for i in 1 2 3; do DD_API_KEY=%%s %v DD_INSTALL_ONLY=true bash install-script.sh  && break || sleep $((2**$i)); done`,
```
pducolin, line 54: **❓ question**
What is `**` doing at `sleep $((2**$i)`?

pducolin (approved):

components/datadog/agent/host_linuxos.go
```diff
@@ -35,7 +35,7 @@ func (am *agentLinuxManager) getInstallCommand(version agentparams.PackageVersio
 		commandLine := strings.Join(testEnvVars, " ")
 
 		return fmt.Sprintf(
-			`curl --retry 10 -fsSL https://s3.amazonaws.com/dd-agent/scripts/%v  -o install-script.sh && for i in 1 2 3; do DD_API_KEY=%%s %v bash install-script.sh  && break || sleep 2; done`,
+			`for i in 1 2 3 4 5; do curl -L https://s3.amazonaws.com/dd-agent/scripts/%v -o install-script.sh && break || sleep $((2**$i)); done &&  for i in 1 2 3; do DD_API_KEY=%%s %v DD_INSTALL_ONLY=true bash install-script.sh  && break || sleep $((2**$i)); done`,
```
pducolin, line 38: **❓ question**
Should we keep the fail, silent, show errors options?

pducolin:
Oh sorry I clicked here instead of on my PR 🤦‍♀️

components/datadog/agent/host_linuxos.go
KevinFairise2: It is an exponential sleep. We wait for 2**1 = 2 then 2**2 = 4 , 2**3 = 8 ....
`**` is just the `exponentiation` operator

PR #682, opened 2024-03-12
[kind] fix kind deployment

pducolin:
What does this PR do?
---------------------

* fix kind install command
* use AmazonECS family OS by default for Kind
* adds inv create-kind task
* adds integration test for kind

-------------------

kind

Motivation
----------

I broke kind install command with https://github.com/DataDog/test-infra-definitions/pull/671, this PR fixes it and adds an integration test for kind vm

Additional Notes
----------------

I tried to run tests in parallel with `t.Parallel` in subtests, but it failed, might race on config file. To be investigated.

### DIFF
--- a/components/kubernetes/kind.go
+++ b/components/kubernetes/kind.go
@@ -56,7 +56,7 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
 		kindInstall, err := runner.Command(
 			commonEnvironment.CommonNamer.ResourceName("kind-install"),
 			&command.Args{
-				Create: pulumi.Sprintf(`curl --retry 10 --fsSLo ./kind "https://kind.sigs.k8s.io/dl/%s/kind-linux-%s" && sudo install kind /usr/local/bin/kind`, kindVersionConfig.kindVersion, kindArch),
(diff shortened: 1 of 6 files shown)


components/kubernetes/kind.go
```diff
@@ -38,7 +38,12 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
 			return err
 		}
 
-		_, dockerInstallCmd, err := docker.NewManager(env, vm, true, opts...)
+		shouldInstallDocker := true
+		// docker is installed by default on Amazon Linux ECS
+		if vm.OS.Descriptor().Flavor == os.AmazonLinuxECS {
+			shouldInstallDocker = false
+		}
+		_, dockerInstallCmd, err := docker.NewManager(env, vm, shouldInstallDocker, opts...)
```
vboulineau, line 46: We need to add a `TODO: Lazy Docker install`

vboulineau (approved):

components/kubernetes/kind.go
pducolin: We have a card covering it

pducolin: https://datadoghq.atlassian.net/browse/ADXT-88, part of this sprint

pducolin:
I fixed the parallel running

pducolin:
Getting limited by memory, will increase limits

pducolin:
Giving up to parallel test execution, it takes over 16GiB to run all 3 tests, might come back later and run docker + vm in // and kind in series

PR #684, opened 2024-03-13
Improvement(e2e): Make Docker Manager lazily install docker

KevinFairise2:
What does this PR do?
---------------------

Change docker manager to install Docker only if it is not already installed.
To do this we had a new parameter to `Ensure` method implemented by Package managers:
- `check string` attribute: When this command is specified it will prefix the package manager install command, if the command succeed we skip the installation

-------------------

Docker

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/command/package.go
+++ b/components/command/package.go
@@ -37,7 +37,7 @@ func NewGenericPackageManager(
 	return packageManager
 }
 
-func (m *GenericPackageManager) Ensure(packageRef string, transform Transformer, opts ...pulumi.ResourceOption) (*remote.Command, error) {
+func (m *GenericPackageManager) Ensure(packageRef string, transform Transformer, check string, opts ...pulumi.ResourceOption) (*remote.Command, error) {
 	opts = append(opts, m.opts...)
 	if m.updateCmd != "" {
 		updateDB, err := m.updateDB(opts)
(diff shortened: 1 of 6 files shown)


components/command/package.go
```diff
@@ -47,8 +47,13 @@ func (m *GenericPackageManager) Ensure(packageRef string, transform Transformer,
 
 		opts = append(opts, utils.PulumiDependsOn(updateDB))
 	}
+	var cmdStr string
+	if checkCmd != "" {
+		cmdStr = fmt.Sprintf("bash -c '%s || %s %s'", checkCmd, m.installCmd, packageRef)
```
vboulineau, line 52: Commands are already run in a shell. With the current package managers we have, it should be enough.

KevinFairise2: I like `command` to check if the binary already exists, I will go for it.
However without changing how we format the command: [here](https://github.com/DataDog/test-infra-definitions/blob/main/components/command/unixOSCommand.go#L99) we need to wrap our command in `bash-c` otherwise the formatted command is:
`sudo command -v docker || apt-get install docker.io`
Which does not work because sudo applies only to `command -v`
I could modify this [line](https://github.com/DataDog/test-infra-definitions/blob/main/components/command/unixOSCommand.go#L99) to wrap everyhing in `bash -c` tho

vboulineau: Ah right the `sudo` wrapping. LGTM

vboulineau (approved):

PR #686, opened 2024-03-13
Add IIS component

julien-lebot:
What does this PR do?
---------------------
This PR adds a new IIS component  which allows one to deploy IIS on a given Windows VM.

-------------------
This will impact only Windows scenarios, it should not modify any existing scenarios but allow us to create tests that involve a VM with IIS installed.

Motivation
----------
1. System-probe tests need IIS installed.
2. The APM ETW provider in the Agent does not have any E2E tests and writing some would involve installing the .Net tracer + IIS.

Additional Notes
----------------
Using this component requires a custom environment / provisioner.

https://datadoghq.atlassian.net/browse/WINA-694


pducolin (approved):
**💬 suggestion**
What about moving this to `components/windows` and `call the file `iis.go`?

The component would be `windows.IIS` rather than `iis.Component`

components/iis/component.go
```diff
@@ -0,0 +1,148 @@
...
+//			Name:            "TestSite1",
+//			BindingPort:     "*:8081:",
+//			SourceAssetsDir: srcDir,
+//		}),
+//	)
+//	if err != nil {
+//		return err
+//	}
+//	err = iisServer.Export(ctx, env.IISServer)
+func NewServer(ctx *pulumi.Context, e *config.CommonEnvironment, host *remote.Host, options ...Option) (*Component, error) {
```
pducolin, line 53: **💬 suggestion**
```suggestion
func NewIISServer(ctx *pulumi.Context, e *config.CommonEnvironment, host *remote.Host, options ...Option) (*Component, error) {
```

components/iis/component.go
```diff
@@ -0,0 +1,148 @@
...
+	if host.OS.Descriptor().Family() != os.WindowsFamily {
+		// Print flavor in case OS family don't match, as it's more precise than the family (and the .String() conversion already exists).
+		return nil, fmt.Errorf("wrong Operating System family, expected Windows, got %s", host.OS.Descriptor().Flavor.String())
```
pducolin, line 56: **👏 praise**

components/iis/component.go
```diff
@@ -0,0 +1,148 @@
...
+
+				if site.SourceAssetsDir != "" {
+					copyAssets, err := host.OS.FileManager().CopyAbsoluteFolder(site.SourceAssetsDir, site.TargetAssetsDir, dependencies...)
+					if err != nil {
+						return err
+					}
+					dependencies = append(dependencies, pulumi.DependsOn(copyAssets))
+				}
+
+				_, err = host.OS.Runner().Command(comp.namer.ResourceName("create-iis-site", site.Name), &command.Args{
```
pducolin, line 128: **💬 suggestion**
Same here, you might want to ensure names are all differents or pulumi will fail with cryptic "not unique resource"

pducolin (approved):

components/iis/component.go
julien-lebot: Wouldn't that `stutter`? https://stackoverflow.com/a/45886055

components/iis/component.go
julien-lebot: Done ✔️  https://github.com/DataDog/test-infra-definitions/pull/686/commits/29d83398919041526bf96bb81fe33337ccd923f8

components/iis/component.go
julien-lebot: I've resumed the discussion points here: https://github.com/DataDog/test-infra-definitions/pull/686#issuecomment-1999678698

PR #687, opened 2024-03-13
Add Active-Directory component

julien-lebot:
What does this PR do?
---------------------
This PR adds (well, moves) the Active-Directory component so that it's possible to provision a domain environment.

-------------------
This will impact Windows scenarios that need to use a domain environment.

Motivation
----------
Currently the Active Directory code lives in the Datadog Agent repository and that makes it harder to be used to provision VMs and shared with other repositories.

Additional Notes
----------------
A Windows-specific provisioner will be created in the Datadog-Agent repository to make use of the Active Directory component.


components/activedirectory/active_directory.go
```diff
@@ -0,0 +1,130 @@
...
+	if host.OS.Descriptor().Family() != os.WindowsFamily {
+		// Print flavor in case OS family don't match, as it's more precise than the family (and the .String() conversion already exists).
+		return nil, nil, fmt.Errorf("wrong Operating System family, expected Windows, got %s", host.OS.Descriptor().Flavor.String())
+	}
+
+	params, err := common.ApplyOption(&Configuration{}, options)
+	if err != nil {
+		return nil, nil, err
+	}
+	// Make the resource options that will be passed to the component
```
pducolin, line 69: **💬 suggestion**
```suggestion
	// Tell the component its parent is host and allow to delete it with host. This will prevent uninstalling AD on host delete, we will directly delete the host
```

components/activedirectory/active_directory.go
```diff
@@ -0,0 +1,130 @@
...
+
+		if params.DomainControllerConfiguration != nil {
+			err = adCtx.installDomainController(params.DomainControllerConfiguration)
+			if err != nil {
+				return err
+			}
+		}
+
+		if len(params.DomainUsers) > 0 {
+			// Create users in parallel
```
pducolin, line 103: **👏 praise**
Yay!

components/activedirectory/active_directory.go
```diff
@@ -0,0 +1,130 @@
...
+			var createUserResources []pulumi.Resource
+			for _, user := range params.DomainUsers {
+				createDomainUserCmd, err := host.OS.Runner().Command(comp.namer.ResourceName("create-domain-users", user.Username), &command.Args{
```
pducolin, line 106: **💬 suggestion**
You might want to early check that `Username` do not have duplicates, or this will fail as pulumi expects unique command names

components/activedirectory/domain_user.go
```diff
@@ -0,0 +1,20 @@
...
+// DomainUser represents an Active Directory user
+type DomainUser struct {
+	Username string
+	Password string
+}
+
+// WithDomainUser adds a user in Active Directory.
+// Note: We don't need to be a Domain Controller to create new user in AD but we need
+// the necessary rights to modify the AD.
+func WithDomainUser(username, password string) func(params *Configuration) error {
```
pducolin, line 12: **💬 suggestion**
I would move all the `With` to an unique file, they're not that many

go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/pulumi/pulumi-random/sdk/v4 v4.16.0
 	github.com/pulumi/pulumi-tls/sdk/v4 v4.11.1
 	github.com/pulumi/pulumi/sdk/v3 v3.108.1
+	github.com/pulumiverse/pulumi-time/sdk v0.0.0-20231010123146-089d7304da13
```
pducolin, line 28: **💬 suggestion**
You can remove the explicit `pulumi-time` plugin install call in https://github.com/DataDog/test-infra-definitions/blob/main/Dockerfile#L112

pducolin (approved):
Very nice!

go.mod
julien-lebot: Good point !

components/activedirectory/domain_user.go
julien-lebot: Done ✔️

PR #689, opened 2024-03-14
[create-vm-verify-pipeline] Added pipeline verification to deploy

CelianR:
What does this PR do?
---------------------

This is a follow-up PR to https://github.com/DataDog/datadog-agent/pull/23581.
It prints a detailed error if the pipeline is outdated as images from pipelines expire after 24h.

-------------------

`create-vm` task

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/deploy.py
+++ b/tasks/deploy.py
@@ -2,11 +2,13 @@ import os
 import subprocess
 from typing import Any, Callable, Dict, List, Optional
 
+import boto3
 from invoke.context import Context
 from invoke.exceptions import Exit
+from invoke.tasks import task
 from pydantic import ValidationError
 
-from . import config, pipeline, tool
+from . import config, tool
 from .config import Config, get_full_profile_path
 
 default_public_path_key_name = "ddinfra:aws/defaultPublicKeyPath"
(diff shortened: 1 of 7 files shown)


KevinFairise2:
Small comment, but that's a very cool feature!

tasks/deploy.py
```diff
@@ -59,6 +62,16 @@ def deploy(
     aws_account = cfg.get_aws().get_account()
     flags.setdefault("ddinfra:env", "aws/" + aws_account)
 
+    # Verify image deployed and not outdated in s3
+    if deploy_job is not None and pipeline_id is not None:
+        cmd = f'inv -e check-s3-image-exists --pipeline-id={pipeline_id} --deploy-job={deploy_job}'
+        cmd = tool.get_aws_wrapper(aws_account) + cmd
+        output = ctx.run(cmd)
```
KevinFairise2, line 69: If you use `ctx.run(cmd)` without any other parameters, if the executed cmd fails (non-zero exit code) it will automatically fails the whole python execution.

If you want output to include the result and the exit_code of the command even when it fail you should use `ctx.run(cmd, warn=True)`.

KevinFairise2 (approved):
One minor comment. But LGTM

PR #692, opened 2024-03-15
[CONTINT-3776] Change refresh period of the SBOM and container image checks

AliDatadog:
What does this PR do?
---------------------
This check reduces the refresh period of the sbom and container image checks

-------------------
All k8s scenarios

Motivation
----------

Additional Notes
----------------


L3n41c (approved):

L3n41c (approved):
This will impact the fake intake memory consumption.
Let’s closely monitor if it’s OOMing.

PR #694, opened 2024-03-19
Fix lint-go

KevinFairise2:
What does this PR do?
---------------------

FIx lint-go

-------------------

Motivation
----------

Additional Notes
----------------


gjulianm (approved):
Sorry about that, did not notice the linter was failing. Thanks for fixing it!

KevinFairise2:
> Sorry about that, did not notice the linter was failing. Thanks for fixing it!

No worries, actually you could not see it on the PR and the merge queue merged the PR with the linter failing. I update the configuration of the MQ so it should not happen again

PR #700, opened 2024-03-20
Start gdb server inside microvms

usamasaqib:
What does this PR do?
---------------------
This PR adds a qemu command line option to start the gdb server inside the microvms.

-------------------

Motivation
----------

Additional Notes
----------------


scenarios/aws/microVMs/microvms/domain.go
```diff
@@ -139,6 +139,9 @@ func newDomainConfiguration(e *config.CommonEnvironment, set *vmconfig.VMSet, vc
 		hostOS = "linux" // Remote VMs are always on Linux hosts
 	}
 
+	qemuArgs := map[string]pulumi.StringInput{
+		"-s": pulumi.String(""),
```
brycekahle, line 143: do we always want this? Can you add a comment about what this flag does too?

PR #702, opened 2024-03-20
[CONTINT-3935] Create a priorityClass for the agent

L3n41c:
What does this PR do?
---------------------

Create a `priorityClass` for the agent.

-------------------

* `aws/eks`
* `aws/kindvm`

Motivation
----------

In the `datadog-agent` repository, the `TestEKSSuite` suite is currently flaky and fails to complete the pulumi stack creation from time to time with following error:
```
resource "urn:pulumi:ci-30352959-4670-eks-cluster::e2eci::pulumi:providers:kubernetes$dd:apps$kubernetes:apps/v1:Deployment::dogstatsd-uds-8125" was successfully created, but the Kubernetes API server reported that it failed to fully initialize or become live: 'dogstatsd-uds' timed out waiting to be Ready
```

It turned out that the `dogstatsd-uds` pod remain in `ContainerCreating` forever:
```
NOMINATED NODE   READINESS GATES
        workload-dogstatsd              pod/dogstatsd-uds-6b9544d9cf-9stph                     0/1     ContainerCreating   0          10m     <none>        ip-10-1-48-101.ec2.internal   <none>           <none>
```

This is because it cannot mount the dogstatsd socket that is supposed to be created by the agent.
Indeed, the agent pod supposed to run on the same node isn’t starting either:
```
        NAMESPACE                       NAME                                                   READY   STATUS              RESTARTS   AGE     IP            NODE                          NOMINATED NODE   READINESS GATES
        datadog                         pod/dda-linux-datadog-2twwr                            0/3     Pending             0          9m39s   <none>        <none>                        <none>           <none>
```

The issue is that recent PRs have added more fake app pods on the EKS cluster and EKS nodes have a capacity of 17 pods.
From time to time, the spread of pods on nodes is unbalanced and some nodes receive 17 pods before the agent is started.
image
In such a case, we must allow the agent to evict another pod so that it will be scheduled elsewhere.
That’s the purpose of the `priorityClass`.

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/kubernetes_helm.go
+++ b/components/datadog/agent/kubernetes_helm.go
@@ -241,6 +241,7 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 				"tag":           pulumi.String(agentImageTag),
 				"doNotCheckTag": pulumi.Bool(true),
 			},
+			"priorityClassCreate": pulumi.Bool(true),
 			"podAnnotations": pulumi.StringMap{
 				"ad.datadoghq.com/agent.checks": pulumi.String(utils.JSONMustMarshal(
 					map[string]interface{}{


vboulineau (approved):

pducolin:
Should we instead reduce and limit the number of apps on the agent's node? Do we need all those apps within the same test? Now that we have a kind provider and the option to deploy apps, I think we can reduce the apps in a tests to the required ones

vboulineau:
The Agent is on all nodes. The issue is that it depends when the nodes join the cluster. If a node joins before others, it will get all the pods because the scheduler cannot know more nodes will join in the future.

L3n41c:
@pducolin Starting and stopping apps at the beginning and end of every single test instead of parallelizing would considerably increase the duration of the e2e tests, which are already lasting more than 1 hour.

pducolin:
Is `priorityClass` something we recommend to customers? Can we balance apps on nodes? Do we deploy all apps to all nodes? My concern is we hack the test setup to stabilise it, making the setup less realistic.

PR #707, opened 2024-03-22
CI(auto-bump): Try to fix auto bump

KevinFairise2:
What does this PR do?
---------------------
The auto-bump is not working as intendent since it runs on pull_request close event, and it does not take the commit on main branch

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/.github/workflows/auto-bump-datadog-agent.yml
+++ b/.github/workflows/auto-bump-datadog-agent.yml
@@ -26,12 +26,31 @@ jobs:
           path: datadog-agent
           token: ${{ steps.app-token.outputs.token }}
           fetch-depth: 0
+
+      - name: Clone test-infra-definitions repo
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-bump-datadog-agent.yml
```diff
@@ -26,12 +26,31 @@ jobs:
           path: datadog-agent
           token: ${{ steps.app-token.outputs.token }}
           fetch-depth: 0
+
```
pducolin, line 29: **💬 suggestion**
What if you instead change the trigger to 
```yaml
on:
  push:
    branches:
      - main
```

KevinFairise2: I will try that, thanks!

pducolin (approved):

PR #713, opened 2024-03-25
Use latest nightly version of the agent instead of stable as default

KevinFairise2:
What does this PR do?
---------------------

Change the default version of the datadog-agent package used from latest stable to latest nightly.

Users running e2e tests locally prefer to have a version of the agent closer to their changes. Latest stable can be quite different from the latest main version of the agent. Latest nightly is closer to the current main state

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/host_linuxos.go
+++ b/components/datadog/agent/host_linuxos.go
@@ -46,12 +46,8 @@ func (am *agentLinuxManager) getInstallCommand(version agentparams.PackageVersio
 		commandLine += fmt.Sprintf("DD_AGENT_MINOR_VERSION=%v ", version.Minor)
 	}
 
-	if version.BetaChannel {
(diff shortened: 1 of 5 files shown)


components/datadog/agent/host_linuxos.go
```diff
@@ -50,6 +50,10 @@ func (am *agentLinuxManager) getInstallCommand(version agentparams.PackageVersio
 		commandLine += "REPO_URL=datad0g.com DD_AGENT_DIST_CHANNEL=beta "
 	}
 
+	if version.NightlyChannel {
```
pducolin, line 53: **💬 suggestion**
We might refactor this to support `channel` of type `Channel` that is `string` with possible values `prod|beta|nightly`

pducolin (approved):

PR #717, opened 2024-03-27
e2e(kubernetes agent): Change WithHelmValues to not override everytime but to merge them

KevinFairise2:
What does this PR do?
---------------------

Change the `WithHelmValues` optional args of `KubernetesAgent`. This way we can add several `WithHelmValues` args and they are merged automatically

-------------------

Motivation
----------

For `Kind` scenario we need to install the agent with some Helm values. Without this change you are not able to use `WithHelmValues` without specifying all the helm value needed internally to run on kind

Additional Notes
----------------


vboulineau (changes requested):
Helm already does the merge, that's why it's a `AssetOrArchiveArray`. You simply append (to the end) newer values to the array.

KevinFairise2:
> Helm already does the merge, that's why it's a `AssetOrArchiveArray`. You simply append (to the end) newer values to the array.

Oh that's nice. So I'd just need to replace 
```
p.HelmValues = pulumi.AssetOrArchiveArray{pulumi.NewStringAsset(values)}
```
by
```
p.HelmValues = append(p.HelmValues, pulumi.NewStringAsset(values))
```

Thanks for pointing it out!

vboulineau (approved):

pducolin (approved):

PR #722, opened 2024-03-28
Add netflow-generator workload app.

jmw51798:
What does this PR do?
---------------------
Add netflow-generator workload app.

-------------------
This is simply a repackaging of networkstatic/nflow-generator so we can pull the image from our own registry to avoid any Docker Hub rate limits.  It will initially only be used in a docker environment by a new ndm/netflow new-e2e test.

Motivation
----------

Additional Notes
----------------


pducolin (approved):
Nice and welcome back!

PR #738, opened 2024-04-04
[components/datadog] Add deployments with APM instrumentation

davidor:
What does this PR do?
---------------------
This PR configures APM lib injection in a specific namespace. It also creates a couple of deployments in that namespace. One specifies the library in an annotation and the other one doesn't and relies on language detection instead.


davidor:
Thanks @L3n41c I've addressed your comment.

L3n41c (approved):

PR #760, opened 2024-04-12
installer: fix installing on SUSE

chouquette:
What does this PR do?
---------------------

This PR addresses the specificities of installing on suse

-------------------

Installer tests when running on SUSE

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/updater/install_script.sh
+++ b/components/datadog/updater/install_script.sh
@@ -96,7 +96,6 @@ elif [ "${OS}" = "RedHat" ]; then
     $sudo_cmd yum -y install datadog-installer
 elif [ "${OS}" = "SUSE" ]; then
     yum_url="yumtesting.datad0g.com/suse/testing"
-    yum_url="yumtesting.datad0g.com/testing"
     yum_repo_version="${DD_PIPELINE_ID}-i7/7"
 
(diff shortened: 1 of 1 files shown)


components/datadog/updater/install_script.sh
```diff
@@ -98,5 +94,18 @@ elif [ "${OS}" = "RedHat" ]; then
     $sudo_cmd sh -c "echo -e '[datadog]\nname = Datadog, Inc.\nbaseurl = https://${yum_url}/${yum_repo_version}/${ARCH}/\nenabled=1\ngpgcheck=1\nrepo_gpgcheck=1\npriority=1\ngpgkey=${gpgkeys}' > /etc/yum.repos.d/datadog.repo"
     $sudo_cmd yum -y clean metadata
     $sudo_cmd yum -y install datadog-installer
+elif [ "${OS}" = "SUSE" ]; then
+    yum_url="yumtesting.datad0g.com/suse/testing"
+    yum_url="yumtesting.datad0g.com/testing"
```
KevinFairise2, line 99:
```suggestion
```
Otherwise you override the previous line with something that won't work on suse no?

KevinFairise2 (approved):

components/datadog/updater/install_script.sh
chouquette: Indeed, however it's interesting that I was able to create a VM with that code :thinking: 

Maybe there aren't that many differences and we could just merge SUSE & RPM?!

PR #762, opened 2024-04-15
Add WithFileWithPermissions to create files with predefined permissions

Kaderinho:
What does this PR do?
---------------------

Add a way to specify permissions when creating a file. Only on Unix for now but will follow up for Windows permissions 

Example PR: https://github.com/DataDog/datadog-agent/pull/24723

-------------------

Any scenarios that needs to set specific permissions to a file. 

Motivation
----------

Currently to test secret feature, we need to do the following:

1. UpdateEnv without the Agent config
2. Give permissions to the secret script
3. UpdateEnv with the Agent config

In the first UpdateEnv, we can add the secret script file but we cannot give it the right permissions. So if we set `secret_backend_command` directly the Agent won't be able to start and the first UpdateEnv will fail.

Additional Notes
----------------

### DIFF
--- a/components/datadog/agentparams/filepermissions/permissions.go
+++ b/components/datadog/agentparams/filepermissions/permissions.go
@@ -1,0 +1,8 @@
+package filepermissions
+
+// FilePermissions interface defines all commands that can be used to set and reset file permissions.
+// These functions are used to create args.Command that will be run by pulumi.
+type FilePermissions interface {
+	SetupPermissionsCommand(path string) string
+	ResetPermissionsCommand(path string) string
+}
--- a/components/datadog/agentparams/filepermissions/permissions_nix.go
+++ b/components/datadog/agentparams/filepermissions/permissions_nix.go
@@ -1,0 +1,82 @@
+package filepermissions
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/DataDog/datadog-agent/pkg/util/optional"
+	"github.com/DataDog/test-infra-definitions/common"
+)
+
+type UnixPermissionsOption = func(*UnixPermissions) error
+
+// UnixPermissions represents the owner, group, and permissions of a file.
+// Permissions are represented by a string (should be an octal). There is no check on the octal format.
+type UnixPermissions struct {
+	Owner       string
+	Group       string
+	Permissions string
+}
+
+var _ FilePermissions = (*UnixPermissions)(nil)
+
+// NewUnixPermissions creates a new UnixPermissions object and applies the given options.
+func NewUnixPermissions(options ...UnixPermissionsOption) optional.Option[FilePermissions] {
+	p, err := common.ApplyOption(&UnixPermissions{}, options)
+
+	if err != nil {
+		return optional.NewNoneOption[FilePermissions]()
+	}
+	return optional.NewOption[FilePermissions](p)
+}
+
+// SetupPermissionsCommand returns a command that sets the owner, group, and permissions of a file.
+func (p *UnixPermissions) SetupPermissionsCommand(path string) string {
+	var commands []string
+
+	if p.Owner != "" {
+		commands = append(commands, fmt.Sprintf("chown %s %s", p.Owner, path))
+	}
+
+	if p.Group != "" {
+		commands = append(commands, fmt.Sprintf("chgrp %s %s", p.Group, path))
+	}
+
+	if p.Permissions != "" {
+		commands = append(commands, fmt.Sprintf("chmod %s %s", p.Permissions, path))
+	}
+
+	if len(commands) == 0 {
+		return ""
+	}
+	return fmt.Sprintf(`sudo sh -c "%v"`, strings.Join(commands, " && "))
+}
+
+// ResetPermissionsCommand returns a command that resets the owner, group, and permissions of a file to default.
+func (p *UnixPermissions) ResetPermissionsCommand(path string) string {
+	return fmt.Sprintf("sudo chown ubuntu:ubuntu %s && sudo chmod 644 %s", path, path)
+}
+
+// WithOwner sets the owner of the file.
+func WithOwner(owner string) UnixPermissionsOption {
+	return func(p *UnixPermissions) error {
+		p.Owner = owner
+		return nil
+	}
+}
+
+// WithGroup sets the group of the file.
+func WithGroup(group string) UnixPermissionsOption {
+	return func(p *UnixPermissions) error {
+		p.Group = group
+		return nil
+	}
+}
+
+// WithPermissions sets the permissions of the file.
+func WithPermissions(permissions string) UnixPermissionsOption {
+	return func(p *UnixPermissions) error {
+		p.Permissions = permissions
+		return nil
+	}
+}
--- a/components/datadog/agentparams/filepermissions/permissions_nix.go
+++ b/components/datadog/agentparams/filepermissions/permissions_nix.go
@@ -25,7 +25,7 @@ func NewUnixPermissions(options ...UnixPermissionsOption) optional.Option[FilePe
 	p, err := common.ApplyOption(&UnixPermissions{}, options)
 
 	if err != nil {
-		return optional.NewNoneOption[FilePermissions]()
+		panic("Could not create UnixPermissions: " + err.Error())
 	}
 	return optional.NewOption[FilePermissions](p)
 }
--- a/components/datadog/agentparams/filepermissions/permissions_nix.go
+++ b/components/datadog/agentparams/filepermissions/permissions_nix.go
@@ -49,7 +49,7 @@ func (p *UnixPermissions) SetupPermissionsCommand(path string) string {
 	if len(commands) == 0 {
 		return ""
 	}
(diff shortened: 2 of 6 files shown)


components/datadog/agentparams/filepermissions/permissions_nix.go
```diff
@@ -0,0 +1,82 @@
...
+
+	if len(commands) == 0 {
+		return ""
+	}
+	return fmt.Sprintf(`sudo sh -c "%v"`, strings.Join(commands, " && "))
+}
+
+// ResetPermissionsCommand returns a command that resets the owner, group, and permissions of a file to default.
+func (p *UnixPermissions) ResetPermissionsCommand(path string) string {
+	return fmt.Sprintf("sudo chown ubuntu:ubuntu %s && sudo chmod 644 %s", path, path)
```
pgimalac, line 57: I don't think you can assume the default user/group on `Unix` is `ubuntu` (it won't be if using anything else than `Ubuntu`)

components/datadog/agentparams/filepermissions/permissions_nix.go
```diff
@@ -0,0 +1,82 @@
...
+		commands = append(commands, fmt.Sprintf("chown %s %s", p.Owner, path))
+	}
+
+	if p.Group != "" {
+		commands = append(commands, fmt.Sprintf("chgrp %s %s", p.Group, path))
+	}
+
+	if p.Permissions != "" {
+		commands = append(commands, fmt.Sprintf("chmod %s %s", p.Permissions, path))
+	}
```
pgimalac, line 47: **💭 thought:** ‏You might need to prefix every command with `sudo` ?

Otherwise I guess with some combination it would fail, eg. first change the user and group to `root`, then try to change the permissions

components/datadog/agentparams/params.go
```diff
@@ -175,10 +178,16 @@ func WithIntegration(folderName string, content string) func(*Params) error {
 
 // WithFile adds a file with contents to the install at the given path. This should only be used when the agent needs to be restarted after writing the file.
 func WithFile(absolutePath string, content string, useSudo bool) func(*Params) error {
+	return WithFileWithPermissions(absolutePath, content, useSudo, optional.NewNoneOption[perms.FilePermissions]())
+}
+
+// WithFileWithPermissions adds a file like WithFile but we can predefine the permissions of the file.
+func WithFileWithPermissions(absolutePath string, content string, useSudo bool, perms optional.Option[perms.FilePermissions]) func(*Params) error {
```
pgimalac, line 185: **❓ question:** ‏Do we really need an `optional.Option` here ?
It seems like you just do nothing if the permissions are empty anyway

components/datadog/agentparams/filepermissions/permissions.go
```diff
@@ -0,0 +1,8 @@
+package filepermissions
+
+// FilePermissions interface defines all commands that can be used to set and reset file permissions.
+// These functions are used to create args.Command that will be run by pulumi.
+type FilePermissions interface {
+	SetupPermissionsCommand(path string) string
+	ResetPermissionsCommand(path string) string
+}
```
pgimalac, line 8: **👏 praise:** ‏Nice to have something super generic so that we can have very custom implementations for each platform 👍

components/datadog/agentparams/params.go
Kaderinho: You're right, this is useless here

components/datadog/agentparams/filepermissions/permissions_nix.go
Kaderinho: You're right, I've just tested this scenario and it failed without `sudo` 👍

Kaderinho: Actually my test was failing for another reason 😓  so i think the first sudo does the job. I'll keep all `sudo` anyway as it does not hurt I believe

components/datadog/agentparams/filepermissions/permissions.go
```diff
@@ -0,0 +1,8 @@
+package filepermissions
+
+// FilePermissions interface defines all commands that can be used to set and reset file permissions.
+// These functions are used to create args.Command that will be run by pulumi.
+type FilePermissions interface {
+	SetupPermissionsCommand(path string) string
```
pgimalac, line 6: **💭 thought:** ‏I'm thinking it could be eg. `SetupPermissions(path string, executor interface{Execute() error})` so that we can do some more complex conditions on the commands (parse command output, do different things depending on errors, etc)

But no issue if you want to keep the current version, that works too

components/datadog/agentparams/filepermissions/permissions_nix.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+var _ FilePermissions = (*UnixPermissions)(nil)
+
+// NewUnixPermissions creates a new UnixPermissions object and applies the given options.
+func NewUnixPermissions(options ...UnixPermissionsOption) optional.Option[FilePermissions] {
+	p, err := common.ApplyOption(&UnixPermissions{}, options)
+
+	if err != nil {
+		return optional.NewNoneOption[FilePermissions]()
```
pgimalac, line 28: Do you think we should silently ignore the error if there is one ?
IMO the whole test should stop if there is an error when creating the options struct

components/datadog/agentparams/filepermissions/permissions_nix.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return optional.NewOption[FilePermissions](p)
+}
+
+// SetupPermissionsCommand returns a command that sets the owner, group, and permissions of a file.
+func (p *UnixPermissions) SetupPermissionsCommand(path string) string {
+	var commands []string
+
+	if p.Owner != "" {
+		commands = append(commands, fmt.Sprintf("sudo chown %s %s", p.Owner, path))
```
pgimalac, line 38: **💭 thought:** ‏I hope no one will have that, but if `path` contains spaces then the commands won't work properly 😄

components/datadog/agentparams/filepermissions/permissions_nix.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	if p.Permissions != "" {
+		commands = append(commands, fmt.Sprintf("sudo chmod %s %s", p.Permissions, path))
+	}
+
+	if len(commands) == 0 {
+		return ""
+	}
+	return fmt.Sprintf(`sudo sh -c "%v"`, strings.Join(commands, " && "))
```
pgimalac, line 52: **🥜 nitpick:** ‏I guess you don't need to exec in a subshell

components/datadog/agentparams/filepermissions/permissions_nix.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// ResetPermissionsCommand returns a command that resets the owner, group, and permissions of a file to default.
+func (p *UnixPermissions) ResetPermissionsCommand(path string) string {
+	return fmt.Sprintf("sudo chown $USER %s && sudo chmod 644 %s", path, path)
```
pgimalac, line 57:
```suggestion
	return fmt.Sprintf("sudo chown $USER:$(id -gn) %s && sudo chmod 644 %s", path, path)
```
If you want to be super specific this will also reset the group 😄

components/datadog/agentparams/filepermissions/permissions_nix.go
pgimalac: I actually didn't realize you were running all the commands in a subshell, you really didn't need it 🤦

components/datadog/agentparams/filepermissions/permissions.go
Kaderinho: Hmm sounds like an interesting idea, but I'll rather keep this PR simple and add `executor` later when we have a use case for it

components/datadog/agentparams/filepermissions/permissions_nix.go
Kaderinho: You're right, I quoted everything here 880744def3463c076ffbc447f5c1bd305a016a1e

components/datadog/agentparams/filepermissions/permissions_nix.go
Kaderinho: I hesitated because nothing in `options` return erorr but I'm fine with stopping the test as well
 errore26100265f89e40d3ed75c5cebcc13ccec2de4c7

components/datadog/agentparams/filepermissions/permissions_nix.go
Kaderinho: 👍  191ca79db6e87e0d387b815983faafdc097b0ac5

pducolin (approved):

components/datadog/agentparams/params.go
```diff
@@ -5,9 +5,11 @@ import (
 	"path"
 	"strings"
 
+	"github.com/DataDog/datadog-agent/pkg/util/optional"
```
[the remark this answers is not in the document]
Kaderinho: We need a way to check that the object exists before calling interface methods but we can easily change it for a pointer if we have issues

pgimalac (approved):

PR #763, opened 2024-04-15
use Pull through cache

KevinFairise2:
What does this PR do?
---------------------

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/kubernetes/kind-cluster.yaml
+++ b/components/kubernetes/kind-cluster.yaml
@@ -2,10 +2,15 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
+  extraMounts:
+    - containerPath: /var/lib/kubelet/config.json
+      hostPath: /tmp/kind-config/config.json
+      propagation: HostToContainer
 containerdConfigPatches:
   - |-
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
       endpoint = ["https://mirror.gcr.io", "https://registry-1.docker.io"]
+
 networking:
   apiServerAddress: "0.0.0.0"
   apiServerPort: 8443
--- a/components/kubernetes/kind-cluster.yaml
+++ b/components/kubernetes/kind-cluster.yaml
@@ -10,7 +10,6 @@ containerdConfigPatches:
   - |-
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
       endpoint = ["https://mirror.gcr.io", "https://registry-1.docker.io"]
-
 networking:
   apiServerAddress: "0.0.0.0"
   apiServerPort: 8443
--- a/components/kubernetes/kind.go
+++ b/components/kubernetes/kind.go
@@ -42,6 +42,7 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
 		if err != nil {
 			return err
 		}
+
 		opts = utils.MergeOptions(opts, utils.PulumiDependsOn(dockerInstallCmd, curlCommand))
 
 		kindVersionConfig, err := getKindVersionConfig(kubeVersion)
--- a/components/kubernetes/kind.go
+++ b/components/kubernetes/kind.go
@@ -42,7 +42,6 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
 		if err != nil {
 			return err
 		}
-
 		opts = utils.MergeOptions(opts, utils.PulumiDependsOn(dockerInstallCmd, curlCommand))
 
 		kindVersionConfig, err := getKindVersionConfig(kubeVersion)
--- a/resources/azure/environment.go
+++ b/resources/azure/environment.go
@@ -64,6 +64,10 @@ func (e *Environment) InternalRegistry() string {
 	return "none"
 }
 
+func (e *Environment) InternalDockerhubMirror() string {
+	return "none"
+}
+
 // Common
 func (e *Environment) DefaultResourceGroup() string {
 	return e.GetStringWithDefault(e.InfraConfig, DDInfraDefaultResourceGroup, e.envDefault.ddInfra.defaultResourceGroup)
--- a/resources/azure/environment.go
+++ b/resources/azure/environment.go
@@ -65,7 +65,7 @@ func (e *Environment) InternalRegistry() string {
 }
 
 func (e *Environment) InternalDockerhubMirror() string {
-	return "none"
+	return "registry-1.docker.io"
 }
 
 // Common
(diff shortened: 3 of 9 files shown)


pducolin (approved):

resources/azure/environment.go
```diff
@@ -64,6 +64,10 @@ func (e *Environment) InternalRegistry() string {
 	return "none"
 }
 
+func (e *Environment) InternalDockerhubMirror() string {
+	return "none"
```
vboulineau, line 68: You can return public DockerHub so that it does not fail: `registry-1.docker.io`

scenarios/aws/kindvm/run.go
```diff
@@ -39,9 +45,20 @@ func Run(ctx *pulumi.Context) error {
 		return err
 	}
 
+	// Install docker if not installed yet, we need it to configure docker credentials
+	_, dockerInstallCmd, err := docker.NewManager(*awsEnv.CommonEnvironment, vm)
```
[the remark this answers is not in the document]
pducolin: It's fine, I am wondering if we should have it inside the `NewKindCluster` component, as it needs it. Can be done later

scenarios/aws/kindvm/run.go
```diff
@@ -154,3 +170,36 @@ agents:
 
 	return nil
 }
+
+func ConfigureECRCredentials(e resAws.Environment, vm *remote.Host, arch os.Architecture, opts ...pulumi.ResourceOption) (*goremote.Command, error) {
```
pducolin, line 174: **💬 suggestion**
If this is meant to be used from `datadog-agent`, could be moved to `docker` or some helpers package, rather than scenarios.

components/kubernetes/kind-cluster.yaml
```diff
@@ -2,10 +2,15 @@ kind: Cluster
...
 - role: control-plane
+  extraMounts:
+    - containerPath: /var/lib/kubelet/config.json
+      hostPath: /tmp/kind-config/config.json
+      propagation: HostToContainer
 containerdConfigPatches:
   - |-
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
       endpoint = ["https://mirror.gcr.io", "https://registry-1.docker.io"]
+
```
KevinFairise2, line 13:
```suggestion
```

components/kubernetes/kind.go
```diff
@@ -42,6 +42,7 @@ func NewKindCluster(env config.CommonEnvironment, vm *remote.Host, resourceName,
 		if err != nil {
 			return err
 		}
+
```
KevinFairise2, line 45:
```suggestion
```

scenarios/aws/kindvm/run.go
```diff
@@ -154,3 +170,36 @@ agents:
...
+	architecture := "x86_64"
+	if arch == os.ARM64Arch {
+		architecture = "aarch64"
+	}
+
+	unzipInstallCommand, err := vm.OS.PackageManager().Ensure("unzip", nil, "", opts...)
```
pducolin, line 180: **💭 thought**
Unrelated, noticing now that calls to `Ensure` do not pin versions, might be something we work on

scenarios/aws/kindvm/run.go
```diff
@@ -154,3 +170,36 @@ agents:
...
+		return nil, err
+	}
+
+	ecrLoginCommand, err := vm.OS.Runner().Command(
+		e.CommonNamer.ResourceName("ecr-login"),
+		&command.Args{
+			Create: pulumi.Sprintf("aws ecr get-login-password | docker --config /tmp/kind-config login  --username AWS --password-stdin %s", e.CloudProviderEnvironment.InternalRegistry()),
+		},
+		utils.PulumiDependsOn(awsCliInstallCommand),
+	)
```
pducolin, line 202: **❓ question**
Oh so this is generating the `config.json` that is mounted to kind?

pducolin (approved):
Minor comments that can be addressed in a separate PR, thank you so much for this!

scenarios/aws/kindvm/run.go
KevinFairise2: Yep we retrieve the password from `aws` cli and we use it in the `docker login` command that will generate an auth token stored in the config file specified with `--config`
Note: I verified that the auth token generated expires when the password given by AWS does. The token should remaind valid for 12 hours

scenarios/aws/kindvm/run.go
KevinFairise2: I think replacing `unzip` by `unzip==<version>` could work. But would need to be tested

scenarios/aws/kindvm/run.go
KevinFairise2: Since it is something specific to AWS I did not want to have it in `components/docker` that should remain cloud agnostic. But yes we can probably find a better place for it

scenarios/aws/kindvm/run.go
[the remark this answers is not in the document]
KevinFairise2: Would have been easier, but we do not want to put AWS specific logic in the `NewKindCluster` component. That could be used on any cloud provider

scenarios/aws/kindvm/run.go
vboulineau: I am not sure there is an added value in pinning versions, for two reasons:
- Versions are usually pinned by distro release (normally only minor/security changes).
- We don't pin AMIs and in most cases we don't even pin OS version.

scenarios/aws/kindvm/run.go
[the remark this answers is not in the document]
vboulineau: We'll need to duplicate this code in `datadog-agent` with current code. However you could have a `WithECRCredentials` that implements this. Note the `ExtraMount` you're adding means that it will currently fail when not using any credential helper. You may need to always create a empty `{}` JSON file.

You also don't need `docker login`, what `docker login` does is very basic JSON gen. The way I see it the clean way would be to build a small `Go` code that imports https://github.com/awslabs/amazon-ecr-credential-helper/blob/main/ecr-login/ecr.go#L83-L110 and produces the `config.json` on `stdout`.

It's much smaller and faster to download then the AWS SDK and you have the option to package as a docker container and get the `config.json` with `docker run --network host <your_helper_image> > config-tmp.json`.

scenarios/aws/kindvm/run.go
pducolin: We do pin some AMIs, for example for the installer tests. We don't probably want to maintain all images and versions, but for folks who would like to, we might want to offer the option. No one ever asked for it, just noticing it now.

scenarios/aws/kindvm/run.go
pducolin: Maybe `resources/aws/ecr`

PR #764, opened 2024-04-16
[FA] Bootstrap agents on lab hosts

coignetp:
What does this PR do?
---------------------

Bootstrap agents on the installer lab hosts

-------------------

* Installer lab
* VM creation with installer (should be a noop)

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/updater/install.go
+++ b/components/datadog/updater/install.go
@@ -3,6 +3,7 @@ package updater
 import (
 	_ "embed"
 	"fmt"
+	"strings"
 
 	"github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/namer"
@@ -28,9 +29,6 @@ type HostUpdater struct {
 	namer namer.Namer
 	host  *remoteComp.Host
 }
-
-const installerPath = "/opt/datadog-installer/bin/installer/installer"
-const latestOciUrlFormatString = "oci://docker.io/datadog/%s:latest"
 
 func (h *HostUpdater) Export(ctx *pulumi.Context, out *HostUpdaterOutput) error {
 	return components.Export(ctx, h, out)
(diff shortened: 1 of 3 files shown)


BaptisteFoy (approved):

components/datadog/updater/install.go
```diff
@@ -87,16 +86,20 @@ func NewHostUpdaterWithPackages(e *config.CommonEnvironment, host *remoteComp.Ho
...
 	pipelineID := fmt.Sprintf("DD_PIPELINE_ID=%v", params.Version.PipelineID)
-	installCmdStr := fmt.Sprintf(`export %v && bash -c %s`, pipelineID, installScript)
-
-	for _, pkg := range packages {
-		installCmdStr = fmt.Sprintf("%s\nsudo %s bootstrap -P %s", installCmdStr, installerPath, pkg)
+	agentConfig := pulumi.Sprintf("")
+	for _, extraConfig := range params.ExtraAgentConfig {
+		agentConfig = pulumi.Sprintf("%v\n%v", agentConfig, extraConfig)
 	}
+	agentConfig = pulumi.Sprintf("AGENT_CONFIG='%v'", agentConfig)
```
paullegranddc, line 93: I think that doesn't work if the extra configs contain `'` .
I don't think it's worth escaping them but you should document that somewhere

components/datadog/updater/install_script.sh
```diff
@@ -82,6 +84,14 @@ if [ "${OS}" = "Debian" ]; then
 
     $sudo_cmd DEBIAN_FRONTEND=noninteractive apt-get update
     $sudo_cmd apt-get install -y --force-yes datadog-installer
+
+    # Add packages
+    $sudo_cmd systemctl stop datadog-agent
+    for pkg in $PACKAGES; do
+        $sudo_cmd $INSTALLER_BIN bootstrap --url "${OCI_URL_PREFIX}${pkg}:latest"
```
paullegranddc, line 91: I think it'd be better if we passed ${pkg:version} as parameter instead of always using latest since you might want to test a new package without having to push a new OCI tagged lattes

KevinFairise2 (approved):

PR #769, opened 2024-04-16
Use restart-service powershell command to restart the agent

KevinFairise2:
What does this PR do?
---------------------
Use `Restart-Service` from powershell instead of `restart-services` from agent.exe
The `restart-services` is known to be impacted by race conditions

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/host_windowsos.go
+++ b/components/datadog/agent/host_windowsos.go
@@ -81,7 +81,7 @@ func (am *agentWindowsManager) restartAgentServices(transform command.Transforme
 	// TODO: When we introduce Namer in components, we should use it here.
 	cmdName := am.host.Name() + "-" + "restart-agent"
 	cmdArgs := command.Args{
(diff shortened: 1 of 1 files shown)


components/datadog/agent/host_windowsos.go
```diff
@@ -81,7 +81,7 @@ func (am *agentWindowsManager) restartAgentServices(transform command.Transforme
 	// TODO: When we introduce Namer in components, we should use it here.
 	cmdName := am.host.Name() + "-" + "restart-agent"
 	cmdArgs := command.Args{
-		Create: pulumi.String(`Start-Process "$($env:ProgramFiles)\Datadog\Datadog Agent\bin\agent.exe" -Wait -ArgumentList restart-service`),
+		Create: pulumi.String(`Restart-Service datadogagent -Force`),
```
vboulineau, line 84: I wanted to do this change when I implemented the component refactor, but IIRC, this only restarts the Core Agent and not other Agents?

iglendd: -Force option will stop dependent services first.

iglendd (approved):
LGTM

PR #771, opened 2024-04-17
Make restart-service more robust to avoid weird test flakiness

KevinFairise2:
What does this PR do?
---------------------

When the `restart-service` command is used on an agent that is not fully started. The command can fail and was ignored. Which leads to an agent in a very unexpected state after a restart that does not occured.

This change should make the restart more robust as well as adding some debug logs, and a hard failure if the `restart-service` fails several times

-------------------

Motivation
----------

Additional Notes
----------------


components/datadog/agent/host_windowsos.go
```diff
@@ -80,8 +80,24 @@ func (am *agentWindowsManager) getAgentConfigFolder() string {
 func (am *agentWindowsManager) restartAgentServices(transform command.Transformer, opts ...pulumi.ResourceOption) (*remote.Command, error) {
 	// TODO: When we introduce Namer in components, we should use it here.
 	cmdName := am.host.Name() + "-" + "restart-agent"
+	cmd := fmt.Sprintf(`
```
pducolin, line 83: **💬 suggestion**
As discussed I would rather have a more robust agent start.

Noticed we don't follow the public documentation, we should directly call the binary https://docs.datadoghq.com/agent/basic_agent_usage/windows/?tab=commandline#agent-commands

KevinFairise2: Yes ideally we'd want the subcommand to handle this case. But I think this is an issue known for a long time and I do not think Windows team planned to work on it yet. So I think this is the only option we have to stabilize tests.

Sounds good to modify the script to call the binary directly instead of wrapping it in `Start-Process`

components/datadog/agent/host_windowsos.go
```diff
@@ -80,8 +80,25 @@ func (am *agentWindowsManager) getAgentConfigFolder() string {
 func (am *agentWindowsManager) restartAgentServices(transform command.Transformer, opts ...pulumi.ResourceOption) (*remote.Command, error) {
 	// TODO: When we introduce Namer in components, we should use it here.
 	cmdName := am.host.Name() + "-" + "restart-agent"
+	cmd := `
```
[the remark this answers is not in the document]
pducolin: Adding it https://datadoghq.atlassian.net/browse/WINA-747

pducolin, line 83:
```suggestion
	// Retry restart several time, workaround to https://datadoghq.atlassian.net/browse/WINA-747
	cmd := `
```

pducolin (approved):

PR #774, opened 2024-04-17
[kind] mount host proc into kind node

YoannGh:
What does this PR do?
---------------------
The system-probe container needs access to the host's `proc` filesystem for some modules to work properly (CWS).
This PR updates the kind node configuration to mount the host `/proc` into the container so that it can be further propagated to the system-probe container.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/kubernetes/kind-cluster.yaml
+++ b/components/kubernetes/kind-cluster.yaml
@@ -2,6 +2,9 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
+  extraMounts:
+  - hostPath: /proc
+    containerPath: /host/proc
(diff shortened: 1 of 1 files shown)


pducolin (approved):

components/kubernetes/kind-cluster.yaml
```diff
@@ -2,6 +2,9 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
+  extraMounts:
+  - hostPath: /proc
+    containerPath: /host/proc
```
vboulineau, line 7: It's not clear why CWS needs _real_ `/host/proc` and how you'll be able to use that in a manifest that is not Kind-aware.

PR #780, opened 2024-04-18
Add WindowsPermissions structure that can be used with WithFileWithPermissions

Kaderinho:
Equivalent of https://github.com/DataDog/test-infra-definitions/pull/762 for Windows

-------------------

Any scenarios that needs to set specific permissions to a file. See [usage example](https://github.com/DataDog/datadog-agent/pull/24829/files)

Motivation
----------
ASCII-1491
Currently to test secret feature, we need to do the following:

1. UpdateEnv without the Agent config
2. Give permissions to the secret script
3. UpdateEnv with the Agent config

In the first UpdateEnv, we can add the secret script file but we cannot give it the right permissions. So if we set secret_backend_command directly the Agent won't be able to start and the first UpdateEnv will fail.

Additional Notes
----------------

I started to work on structures containing ACE information, which would contain identifier like explained in [icacls documentation](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/icacls#remarks).
 
```
const (
       Allow ACEType = iota
       Deny
)

type WindowsACE struct {
       SID        string
       AceType    ACEType  // Allow or Deny
       AccessMask []string // FullControl, Modify, ReadAndExecute, Read, Write
       AceFlags   []string // ObjectInherit, ContainerInherit, NoPropagateInherit, InheritOnly, Inherited
}
```

For instance AccessMask could contain ["RX", "M", WDAC"] and then we would be able to transform it into `(RX,M,WDAC)` to use it with `icacls`

I'm doing this PR to have a first version for Windows and make sure I can do wrapper around secrets management. I'll try to follow up later to add a more friendly API (e.g. having function like `WithFullControl()`, `WithReadExec()`, etc ...)

### DIFF
--- a/components/datadog/agentparams/filepermissions/permissions_win.go
+++ b/components/datadog/agentparams/filepermissions/permissions_win.go
@@ -11,7 +11,10 @@ type WindowsPermissionsOption = func(*WindowsPermissions) error
 
 // WindowsPermissions contains the information to configure the permissions of a file on Windows.
 type WindowsPermissions struct {
-	// If you are familiar with the icacls command, you can provide a custom command directly. This has precedence over any other option.
+	// RemoveDefaultPermissions removes the default permissions from the file. This is useful when you want to set permissions for secrets management.
+	RemoveDefaultPermissions bool
+
+	// If you are familiar with the icacls command, you can provide a custom command directly.
 	IcaclsCommand string
 }
 
@@ -29,10 +32,14 @@ func NewWindowsPermissions(options ...WindowsPermissionsOption) optional.Option[
 
 // SetupPermissionsCommand returns a command that sets the permissions of a file. It relies on the icacls command.
 func (p *WindowsPermissions) SetupPermissionsCommand(path string) string {
-	if p.IcaclsCommand != "" {
-		return fmt.Sprintf("icacls “%v” %v", path, p.IcaclsCommand)
+	cmd := ""
+	if p.RemoveDefaultPermissions {
+		cmd = fmt.Sprintf(`icacls "%v" /remove (Get-LocalUser) "Everyone" "Users" "Administrators" "SYSTEM" "Authenticated Users" "CREATOR OWNER" /inheritance:r /t /c /l;`, path)
 	}
-	return ""
+	if p.IcaclsCommand != "" {
+		return fmt.Sprintf(`%v icacls "%v" %v`, cmd, path, p.IcaclsCommand)
+	}
+	return cmd
 }
 
 // ResetPermissionsCommand returns a command that resets the owner, group, and permissions of a file to default.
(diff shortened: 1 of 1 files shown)


components/datadog/agentparams/filepermissions/permissions_win.go
```diff
@@ -0,0 +1,64 @@
...
+		panic("Could not create WindowsPermissions: " + err.Error())
+	}
+	return optional.NewOption[FilePermissions](p)
+}
+
+// SetupPermissionsCommand returns a command that sets the permissions of a file. It relies on the icacls command.
+func (p *WindowsPermissions) SetupPermissionsCommand(path string) string {
+	cmd := ""
+	if p.RemoveDefaultPermissions {
+		cmd = fmt.Sprintf(`icacls "%v" /remove (Get-LocalUser) "Everyone" "Users" "Administrators" "SYSTEM" "Authenticated Users" "CREATOR OWNER" /inheritance:r /t /c /l;`, path)
```
Kaderinho, line 37: Feel free to give suggestions on how to do it better 😅

clarkb7: if I understand correctly, `icacls.exe /reset` will remove all explicit ACE and leave the file with only inherited ACE. Then you can use an option to decide if you want to enable or disable inheritance.

i think `icacls.exe /reset` followed by `icacls.exe /inheritance:r ...` will leave the file with only the explicit permissions specified by the `/grant` or `/deny` parameters.

if you use `get-acl path | fl` you can examine the SDDL and the ACE with the `ID` option are the inherited ACE.

Kaderinho: You're right but the goal here is to remove everything including permissions specified by the `/grant` or `/deny` that's why I'm using `remove` instead of `reset`. I'm using `/reset` in ResetPermissionsCommand for the use case you described

Kaderinho: Here, `/reset` should work in most cases (especially in our secrets e2e tests) but I  wasn't sure if we could `grant`/`deny` new permissions by default in which case `/reset` might have failed 🤷

clarkb7: `/reset` will remove all explicit permissions "`/grant` or `/deny" )

Kaderinho: You're right, I just checked and `/inheritance:r` does the job. I'll change it, thanks !

Kaderinho: I'll replace this option by `DisableInheritance` since we don't have use cases where this would not be enough yet.
We will be able to add `/reset` later if needed. Does it sound good to you ?

clarkb7: the only files with explicit permissions tests are likely to encounter are ones that they themselves create. so as long as each test uses a new file or `/reset` the permissions some other way (`ResetPermissionsCommand` ? ) inbetween tests it should be fine.

Kaderinho: I've made some modifications 7acdbecdcd90610d411b07eebf47d7f036e41417

components/datadog/agentparams/filepermissions/permissions_win.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	if p.IcaclsCommand != "" {
+		return fmt.Sprintf(`%v icacls "%v" %v`, cmd, path, p.IcaclsCommand)
+	}
+	return cmd
+}
+
+// ResetPermissionsCommand returns a command that resets the owner, group, and permissions of a file to default.
+func (p *WindowsPermissions) ResetPermissionsCommand(path string) string {
+	return fmt.Sprintf("icacls “%[1]v” /inheritance:e /t /c /l; icacls “%[1]v” /reset /t /c /l;", path)
```
clarkb7, line 47: `/inheritance:e` shouldn't be necessary, `/reset` will handle it.

clarkb7 (approved):

components/datadog/agentparams/filepermissions/permissions_win.go
Kaderinho: Alright 👍 Thanks for your help !

pgimalac (approved):
Not super familiar with the Windows commands but LGTM for the code itself

PR #781, opened 2024-04-18
Creating e2e test for udp logs

carlosroman:
What does this PR do?
---------------------

This updates the test logger app to log in various different ways.

-------------------

None as `ghcr.io/datadog/apps-logger:main` not used anywhere yet.

Motivation
----------

Want ability to send logs to Agent via `stdout`, `stderr`, `UDP` and `TCP`.

Additional Notes
----------------


components/datadog/apps/logger/images/logger/README.md
```diff
@@ -0,0 +1,41 @@
+#  Logger app
```
pducolin, line 1: **👏 praise**
Thank you for adding a doc file!

components/datadog/apps/logger/images/logger/main_test.go
```diff
@@ -0,0 +1,222 @@
+package main
```
pducolin, line 1: **❓ question**
Are you interested in running this on the CI? I can help! We have an integration test stage, we can add an unit test stage

Can also help migrating it to `testify`

pducolin (approved):
Super nice, with tests and docs 💯

components/datadog/apps/logger/images/logger/main_test.go
carlosroman: The app pretty simple and just there for some sanity checking. If it gets more complicated I'll definitely move to using `testify`.

> We have an integration test stage, we can add an unit test stage

PR #804, opened 2024-04-26
Increase the CPU and memory limits of the agents to avoid starvation

L3n41c:
What does this PR do?
---------------------

Increase the CPU and memory resources allocated to the agent.

-------------------

* `aws/eks`
* `aws/kind`

Motivation
----------

Despite #799, we still have some instabilities due to agent restarting because they are starved.
We need to increase the resources allocated to the agent a little bit more.

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/kubernetes_helm.go
+++ b/components/datadog/agent/kubernetes_helm.go
@@ -292,11 +292,11 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 				"agent": pulumi.Map{
 					"resources": pulumi.StringMapMap{
 						"requests": pulumi.StringMap{
-							"cpu":    pulumi.String("300m"),
+							"cpu":    pulumi.String("600m"),
 							"memory": pulumi.String("500Mi"),
 						},
 						"limits": pulumi.StringMap{
-							"cpu":    pulumi.String("500m"),
+							"cpu":    pulumi.String("1000m"),
 							"memory": pulumi.String("700Mi"),
 						},
 					},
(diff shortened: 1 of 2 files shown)


components/datadog/agent/kubernetes_helm.go
```diff
@@ -292,36 +292,36 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 				"agent": pulumi.Map{
 					"resources": pulumi.StringMapMap{
 						"requests": pulumi.StringMap{
-							"cpu":    pulumi.String("300m"),
+							"cpu":    pulumi.String("600m"),
 							"memory": pulumi.String("500Mi"),
```
L3n41c, line 296: @vboulineau : A memory request of 500 MiB looks generous.
But the real measured consumption isn’t a lot less than that: 
image
Also, on the Kind node, with the `t3.xlarge` type, our usage of memory is lower that our usage in CPU:
```
kubectl describe node/lenaic-kind-control-plane
[…]
  Resource           Requests      Limits
  --------           --------      ------
  cpu                2156m (53%)   3880m (97%)
  memory             3090Mi (19%)  3802Mi (24%)
```

vboulineau: What about the 0.6 CPU usage? Seems high for `kind` but perhaps SBOM scanning activated?

L3n41c: For the CPU, here is the past observed consumption:
image
image
I’m not sure to understand why we see those outliers above 400 millicores.
SBOM is indeed activate, but so it is for every single test run.
Given the fact that CPU is elastic, we can make a try with a request at 400 millicores ?

components/datadog/agent/kubernetes_helm.go
```diff
@@ -292,36 +292,36 @@ func buildLinuxHelmValues(installName, agentImagePath, agentImageTag, clusterAge
 				"agent": pulumi.Map{
 					"resources": pulumi.StringMapMap{
 						"requests": pulumi.StringMap{
-							"cpu":    pulumi.String("300m"),
+							"cpu":    pulumi.String("600m"),
```
L3n41c, line 295:
```suggestion
							"cpu":    pulumi.String("400m"),
```

@vboulineau 👍🏻 / 👎🏻 ?

vboulineau (approved):

PR #805, opened 2024-04-26
Remove the overrides to grant more resources to the fake intake

L3n41c:
What does this PR do?
---------------------

Removes the overrides to grant more resources to the fake intake.

-------------------

* `aws/eks`
* `aws/kind`

In the past, it has been necessary to bump the amount of memory granted to the fake intake because of the huge volume of data generated on the EKS and kind scenarios.
Since DataDog/datadog-agent#24064, the data are now kept on disk instead of memory.

Motivation
----------

Save money.

Additional Notes
----------------

image

### DIFF
--- a/scenarios/aws/eks/run.go
+++ b/scenarios/aws/eks/run.go
@@ -258,8 +258,7 @@ func Run(ctx *pulumi.Context) error {
 		var fakeIntake *fakeintakeComp.Fakeintake
 		if awsEnv.GetCommonEnvironment().AgentUseFakeintake() {
 			fakeIntakeOptions := []fakeintake.Option{
-				fakeintake.WithCPU(1024),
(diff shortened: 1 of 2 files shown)


vboulineau (approved):

L3n41c:
This graph shows the memory usage and the memory limit of the fake intake on a kind scenario:
image
A change in the color of the graph indicates a container restart.

We can see that the default 1GiB memory limit is too low and the fake intake restarts in loop.
This is also because `GOMEMLIMIT` is set 600MiB lower than the memory limit: https://github.com/DataDog/test-infra-definitions/blob/a62ab931e64875504b7d7a839d311d2f94fbb211/scenarios/aws/fakeintake/fakeintake.go#L60

PR #809, opened 2024-04-29
Remove triggers to not force a replace on manifest change

KevinFairise2:
What does this PR do?
---------------------

Modify the `docker-compose` command to include a hash of the content manifest as an environment variable.
This is a workaround to stop using `Triggers` that will create a replacement (`docker-compose down` + `docker-compose up`) to avoid some flakiness we have when this happen. We now let `docker-compose` handle the update.

-------------------

docker

Motivation
----------

Fix flakiness

Additional Notes
----------------

### DIFF
--- a/components/docker/docker.go
+++ b/components/docker/docker.go
@@ -85,9 +85,8 @@ func (d *Manager) ComposeStrUp(name string, composeManifests []ComposeInlineMani
 	if err != nil {
 		return nil, err
 	}
-
 	var remoteComposePaths []string
-	manifestHashes := map[string]pulumi.StringOutput{}
+	var manifestContents pulumi.StringArray
 	runCommandDeps := make([]pulumi.Resource, 0)
 	for _, manifest := range composeManifests {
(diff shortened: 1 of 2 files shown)


components/docker/docker.go
```diff
@@ -102,10 +101,18 @@ func (d *Manager) ComposeStrUp(name string, composeManifests []ComposeInlineMani
...
 
 		runCommandDeps = append(runCommandDeps, writeCommand)
-		runCommandTriggers = append(runCommandTriggers, manifest.Content)
 	}
+	contentHash := manifestContents.ToStringArrayOutput().ApplyT(func(inputs []string) string {
+		mergedContent := strings.Join(inputs, "\n")
+		return utils.StrHash(mergedContent)
+	}).(pulumi.StringOutput)
+
+	// We include a hash of the manifest contents in the environment variables to trigger an update when a manifest changes
```
KevinFairise2, line 113:
```suggestion
	// We include a hash of the manifests content in the environment variables to trigger an update when a manifest changes
```

vboulineau (approved):

pducolin:
`Environments` triggers an `Update`? So we did not need to have `envVars` in triggers?

PR #812, opened 2024-05-01
(fleet) cleanup the test infra definitions

arbll:
This PR cleans up the test infra definitions for the installer:
- Configuration dir/file is *not* required during initial installation. We want to test that property.
- No more systemd units to restart
- No packages to "bootstrap"

### DIFF
--- a/components/datadog/updater/install_script.sh
+++ b/components/datadog/updater/install_script.sh
@@ -10,14 +10,7 @@ if [ "$UID" == "0" ]; then
 else
     sudo_cmd="sudo"
 fi
-config_file="/etc/datadog-agent/datadog.yaml"
-$sudo_cmd mkdir -p /etc/datadog-agent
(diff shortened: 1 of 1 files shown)


components/datadog/updater/install_script.sh
```diff
@@ -10,14 +10,7 @@ if [ "$UID" == "0" ]; then
 else
     sudo_cmd="sudo"
 fi
-config_file="/etc/datadog-agent/datadog.yaml"
```
raphaelgavache, line 13: you need to keep the config file to avoid breaking tests with Agent unit assertions in datadog-agent installer E2E suite

raphaelgavache (approved):

components/datadog/updater/install_script.sh
raphaelgavache: removing it will need a first migration step in datadog-agent

PR #817, opened 2024-05-03
[agent] allow nested, exportable components

pducolin:
What does this PR do?
---------------------

* allow exporting nested components
* add docker manager component
* export host in docker manager and agent component
* export docker manager in docker agent component

-------------------

vm, docker, kind

Motivation
----------

We want to have fully autonomous components, that exporting anything they need to setup a client. This changes are then handled by [this PR](https://github.com/DataDog/datadog-agent/pull/25370)

Additional Notes
----------------


components/datadog/agent/docker.go
```diff
@@ -37,7 +39,7 @@ func (h *DockerAgent) Export(ctx *pulumi.Context, out *DockerAgentOutput) error
 }
 
 func NewDockerAgent(e config.CommonEnvironment, vm *remoteComp.Host, manager *docker.Manager, options ...dockeragentparams.Option) (*DockerAgent, error) {
-	return components.NewComponent(e, vm.Name(), func(comp *DockerAgent) error {
+	return components.NewComponent(e, vm.Name()+"-dockerAgent", func(comp *DockerAgent) error {
```
vboulineau, line 42: `DockerAgent` is already in the type name, which is both visible in Pulumi output and in exported key.

components/docker/component.go
```diff
@@ -20,33 +21,46 @@ const (
...
 }
 
-func NewManager(e config.CommonEnvironment, host *remoteComp.Host, opts ...pulumi.ResourceOption) (*Manager, pulumi.Resource, error) {
-	manager := &Manager{
-		namer: e.CommonNamer.WithPrefix("docker"),
-		host:  host,
-		opts:  opts,
-	}
+func NewManager(e config.CommonEnvironment, host *remoteComp.Host, opts ...pulumi.ResourceOption) (*Manager, error) {
+	return components.NewComponent(e, host.Name()+"-dockerManager", func(comp *Manager) error {
```
vboulineau, line 41: Same as above here

components/docker/component.go
```diff
@@ -20,33 +21,46 @@ const (
...
+		comp.namer = e.CommonNamer.WithPrefix("docker")
+		comp.Host = host
+		comp.opts = opts
```
vboulineau, line 44: `pulumi.Parent(comp)` needs to be added here if you want to be able to `DependOn` the component itself later.

components/docker/component.go
```diff
@@ -20,33 +21,46 @@ const (
 	defaultTimeout = 300
 )
 
+type ManagerOutput struct {
+	components.JSONImporter
+
+	Host remoteComp.HostOutput `json:"host"`
+}
+
 type Manager struct {
```
vboulineau, line 30: Wonder if the name should be `docker.Host` instead, with perhaps anonymous embed of `remote.Host`?

components/datadog/agent/docker.go
pducolin: Removing it

components/docker/component.go
pducolin: Good point, logically it's rather part of the host where the docker agent stands on. I have more renaming ideas, I would call all the `*Output` types `*Config`, and move all test-infra components names to `*ConfigOutput` as in `pulumi.Output`. But rather for later.

components/datadog/agent/docker.go
pducolin: Wait, it's not visible in the `Host` name though, I don't understand how the name works then. Should I extract the name from the component type?

pducolin: Updated `NewComponent` doc to explicit how `name` is used with the reflected component name

components/docker/component.go
pducolin: Not sure here, should I add `pulumi.Parent(comp)` to `comp.opts`? Wouldn't that mean that `comp` is the parent of `comp`?

tasks/config.py
```diff
@@ -36,7 +36,7 @@ class Agent(BaseModel, extra=Extra.forbid):
         class Pulumi(BaseModel, extra=Extra.forbid):
             logLevel: Optional[int] = None
             logToStdErr: Optional[bool] = None
-            verboseProgressStreams: Optional[bool] = None  # noqa
+            verboseProgressStreams: Optional[bool] = None
```
KevinFairise2, line 39: I added it there to not have failure when using a `~/.test_infra_config.yaml` with that key. The issue with that key is that there is no equivalent between what we use this flag for on `datadog-agent` with Pulumi API Stack manager and the flags that are available with `pulumi` cli. 
With the Pulumi API, when this flag is set to `false` we will only display resource deployment logs for resource that failed to be deployed.
With `pulumi up` CLI here it would only suppress progress points.

pducolin: Yes, I forced it into dropping the points progress, but I can revert and just skip the usage of this key. Changing it

components/docker/component.go
```diff
@@ -141,7 +141,8 @@ func (d *Manager) ComposeStrUp(name string, composeManifests []ComposeInlineMani
 }
 
 func (d *Manager) install() (*remote.Command, error) {
-	dockerInstall, err := d.Host.OS.PackageManager().Ensure("docker.io", nil, "docker", d.opts...)
+	opts := append(d.opts, pulumi.Parent(d))
```
vboulineau, line 144: This may modify `d.opts` depending on whether there is available capacity in `d.opts` or not, that's why it's important to always use `utils.MergeOptions`

components/docker/component.go
vboulineau: That's a valid point. I think the diff mostly comes to the outside user (which does not need to know how it's built), IMHO having a `docker.Host` that builds upon on `Host` (basically transforming a `Host` into a `DockerHost`) does make sense as an API but no strong opinion or changes needed.

components/docker/component.go
pducolin: TIL: `append` modifies the passed slice even if we assign the result to a new slice

PR #824, opened 2024-05-10
Revert "Make restart-service more robust to avoid weird test flakiness"

pducolin:
Reverts DataDog/test-infra-definitions#771

Since https://github.com/DataDog/datadog-agent/pull/25282 this workaround should not be needed anymore


FlorentClarret (approved):
Nice!

iglendd (approved):

PR #830, opened 2024-05-16
Add the ability to transfer files to system folders

FlorentClarret:
What does this PR do?
---------------------

Add the ability to transfer files to system folders copying it to `/tmp` and then copying it to the right destination using `sudo`. 

Note that this is a bit less efficient, it's not great but it works

-------------------

All linux ones that copy files.

Motivation
----------

- Address https://datadoghq.atlassian.net/wiki/spaces/AW/pages/3659661423/Known+E2E+issues+idiosyncrasy#Permission-denied-during-SFTP-operations

Additional Notes
----------------

- Part 1 of https://datadoghq.atlassian.net/browse/ADXT-228
- In a follow-up PR I'll address the encoding issue by creating a local file and using scp to send it instead of using `stdin`

### DIFF
--- a/components/command/unixOSCommand.go
+++ b/components/command/unixOSCommand.go
@@ -2,6 +2,8 @@ package command
 
 import (
 	"fmt"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+	"path/filepath"
 	"strings"
 
 	"github.com/alessio/shellescape"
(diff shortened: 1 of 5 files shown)


components/command/unixOSCommand.go
```diff
@@ -84,6 +86,36 @@ func (fs unixOSCommand) IsPathAbsolute(path string) bool {
...
+		LocalPath:  pulumi.String(localPath),
+		RemotePath: pulumi.String(tempRemotePath),
+		Triggers:   pulumi.Array{pulumi.String(localPath), pulumi.String(tempRemotePath)},
+	}, utils.MergeOptions(runner.options, opts...)...)
+
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = runner.osCommand.CopyRemoteFile(runner, tempRemotePath, remotePath, true, pulumi.DependsOn([]pulumi.Resource{tempCopyFile}))
```
KevinFairise2, line 111: **🥜 nitpick:** ‏
```suggestion
	_, err = runner.osCommand.CopyRemoteFile(runner, tempRemotePath, remotePath, true, utils.PulumiDependsOn(tempCopyFile))
```

components/command/unixOSCommand.go
```diff
@@ -84,6 +86,36 @@ func (fs unixOSCommand) IsPathAbsolute(path string) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return tempCopyFile, err
+}
```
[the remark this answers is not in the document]
KevinFairise2: I do not think you can do that. Pulumi does not really support `Resources` that are created or not depending of the result of the creation of a previous command

components/command/windowsOSCommand.go
```diff
@@ -12,6 +14,15 @@ var _ OSCommand = (*windowsOSCommand)(nil)
 
 type windowsOSCommand struct{}
 
+func (fs windowsOSCommand) NewCopyFile(runner *Runner, localPath, remotePath string, opts ...pulumi.ResourceOption) (*remote.CopyFile, error) {
```
KevinFairise2, line 17: Is it expected that this function does not have the same behavior as on Unix? I think this one will not copy the file to a world-writable folder first and then copy it. Hence the `CopyRemoteFile` is not used for now

components/command/unixOSCommand.go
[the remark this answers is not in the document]
FlorentClarret: Yup that's what I figured

components/command/windowsOSCommand.go
FlorentClarret: Yes! For Windows we use `Administrator`, that has the permission to copy files on system folders. The problem only exists on Linux because we do not use `root`. I can update the description to make it explicit

KevinFairise2: Maybe we could remove the `CopyRemoteFile` method then? Or should it be available to be used elsewhere?

FlorentClarret: I can drop it for now I guess. At first I added it to the interface so I wrote both implementations. I can convert it to to a private function in the `unix` command. On it!

KevinFairise2 (approved):

PR #833, opened 2024-05-17
Use CopyFile in the CopyInlineFile to avoid encoding issues

FlorentClarret:
What does this PR do?
---------------------

Use CopyFile in the CopyInlineFile to avoid encoding issues

-------------------

All the scenarios since they all use the `CopyInlineFile` to at least create the Agent config remotely.

Motivation
----------

We currently use commands in order to create a file from a string. This approach causes some issues with the encoding sometimes. To avoid them, we first create a temporary file locally that we then copy using Pulumi to avoid altering the content. 

CopyFile can now [transfer files to system folders](https://github.com/DataDog/test-infra-definitions/pull/830) so this should not be a problem.

Additional Notes
----------------

In the CopyInlineFile, I first get the content of the StringInput through the ApplyT function that runs a goroutine. We need to wait for it to be sure we get the path to the file, and it is well created. I used a WaitingGroup for that.

I'm not a Go/Pulumi expert at all. I'm not sure this the best approach but it seems to be working fine. If you have a better approach, please let me know 🙇

### DIFF
--- a/components/command/filemanager.go
+++ b/components/command/filemanager.go
@@ -3,15 +3,15 @@ package command
 import (
 	"errors"
 	"fmt"
+	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
-
-	"github.com/pulumi/pulumi-command/sdk/go/command/remote"
-	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"sync"
 )
 
 type FileManager struct {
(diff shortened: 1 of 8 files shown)


components/command/filemanager.go
```diff
@@ -69,8 +69,34 @@ func (fm *FileManager) CopyFile(localPath, remotePath string, opts ...pulumi.Res
...
+		tempFilePath = tempFile.Name()
+		_, err = tempFile.WriteString(content)
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+
+	waitingGroup.Wait()
```
pducolin, line 97: **💬 suggestion**
I suspect this blocks in preview mode. The proper way would be to return the count of written bytes from `tempFile.WriteString(content)` as a `IntOutput`, which you can depend on when calling `fm.CopyFile`

FlorentClarret: Yes this blocks to be sure the file is properly written when we declare the `CopyFile` right after, otherwise we could get a "file not found" error. 

I don't know how to depend on an `IntOutput`, I'll have a look. Looks like I can't just `utils.MergeOptions(opts, pulumi.DependsOn([]pulumi.Resource{output}))`

FlorentClarret: I broke the CI so yeah there is definitely something wrong

pducolin (approved):

PR #836, opened 2024-05-23
Add agent-with-operator install method

fanny-jiang:
What does this PR do?
---------------------

Add new agent install method via the operator

-------------------
* kind

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -7,9 +7,10 @@ import (
 	"os/user"
 	"strings"
 
-	"github.com/DataDog/test-infra-definitions/common/namer"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	sdkconfig "github.com/pulumi/pulumi/sdk/v3/go/pulumi/config"
+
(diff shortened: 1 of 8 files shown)


vboulineau:
Awesome to see Operator support coming. Some thoughts about the structure:
- The Operator should be standalone as it can be used for many things, i.e. dedicated folder in `components/datadog/operator`. In which you could have a `Helm` way and perhaps later an OPM way.
- The component system focuses on producing component that can later used, abstracting how the component was built. Currently the `KubernetesAgent` component exposes Helm info, but it should not, it should only expose info to work with deployed Agent pods (which we don't do today, and probably should not do either, so not a big deal). It's the same with the `OperatorHelmComponent`, it should be the `Operator` component, regardless of it was created from an Helm install or an OPM install in the future.
- You can then create a `agent/kubernetes_operator.go` that takes in a `KubernetesCluster`, an `Operator` and returns a `KubernetesAgent`. The way to configure the deployment should be to take the `kubernetesagentparams` similar to Helm install (probably some refactoring required to share what can be shared and then extend for Helm and Operator)

PR #865, opened 2024-06-07
[ADXT-333] Clean up and align invoke create/destroy commands

KevinFairise2:
What does this PR do?
---------------------

Fix some destroy tasks command that were not working anymore after we update `destroy` signature. Forcing keywords argument to avoid breaking it again in the future

-------------------

Motivation
----------

Additional Notes
----------------


FlorentClarret (approved):

tasks/destroy.py
```diff
@@ -11,6 +11,7 @@
 
 def destroy(
     ctx: Context,
+    *,
```
pducolin, line 14: **💬 suggestion**
TIL, should it be at the end instead?

pducolin (approved):

tasks/destroy.py
FlorentClarret: No, this is how we enforce callers to use named params and not positional ones 🙂

PR #870, opened 2024-06-10
[ASCII-1506] Use backslashes in paths for `icacls` command

pgimalac:
What does this PR do?
---------------------
Ensure paths given to `SetupPermissionsCommand` use backslashes, and remove the recursive (`/t`) flag of `icacls`.

-------------------
Windows E2e test using this helper to set file permissions.

Motivation
----------
When using forward slash `icacls` was behaving very wrongly, reading the wrong path, and entering an infinite loop.
The function changes permissions of a single path so no need to set the recursive flag.

Additional Notes
----------------
Ran an e2e test successfully with this change.


components/datadog/agentparams/filepermissions/permissions_win.go
```diff
@@ -44,7 +46,7 @@ func (p *WindowsPermissions) SetupPermissionsCommand(path string) string {
 
 // ResetPermissionsCommand returns a command that resets the owner, group, and permissions of a file to default.
 func (p *WindowsPermissions) ResetPermissionsCommand(path string) string {
-	return fmt.Sprintf("icacls “%v” /reset /t /c /l;", path)
```
clarkb7, line 47: I don't think you should remove the `/t` here, in case there's a directory that you do want to reset the whole tree for. I think rather there should be another option/field to designate that `/t` should or should not be included.

Including it when path is a file has an interesting behavior that may not be desirable, too, so always including it doesn't seem good either.
```
> icacls.exe C:\Users\Administrator\Desktop\test.txt /reset /t
processed file: C:\Users\Administrator\Desktop\test.txt
processed file: C:\Users\Administrator\Desktop\test\test.txt
Successfully processed 2 files; Failed processing 0 files
```

pducolin (approved):

components/datadog/agentparams/filepermissions/permissions_win.go
pgimalac: Currently these permission helpers are only used with files (with the `WithFileWithPermissions` param), they can't be used with directories

If at some point we want to add helpers for directory permissions then we can revisit this, but for now I think it's just simpler and less-error prone not to have the recursive flag (your example is pretty surprising, and as we discussed I also hit a pretty weird behavior when using the recursive flag 😅)

By the way the Linux version is also not recursive, so it just makes it more similar

clarkb7: sounds good!

PR #885, opened 2024-06-13
Add RDP helper for VMs

clarkb7:
What does this PR do?
---------------------
Add two new tasks
- `get-vm-password`: fetch password for hosts in stack
- `rdp-vm`: copy password to clipboard and run RDP tool to connect to hosts in stack

example usage
```bash
$ pulumi stack ls -a
NAME                                                        LAST UPDATE     RESOURCE COUNT
organization/e2elocal/brandenclark-windows-msi-test         37 seconds ago  8

$ inv rdp-vm --stack-name organization/e2elocal/brandenclark-windows-msi-test
```

-------------------
N/A

Motivation
----------
To make it easier to RDP into VMs

Additional Notes
----------------
Can someone test the macos rdp command and/or provide a Linux rdp command?

EC2 requires the VM be provisioned with an RSA key in order to decrypt the Windows password. `inv setup.debug` will notify you of this. Perhaps we could consider having the framework set the password to avoid this requirement.


KevinFairise2 (approved):

tasks/setup.py
```diff
@@ -498,6 +498,9 @@ def debug_keys(ctx: Context, config_path: Optional[str] = None):
         # https://github.com/pulumi/pulumi-command/blob/58dda0317f72920537b3a0c9613ce5fed0610533/provider/pkg/provider/remote/connection.go#L105-L118
         error("Private key is not provided in the config. Pulumi does not support Windows SSH agent.")
         info("Configure privateKeyPath and provide the privateKeyPassword if the key is encrypted.")
+    if not awsConf.privateKeyPath:
```
pducolin, line 501: **💬 suggestion**
If not in there yet, could be part of the `setup` task

tasks/tool.py
```diff
@@ -165,6 +176,25 @@ def is_linux():
...
+def get_aws_instance_password_data(
+    ctx: Context, vm_id: str, key_path: str, aws_account: Optional[str] = None, use_aws_vault: Optional[bool] = True
+) -> str:
+    buffer = StringIO()
+    with ctx.cd(_get_root_path()):
+        cmd = f"aws ec2 get-password-data --instance-id {vm_id} --priv-launch-key {key_path}"
+        if use_aws_vault:
+            if aws_account is None:
+                raise Exit("AWS account is required when using aws-vault.")
+            cmd = get_aws_wrapper(aws_account) + cmd
```
pducolin, line 192: **💬 suggestion**
Should not be required since https://github.com/DataDog/test-infra-definitions/pull/917, as you call `aws` directly you might prepend with `AWS_PROFILE` and the default profile name for the current target from config, if `AWS_PROFILE` is not set yet

Something like
```
prefix = ""
if os.env("AWS_PROFILE") is None:
  prefix = `"AWS_PROFILE=exec-sso-agent-[sandbox|qa]-account-admin "`
```

The advantage is that you don't need to pass the `use_aws_vault` option

pducolin (approved):
Amazing! Is `RDP` port already open in `agent-sandbox`?

tasks/doc.py
```diff
@@ -23,6 +23,7 @@
 ami_id: str = "A full Amazon Machine Image (AMI) id (e.g. ami-0123456789abcdef0)"
 architecture: str = f"The architecture to use. Possible values are {tool.get_architectures()}. Default '{tool.get_default_architecture()}'"
 yes: str = "Automatically approve and perform the destroy without previewing it"
+use_aws_vault: str = "Wrap aws command with aws-vault, default to True"
```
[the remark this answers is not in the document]
FlorentClarret: yes, AWS_PROFILE can be used https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html#cli-configure-files-using-profiles

tasks/tool.py
clarkb7: discussed on slackL

will merge this PR as is, since the profile isn't used by python yet, and there are other places that still use `get_aws_wrapper`. They can all be changed together in a separate PR.

PR #894, opened 2024-06-20
Do not break sandbox where we do not have load balancers

KevinFairise2:
What does this PR do?
---------------------

We have some user still using `sandbox` to create VMs.
We do not maintain load balancer for fakeintake on sandbox so let's not call the code to randomly select one

-------------------

Motivation
----------

Additional Notes
----------------


amenasria (approved):
LGTM thanks !

FlorentClarret (approved):

PR #896, opened 2024-06-20
upgrade(VMs): Add AmazonLinux2018 types

BaptisteFoy:
What does this PR do?
---------------------
Adds AmazonLinux2018 VM. That VM runs on SysVInit and not systemd, which is useful to test in E2Es as we are supposed to test it.

-------------------
Installer tests for now, maybe more if users chose to use that VM :)

Motivation
----------
Testing installer + injector on SysVInit scenarii in E2Es

Additional Notes
----------------
Image: https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#ImageDetails:imageId=ami-07541a4f680f1ba8e

### DIFF
--- a/components/os/linux.go
+++ b/components/os/linux.go
@@ -18,24 +18,26 @@ func newLinuxOS(e config.Env, desc Descriptor, runner *command.Runner) OS {
 	case AmazonLinux, AmazonLinuxECS, CentOS:
 		// AL2 is YUM, AL2023 is DNF (but with yum compatibility)
 		os.packageManager = newYumManager(runner)
-		os.serviceManager = newSystemdServiceManager(e, runner)
 
 	case Fedora, RedHat, RockyLinux:
 		os.packageManager = newDnfManager(runner)
-		os.serviceManager = newSystemdServiceManager(e, runner)
 
 	case Debian, Ubuntu:
 		os.packageManager = newAptManager(runner)
-		os.serviceManager = newSystemdServiceManager(e, runner)
 
 	case Suse:
 		os.packageManager = newZypperManager(runner)
-		os.serviceManager = newSystemdServiceManager(e, runner)
 
 	case Unknown, WindowsServer, MacosOS:
 		fallthrough
 	default:
 		panic(fmt.Sprintf("unsupported linux flavor from desc: %+v", desc))
+	}
+
+	if desc.Flavor == AmazonLinux2018.Flavor && desc.Version == AmazonLinux2018.Version {
+		os.serviceManager = newSysvinitServiceManager(e, runner)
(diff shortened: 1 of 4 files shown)


components/os/linux.go
```diff
@@ -18,25 +18,27 @@ func newLinuxOS(e config.Env, desc Descriptor, runner *command.Runner) OS {
 	case AmazonLinux, AmazonLinuxECS, CentOS:
 		// AL2 is YUM, AL2023 is DNF (but with yum compatibility)
 		os.packageManager = newYumManager(runner)
-		os.serviceManager = newSystemdServiceManager(e, runner)
```
pducolin, line 21: **💬 suggestion**
I think you need to add the AmazonLinux2018 case or it would default to `panic`

```suggestion

	case AmazonLinux2018:
		os.PackageManager = new YumManager(runner)
		if desc.Version == AmazonLinux2018.Version {
			os.serviceManager = newSysvinitServiceManager(e, runner)
		} else {
			os.serviceManager = newSystemdServiceManager(e, runner)
		}
```

scenarios/aws/ec2/os_resolver.go
```diff
@@ -92,6 +96,11 @@ func resolveAmazonLinuxAMI(e aws.Environment, osInfo *os.Descriptor) (string, er
 		paramName = fmt.Sprintf("amzn2-ami-hvm-%s-gp2", osInfo.Architecture)
 	case os.AmazonLinuxECS2023.Version:
 		paramName = fmt.Sprintf("al2023-ami-kernel-default-%s", osInfo.Architecture)
+	case os.AmazonLinux2018.Version:
+		if osInfo.Architecture == os.ARM64Arch {
+			return "", errors.New("ARM64 is not supported for Amazon Linux 2018")
+		}
+		return "ami-051394ccb12210d6e", nil // Image name: amzn-ami-2018.03.20240131-amazon-ecs-optimized
```
pducolin, line 103: **❓ question**
Did you copy this ami and add it to the custom AMIs? Referencing a public AMI comes with the risk of AWS removing it suddenly

components/os/linux.go
BaptisteFoy: The switch case is testing the flavors, and AmazonLinux2018 is an AmazonLinux flavor. FWIW I tested this whole PR locally and iterated until it worked

scenarios/aws/ec2/os_resolver.go
BaptisteFoy: I didn't. How can I make sure the AMI ID is the same between multiple AWS accounts (agent-sandbox / agent-qa)

PR #897, opened 2024-06-20
Allow to select os version for create-vm invoke task

KevinFairise2:
What does this PR do?
---------------------

Add a `--os-version` flag to select the OS version

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/vm.py
+++ b/tasks/vm.py
@@ -64,7 +64,7 @@ def create_vm(
     extra_flags = {}
     os_family, os_arch = _get_os_information(ctx, os_family, architecture, ami_id)
     deploy_job = None if no_verify else tool.get_deploy_job(os_family, os_arch, agent_version)
-    extra_flags["ddinfra:osDescriptor"] = f"{os_family}:{os_version}:{os_arch}"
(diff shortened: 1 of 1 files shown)


FlorentClarret (approved):

tasks/vm.py
```diff
@@ -62,7 +64,7 @@ def create_vm(
     extra_flags = {}
     os_family, os_arch = _get_os_information(ctx, os_family, architecture, ami_id)
     deploy_job = None if no_verify else tool.get_deploy_job(os_family, os_arch, agent_version)
-    extra_flags["ddinfra:osDescriptor"] = f"{os_family}::{os_arch}"
+    extra_flags["ddinfra:osDescriptor"] = f"{os_family}:{os_version}:{os_arch}"
```
KSerrania, line 67: User documentation question: how does someone know what `os_version` values are available?

KevinFairise2: That's a good point. Actually the available versions for a given os are the one mentioned in that file https://github.com/DataDog/test-infra-definitions/blob/main/components/os/linux_descriptors.go
I will put a link to that file in the help. But given the list is maintained in Go code it is hard to link both dynamically

PR #917, opened 2024-07-01
Set the Profile in the AwsProvider to avoid wrapping the commands into `aws-vault` commands

FlorentClarret:
What does this PR do?
---------------------

Set the Profile in the AwsProvider to avoid wrapping the commands into `aws-vault` commands. We use:

- First the `AWS_PROFILE` env variable if it is set (this is the current behaviour in the provider internally)
- Read the value from the config
- Fallback to the default value set in the `environnementDefault` otherwise

This way, there's no need to wrap the command into `aws-vault`, nor here or in the Agent to run e2e tests.

-------------------

Everyone

Motivation
----------

- The main benefit is that all command would work without wrapping and for EKS it allows to generate a Kubeconfig with profile inside which means kubectl commands works OOTB without wrappers.
- It's more convenient to avoid wrapping the commands into `aws-vault`

Additional Notes
----------------

https://datadoghq.atlassian.net/browse/ADXT-356


tasks/ecs.py
```diff
@@ -71,13 +71,11 @@ def _show_connection_message(ctx: Context, config_path: Optional[str], full_stac
...
     try:
-        local_config = config.get_local_config(config_path)
+        config.get_local_config(config_path)
     except ValidationError as e:
         raise Exit(f"Error in config {config.get_full_profile_path(config_path)}:{e}")
 
-    command = (
-        f"{tool.get_aws_wrapper(local_config.get_aws().get_account())} aws ecs list-tasks --cluster {cluster_name}"
-    )
+    command = f"aws ecs list-tasks --cluster {cluster_name}"
```
FlorentClarret, line 78: revert

tasks/eks.py
```diff
@@ -93,7 +93,7 @@ def _show_connection_message(ctx: Context, full_stack_name: str, config_path: Op
     except ValidationError as e:
         raise Exit(f"Error in config {config.get_full_profile_path(config_path)}:{e}")
 
-    command = f"KUBECONFIG={kubeconfig} {tool.get_aws_wrapper(local_config.get_aws().get_account())} kubectl get nodes"
+    command = f"KUBECONFIG={kubeconfig} {local_config.get_aws().get_account()} kubectl get nodes"
```
FlorentClarret, line 96: revert

tasks/tool.py
```diff
@@ -152,12 +152,6 @@ def get_stack_json_outputs(ctx: Context, full_stack_name: str) -> Any:
     return json.loads(buffer.getvalue())
 
 
-def get_aws_wrapper(
-    aws_account: str,
-) -> str:
-    return f"aws-vault exec sso-{aws_account}-account-admin -- "
```
FlorentClarret, line 158: keep it

vboulineau (approved):

resources/aws/environment.go
```diff
@@ -96,7 +98,8 @@ func NewEnvironment(ctx *pulumi.Context, options ...func(*Environment)) (Environ
 	env.envDefault = getEnvironmentDefault(config.FindEnvironmentName(env.InfraEnvironmentNames(), awsConfigNamespace))
 
 	awsProvider, err := sdkaws.NewProvider(ctx, string(config.ProviderAWS), &sdkaws.ProviderArgs{
-		Region: pulumi.String(env.Region()),
+		Region:  pulumi.String(env.Region()),
+		Profile: pulumi.String(env.Profile()),
```
vboulineau, line 102: Note, perhaps for later: It may be useful to detect command is already run with credentials and `AWS_PROFILE=<empty>` (probably something like `AWS_SECRET_ACCESS_KEY/AWS_SESSION_TOKEN` being set, but need to be verified further) to avoid breaking already authed context.

integration-tests/invoke_test.go
```diff
@@ -83,12 +83,12 @@ func testInvokeKind(t *testing.T, tmpConfigFile string) {
...
 	stackName := strings.Join(stackParts, "-")
 	t.Log("creating kind cluster")
-	createCmd := exec.Command("invoke", "create-kind", "--no-interactive", "--stack-name", stackName, "--no-use-aws-vault", "--config-path", tmpConfigFile, "--use-fakeintake", "--use-loadBalancer")
+	createCmd := exec.Command("invoke", "create-kind", "--no-interactive", "--stack-name", stackName, "--config-path", tmpConfigFile)
 	createOutput, err := createCmd.Output()
 	assert.NoError(t, err, "Error found creating kind cluster: %s", string(createOutput))
 
 	t.Log("destroying kind cluster")
-	destroyCmd := exec.Command("invoke", "destroy-kind", "--yes", "--stack-name", stackName, "--no-use-aws-vault", "--config-path", tmpConfigFile)
+	destroyCmd := exec.Command("invoke", "destroy-kind", "--yes", "--stack-name", stackName, "--config-path", tmpConfigFile)
```
FlorentClarret, line 91: revert

resources/aws/environment.go
FlorentClarret: Yes, that's what I have with an active session: 

```
AWS_VAULT=sso-agent-sandbox-account-admin
AWS_REGION=us-east-1
AWS_DEFAULT_REGION=us-east-1
AWS_ACCESS_KEY_ID=something
AWS_SECRET_ACCESS_KEY=something
AWS_SESSION_TOKEN=something
AWS_CREDENTIAL_EXPIRATION=2024-07-01T14:57:36Z
```

tasks/deploy.py
```diff
@@ -65,7 +64,6 @@ def deploy(
     # Verify image deployed and not outdated in s3
     if deploy_job is not None and pipeline_id is not None:
         cmd = f"inv -e check-s3-image-exists --pipeline-id={pipeline_id} --deploy-job={deploy_job}"
-        cmd = tool.get_aws_wrapper(aws_account) + cmd
```
FlorentClarret, line 68: revert

pducolin:
Amazing!

PR #921, opened 2024-07-02
[centos] add centos 7 arm64 AMI

pducolin:
What does this PR do?
---------------------

Add CentOS 7 arm64 AMI

-------------------

VM

Motivation
----------

We do have one AMI for arm64 CentOS 64, let's use it!

Additional Notes
----------------


FlorentClarret (approved):
🎉

PR #923, opened 2024-07-02
increase integration-tests job resources to run it in parallel

pducolin:
What does this PR do?
---------------------

This PR refactors the integration test to allow running it locally and in parallel. It increases the CI resources to avoid OOM during the test. 

-------------------

None

Motivation
----------

This allows to reduce the `test-infra-definitions` pipeline time by half, from ~40 min to ~20 min

Additional Notes
----------------


KevinFairise2 (approved):
Nice 🎉 🚤

PR #937, opened 2024-07-10
[ADXT-367] [e2e] Add explicit failure when the image is no longer available in the registry

KevinFairise2:
---------------------

Before trying to install the agent we check if the image exists on the ECR, if not we fail early with an explicit error

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/kubernetesagentparams/params.go
+++ b/components/datadog/kubernetesagentparams/params.go
@@ -58,9 +58,23 @@ func NewParams(e config.Env, options ...Option) (*Params, error) {
 	version := &Params{
 		Namespace: defaultAgentNamespace,
 	}
-
 	if e.PipelineID() != "" && e.CommitSHA() != "" {
+		exists, err := e.InternalRegistryImageTagExists(fmt.Sprintf("%s/agent", e.InternalRegistry()), fmt.Sprintf("%s-%s", e.PipelineID(), e.CommitSHA()))
+		if err != nil {
+			return nil, err
+		}
+		if !exists {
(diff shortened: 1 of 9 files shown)


components/datadog/kubernetesagentparams/params.go
```diff
@@ -58,9 +58,23 @@ func NewParams(e config.Env, options ...Option) (*Params, error) {
...
+			return nil, fmt.Errorf("agent image %s not found in the internal registry", fmt.Sprintf("%s-%s", e.PipelineID(), e.CommitSHA()))
+		}
 		options = append(options, WithAgentFullImagePath(utils.BuildDockerImagePath(fmt.Sprintf("%s/agent", e.InternalRegistry()), fmt.Sprintf("%s-%s", e.PipelineID(), e.CommitSHA()))))
+
+		exists, err = e.InternalRegistryImageTagExists(fmt.Sprintf("%s/cluster-agent", e.InternalRegistry()), fmt.Sprintf("%s-%s", e.PipelineID(), e.CommitSHA()))
+		if err != nil {
+			return nil, err
+		}
+		if !exists {
+			return nil, fmt.Errorf("cluster agent image %s not found in the internal registry", fmt.Sprintf("%s-%s", e.PipelineID(), e.CommitSHA()))
```
FlorentClarret, line 76: Could we also add the `e.InternalRegistry()` similarly to what you did in `params.go`?

FlorentClarret (commented):
Looks great! I just left one tiny suggestion to improve the logs for debugging purpose

components/datadog/kubernetesagentparams/params.go
KevinFairise2: Sure, I will add that

FlorentClarret (approved):

PR #916, opened 2024-07-01
rollback pulumi-aws to v6.25.0

pducolin:
What does this PR do?
---------------------

Rollback `pulumi-aws` dependency

Which scenarios this will impact?
-------------------

All

Motivation
----------

The bump is breaking the `fakeintake` deployment on [e2e tests](https://github.com/DataDog/datadog-agent/pull/27159) 

Additional Notes
----------------

Added `fakeintake` deployment to integration tests. They will run longer, but will early catch such issues

PR #942, opened 2024-07-11
[gitlab] pre-build pulumi go program

pducolin:
What does this PR do?
---------------------

* Build `main.go` pulumi program before running integration tests
* Prevent breaking the current workflow by allowing building at `pulumi up` time locally

-------------------

None

Motivation
----------

Running pulumi on the CI requires big runners, both in e2e tests in `datadog-agent` and here when calling invoke tasks to create remote instances, in `integration_tests`. 

@AliDatadog  few months ago noticed how the high mem consumption in e2e tests is requested at compile time. This happens both when running e2e tests in `datadog-agent` and when running integration_tests in test-infra-definitions. 

[I recently added](https://github.com/DataDog/test-infra-definitions/pull/916) `fakeintake` to `integration_tests` as a `pulumi-aws` bump broke its usage in `datadog-agent` e2e tests. This further increased the pipeline time to ~40 minutes, as tests ran sequentially. In an attempt to reduce the CI time here [I ran tests in parallel](https://github.com/DataDog/test-infra-definitions/pull/923). 

The tests run the `inv create-[vm|docker|kind]` tasks: when running concurrently they each build `main.go`, which requires a huge amount of RAM. I increased the `KUBERNETES_MEMORY_LIMIT` to the maximum allowed of `32GB`. 

This got OOM in a [recent PR](https://github.com/DataDog/test-infra-definitions/pull/915) where we add more dependencies.

@KSerrania , @KevinFairise2 and myself investigated what we could do to allow running tests without starving the CI. Building once in a `16GB` runner and then test in // in a `6GB` runner is my suggested solution. 

Additional Notes
----------------

This requires a change to support the local workflow in `test-infra-definitions` as we need a `Pulumi.yaml` that builds the go program at `pulumi up` time to preserve the current local experience, while we need a `Pulumi.yaml` that uses the generated `main` binary on the CI.

One alternative is reducing // to 2 tests, and further reduce to execute the tests sequentially, but this would slow down the pipeline.


vboulineau:
Should we simplify and just keep `Pulumi.yaml` as-is and simply `mv Pulumi.ci.yaml -> Pulumi.yaml` as the CI is an environment we have precise control on?
Another option is to have an `invoke` task that does the `go build` + `Pulumi.yaml` (CI version) creation in a `dist/` folder or something. That would allow to simply switch between the two with a `pulumi -C <dir> up`

pducolin:
> Should we simplify and just keep Pulumi.yaml as-is and simply mv Pulumi.ci.yaml -> Pulumi.yaml as the CI is an environment we have precise control on?
Another option is to have an invoke task that does the go build + Pulumi.yaml (CI version) creation in a dist/ folder or something. That would allow to simply switch between the two with a pulumi -C <dir> up

Thank you for your suggestions. All of these are valid alternatives. 

The cons I see for the first one is that we modify a file that is monitored by git at CI time, which is not ideal. 

The second option would break the experience for folks not using the invoke tasks: if you want to use `pulumi up` you need to remember to manually rebuild the go program. I prefer preventing errors, I would personally forget to build the program. 

The current solution requires a once-in-a-dev-time call to `inv setup` or of manually copying the `Pulumi.yaml.local` to `Pulumi.yaml` for those who don't want to use invoke tasks.

What do you think?

pducolin:
Implemented https://github.com/DataDog/test-infra-definitions/pull/943 to follow up on your suggestions. I feel like it is adding more changes to preserve one file that requires being generated once in a dev time. I don't expect the binary mode to be used out of the CI.

Anyhow no strong opinion, please review and tell me which version you folks think we should follow up with

PR #954, opened 2024-07-17
Fixing URL used to send logs in a container

carlosroman:
What does this PR do?
---------------------

It sets the correct URL for the Logs Agent when Fake Intake used.

-------------------

Any that run the Agent in Docker and are trying to send logs back to the Fake Intake.

Motivation
----------

I want some e2e tests to check a feature.

Additional Notes
----------------

None.

### DIFF
--- a/components/datadog/dockeragentparams/params.go
+++ b/components/datadog/dockeragentparams/params.go
@@ -3,8 +3,6 @@ package dockeragentparams
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
-
 	"github.com/DataDog/test-infra-definitions/common"
 	"github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/utils"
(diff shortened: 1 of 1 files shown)


components/datadog/dockeragentparams/params.go
```diff
@@ -136,18 +134,16 @@ func WithFakeintake(fakeintake *fakeintake.Fakeintake) func(*Params) error {
...
-		return strings.HasPrefix(host, "https"), nil
-	}).(pulumi.BoolOutput)
 	return func(p *Params) error {
 		envVars := pulumi.Map{
 			"DD_DD_URL":                                 pulumi.Sprintf("%s", url),
 			"DD_PROCESS_CONFIG_PROCESS_DD_URL":          pulumi.Sprintf("%s", url),
 			"DD_APM_DD_URL":                             pulumi.Sprintf("%s", url),
 			"DD_SKIP_SSL_VALIDATION":                    pulumi.Bool(shouldSkipSSLValidation),
 			"DD_REMOTE_CONFIGURATION_NO_TLS_VALIDATION": pulumi.Bool(shouldSkipSSLValidation),
+			"DD_LOGS_CONFIG_FORCE_USE_HTTP":             pulumi.Bool(true), // Force the use of HTTP/HTTPS rather than switching to TCP
```
pducolin, line 144: **❓ question**
Why removing the logic of `shouldEnforceHTTPInput`, that looking at it now is inverted?

carlosroman: This actually enforces the use of HTTP or HTTPS. Without this the Agent will attempt to connect to the Fake Intake and if that fails will then revert to trying to send logs via TCP. This isn't desirable behaviour (unless we are testing TCP fallback) as if it reverts to TCP it never reverts back to HTTP/HTTPS.

See [here](https://github.com/DataDog/datadog-agent/blob/42c07251e2fded462dde557410c1a80e00c5ed40/pkg/config/config_template.yaml#L941) for more info on this config option. It honestly not well named configuration option 😅

pducolin: No way, today I learned! Thank you for the explanation!

pducolin (approved):

PR #961, opened 2024-07-22
Automatically log in Azure if needed

FlorentClarret:
What does this PR do?
---------------------

Azure's counterpart of https://github.com/DataDog/test-infra-definitions/pull/917. I know this is a hack but:

- We can not easily use OIDC because we can't set the parameters in a public repo
- I tried `NewInteractiveBrowserCredential` from [the azure SDK](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#NewInteractiveBrowserCredential) to get the token. Unfortunately I can't use it with the pulumi's SDK: https://github.com/pulumi/pulumi-azure-native/issues/565

The idea is to automatically run `az login` if needed so users don't have to wrap their commands on their side. 

-------------------

Azure scenarios only! We currently don't have any 

Motivation
----------

Avoid asking people to manually run `az login` before running the tests. They might not know the tests is running in Azure, because it could be in AWS (or later in GCP)

Additional Notes
----------------

### DIFF
--- a/resources/azure/environment.go
+++ b/resources/azure/environment.go
@@ -1,11 +1,14 @@
 package azure
 
 import (
+	"fmt"
 	config "github.com/DataDog/test-infra-definitions/common/config"
 	"github.com/DataDog/test-infra-definitions/common/namer"
-
 	sdkazure "github.com/pulumi/pulumi-azure-native-sdk/v2"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+	"os/exec"
+	"strings"
+	"time"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


resources/azure/environment.go
```diff
@@ -118,3 +124,34 @@ func (e *Environment) GetCommonEnvironment() *config.CommonEnvironment {
 func (e *Environment) LinuxKataNodeGroup() bool {
 	return e.GetBoolWithDefault(e.InfraConfig, DDInfraAksLinuxKataNodeGroup, e.envDefault.ddInfra.aks.linuxKataNodeGroup)
 }
+
+func logIn(ctx *pulumi.Context, subscription string) {
```
FlorentClarret, line 128: We gotta check we're not in the CI though

KevinFairise2 (approved):

pducolin (approved):

PR #970, opened 2024-07-23
[task] return error if aws or az cli are not installed

pducolin:
What does this PR do?
---------------------

Add check at setup time to ensure az and aws cli are installed

-------------------

None

Motivation
----------

I did not have az cli installed, don't think it comes by default on our laptops. Better to warn contributors inline.

Additional Notes
----------------

### DIFF
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -27,13 +27,11 @@ def setup(
     Setup a local environment, interactively by default
     """
     # Ensure aws cli is installed
-    result = ctx.run("aws --version", warn=True, hide=True)
-    if result is None or result.exited != 0:
+    if not shutil.which("aws"):
         error("AWS CLI not found, please install it: https://aws.amazon.com/cli/")
         raise Exit(code=1)
     # Ensure azure cli is installed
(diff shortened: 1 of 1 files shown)


tasks/setup.py
```diff
@@ -26,6 +26,17 @@ def setup(
     """
     Setup a local environment, interactively by default
     """
+    # Ensure aws cli is installed
+    result = ctx.run("aws --version", warn=True, hide=True)
```
FlorentClarret, line 30: You could use [shutil.which](https://docs.python.org/3.11/library/shutil.html#shutil.which) (works on Windows too) with `if not shutil.which("aws"):`

FlorentClarret (approved):
LGTM but left one suggestion!

tasks/setup.py
pducolin: Ah, thank you!!

pducolin: Updated

FlorentClarret (approved):

tasks/setup.py
FlorentClarret: Nice!

PR #982, opened 2024-07-26
[CONTP-258] add namespace labels and annotations

zhuminyi:
What does this PR do?
---------------------
[CONTP-258](https://datadoghq.atlassian.net/browse/CONTP-258) 
Add namespaceLabelsAsTags and namespaceAnnotationsAsTags to k8s config

-------------------
k8s provider in kindvm

Motivation
----------
We don’t have any e2e test that uses “namespace labels as tags” and “namespace annotations as tags”.
It’d be good to add a simple test to notice if we accidentally break these features. 
Both configs are supported in [helm charts](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_components-common-env.yaml#L41-L47).

Additional Notes
----------------
Screenshot 2024-07-26 at 2 50 39 AM

inv create-kind

[CONTP-258]: https://datadoghq.atlassian.net/browse/CONTP-258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/datadog/agent/kubernetes_helm.go
+++ b/components/datadog/agent/kubernetes_helm.go
@@ -204,6 +204,12 @@ func buildLinuxHelmValues(baseName, agentImagePath, agentImageTag, clusterAgentI
 			"apiKeyExistingSecret": pulumi.String(baseName + "-datadog-credentials"),
 			"appKeyExistingSecret": pulumi.String(baseName + "-datadog-credentials"),
 			"checksCardinality":    pulumi.String("high"),
+			"namespaceLabelsAsTags": pulumi.Map{
+				"related_team": pulumi.String("team"),
+			},
+			"namespaceAnnotationsAsTags": pulumi.Map{
+				"related_email": pulumi.String("email"),
+			},
 			"logs": pulumi.Map{
 				"enabled":             pulumi.Bool(true),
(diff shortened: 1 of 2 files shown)


components/datadog/agent/kubernetes_helm.go
```diff
@@ -204,6 +204,12 @@ func buildLinuxHelmValues(baseName, agentImagePath, agentImageTag, clusterAgentI
 			"apiKeyExistingSecret": pulumi.String(baseName + "-datadog-credentials"),
 			"appKeyExistingSecret": pulumi.String(baseName + "-datadog-credentials"),
 			"checksCardinality":    pulumi.String("high"),
+			"namespaceLabelsAsTags": pulumi.Map{
```
pducolin, line 207: **❓ question**
Should these be params configurable from a test instead of fixed strings?

pducolin (approved):

components/datadog/agent/kubernetes_helm.go
zhuminyi: They (team and email) are tag names. The values are passed from nginx namespace. It is not convenient to set dummy apps namespace from a test. So we move them to test-infra-def.

PR #985, opened 2024-07-26
Add support for Windows VMs in Azure

pducolin:
What does this PR do?
---------------------

* Configure SSH in Windows VMs on Azure
* Export Windows password into host component

-------------------

Azure VM

Motivation
----------

Azure Windows VMs are configured to use RDP rather than SSH. SSH configuration takes [several steps](https://learn.microsoft.com/en-us/azure/virtual-machines/windows/connect-ssh?tabs=azurecli):

1. Enable SSH capability
2. Add user public key
3. Start sshd and configure it to run at startup

To do so I added a script that is pretty close to https://github.com/DataDog/test-infra-definitions/blob/florentclarret/windows-vm-azure/scenarios/aws/ec2/os_win.go#L8-L27

Additional Notes
----------------


resources/azure/compute/setup-ssh.ps1
```diff
@@ -0,0 +1,16 @@
+param(
+    [Parameter(Mandatory,HelpMessage='authorizedkey for ssh private key for the user')]
+    $authorizedKey
+)
+
+$service = Get-Service -Name sshd -ErrorAction SilentlyContinue
+if ($service -ne $null) {
+  Stop-Service sshd
+}
+Add-WindowsCapability -Online -Name OpenSSH.Server~~~~0.0.1.0
```
pducolin, line 10: Note for reviewers: this single line takes around one minute, an alternative is using a separate extension in `vm.go` and having the `NewVirtualMachineRunCommandByVirtualMachine ` depending on it. Content should be

```go
sshSetupExtension, err := compute.NewVirtualMachineExtension(e.Ctx(), "enable-ssh", &compute.VirtualMachineExtensionArgs{
		VmName:                  vm.Name,
		ResourceGroupName:       pulumi.String(e.DefaultResourceGroup()),
		Publisher:               pulumi.String("Microsoft.Azure.OpenSSH"),
		Type:                    pulumi.String("WindowsOpenSSH"),
		TypeHandlerVersion:      pulumi.String("3.0"),
		VmExtensionName:         pulumi.String("WindowsOpenSSH"),
		AutoUpgradeMinorVersion: pulumi.Bool(true),
	}, e.WithProviders(config.ProviderAzure), pulumi.Parent(vm))
```

resources/azure/compute/vm.go
```diff
@@ -19,10 +21,10 @@ const (
 	AdminUsername     = "azureuser"
 )
 
-func NewLinuxInstance(e azure.Environment, name, imageUrn, instanceType string, userData pulumi.StringPtrInput) (*compute.VirtualMachine, *network.NetworkInterface, error) {
+func NewLinuxInstance(e azure.Environment, name, imageUrn, instanceType string, userData pulumi.StringPtrInput, opts ...pulumi.ResourceOption) (vm *compute.VirtualMachine, privateIP pulumi.StringOutput, err error) {
```
pducolin, line 24: Note for reviewers: preferred returning only the private ip as we don't use any other param from the network configuration

resources/azure/compute/vm.go
```diff
@@ -74,23 +86,60 @@ func NewWindowsInstance(e azure.Environment, name, imageUrn, instanceType string
...
+		TreatFailureAsDeploymentFailure: pulumi.Bool(true),
+	}, pulumi.Parent(vm))
+
+	if err != nil {
+		return nil, pulumi.StringOutput{}, pulumi.StringOutput{}, err
+	}
+
+	privateIP = pulumi.All(nw.IpConfigurations.Index(pulumi.Int(0)).PrivateIPAddress().Elem(), setupSSHCommand.URN()).ApplyT(func(args []interface{}) string {
+		return args[0].(string)
+	}).(pulumi.StringOutput)
```
pducolin, line 123: Note for reviewers: this is a hack to allow returning the private IP after the script is done. Otherwise we start trying to connect to the VM while it's not ready. Not a huge deal, feel free to remove and return directly the `privateIP`

resources/azure/compute/vm.go
```diff
@@ -74,23 +86,60 @@ func NewWindowsInstance(e azure.Environment, name, imageUrn, instanceType string
...
+		},
+		Parameters: compute.RunCommandInputParameterArray{
+			compute.RunCommandInputParameterArgs{
+				Name:  pulumi.String("authorizedKey"),
+				Value: pulumi.String(publicKey),
+			},
+		},
+		TimeoutInSeconds:                pulumi.Int(120),
+		TreatFailureAsDeploymentFailure: pulumi.Bool(true),
+	}, pulumi.Parent(vm))
```
pducolin, line 115: Note for reviewers: this could move to a separate `configureSSH` function

resources/azure/compute/vm.go
```diff
@@ -74,23 +86,60 @@ func NewWindowsInstance(e azure.Environment, name, imageUrn, instanceType string
...
+
+	privateIP = pulumi.All(nw.IpConfigurations.Index(pulumi.Int(0)).PrivateIPAddress().Elem(), setupSSHCommand.URN()).ApplyT(func(args []interface{}) string {
+		return args[0].(string)
+	}).(pulumi.StringOutput)
+
+	return vm, privateIP, windowsAdminPassword.Result, nil
 }
 
-func newVMInstance(e azure.Environment, name, imageUrn, instanceType string, osProfile compute.OSProfilePtrInput) (*compute.VirtualMachine, *network.NetworkInterface, error) {
+func newVMInstance(e azure.Environment, name, imageUrn, instanceType string, osProfile compute.OSProfilePtrInput, opts ...pulumi.ResourceOption) (*compute.VirtualMachine, *network.NetworkInterface, error) {
```
pducolin, line 128: Note for reviewers: this allows passing the parent to all the resources, so that they belong to the parent component, similar to what we have in `aws`

resources/azure/compute/setup-ssh.ps1
pducolin: Never mind, I tried it and it took even longer to setup

```
+   pulumi:pulumi:Stack                                                  dd-paola-ducolin-pippo                       created (267s)      
 +   ├─ pulumi:providers:random                                           random                                       created (0.00s)     
 +   ├─ dd:Host                                                           az-vm                                        created (264s)      
 +   │  └─ command:remote:Command                                         remote-az-vm-cmd-wait-successful-connection  created (3s)        
 +   ├─ pulumi:providers:command                                          command                                      created (0.00s)     
 +   ├─ pulumi:providers:azure-native                                     azure                                        created (0.00s)     
 +   ├─ random:index:RandomString                                         az-az-vm-admin-password                      created (0.00s)     
 +   ├─ azure-native:network:NetworkInterface                             az-az-vm                                     created (5s)        
 +   └─ azure-native:compute:VirtualMachine                               az-az-vm                                     created (66s)       
 +      ├─ azure-native:compute:VirtualMachineExtension                   enable-ssh                                   created (125s)      
 +      └─ azure-native:compute:VirtualMachineRunCommandByVirtualMachine  az-vm-init-cmd                               created (64s) 
```

FlorentClarret (approved):
FTR, I: 

- Rebased on top of master
- Fixed some tiny issues related to the conflicts I did not properly solved during the rebase
- Changed the password for EC2 instances and hyperv becase we were getting panics with empty `StringOutput`. 
- Fixed running Linux on Azure

tasks/destroy.py
```diff
@@ -17,7 +17,7 @@ def destroy(
     scenario_name: str,
     config_path: Optional[str] = None,
     stack: Optional[str] = None,
-    force_yes: Optional[bool] = False,
```
KevinFairise2, line 20: If we set the default to `True` I think we should rather remove this option. I do not think it makes sense to have a `--no-force-yes` option

resources/azure/compute/vm.go
KevinFairise2: I think we could avoid that by making the `VM` resource a child of the `Network`. 
Like that we'd have
Connection waits for Network -> VM -> SSHSetup command
instead of 
Connection waits for Network -/-> VM -> SSH setup command
That can definitely be done in a follow up PR if this is working like that

KevinFairise2 (approved):

tasks/azure/vm.py
```diff
@@ -90,7 +118,7 @@ def destroy_vm(
     ctx: Context,
     config_path: Optional[str] = None,
     stack_name: Optional[str] = None,
-    yes: Optional[bool] = False,
+    yes: Optional[bool] = True,
```
KevinFairise2, line 121: We should remove that. Otherwise you need `--no-yes` to use that option which is weird

PR #986, opened 2024-07-29
Improve management of Otel configuration and configure fakeintake in otel config

KevinFairise2:
…

What does this PR do?
---------------------

Build OTel config to have the fakeintake URL.
It also refactor a bit how we manage OTel configuration

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/utils/yaml.go
+++ b/common/utils/yaml.go
@@ -1,0 +1,52 @@
+package utils
+
+import "gopkg.in/yaml.v3"
+
+func MergeYAML(oldValuesYamlContent string, newValuesYamlContent string) (string, error) {
+	if oldValuesYamlContent == "" {
+		return newValuesYamlContent, nil
+	}
+
+	if newValuesYamlContent == "" {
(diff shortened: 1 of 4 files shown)


common/utils/yaml.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	mergedValuesYAML := MergeMaps(oldValuesYAML, newValuesYAML)
+
+	mergedValues, err := yaml.Marshal(mergedValuesYAML)
+
+	return string(mergedValues), err
+}
+
+func MergeMaps(a, b map[string]interface{}) map[string]interface{} {
```
AliDatadog, line 35: Would this do the trick ? https://github.com/samber/lo?tab=readme-ov-file#assign

KevinFairise2: I quickly read the function code and I think it would not. Because I need deep merge (my function recursively merge nested maps)

FlorentClarret (approved):

PR #1021, opened 2024-08-14
Add option to deploy fakeintake on Azure, in Docker on a VM

KevinFairise2:
What does this PR do?
---------------------

This PR add a component to deploy the fakeintake on Azure.
I first tried to use `Container Apps` or `App Service` to deploy the fakeintake. After several tests I did not succeed to deploy a fakeintake without public IP that can be reached from local through AppGate.
Main issues are:
- Container Apps/App Service need a dedicated subnet different from the one we use for VM -> some more configuration would be needed on AppGate
- Routing to Container Apps and App Services containers is made using Host based routing, the DNS is provided by Azure. It was doable easily with a private DNS zone resolvable only from within Azure VNET but not locally. It should be possible to provide our custom DNS but we'd need to generate a wildcard certificate on a dns zone we own and import it into Azure

In the end I decided to deploy the fakeintake in Docker on a simple VM. It is working and should be enough for the few tests we plan to have on Azure

To have something similar to what we have on AWS we could also setup monitoring with another agent installed on the fakeintake VM: https://datadoghq.atlassian.net/jira/software/c/projects/ADXT/boards/6196?selectedIssue=ADXT-511&useStoredSettings=true

-------------------

Motivation
----------

Additional Notes
----------------


scenarios/azure/fakeintake/fakeintake.go
```diff
@@ -0,0 +1,84 @@
...
+		// 	ResourceGroupName: pulumi.String(e.DefaultResourceGroup()),
+		// 	VnetConfiguration: &app.VnetConfigurationArgs{
+		// 		Internal:               pulumi.Bool(true),
+		// 		InfrastructureSubnetId: pulumi.String(e.DefaultSubnet()),
+		// 	},
+		// }, e.WithProviders(config.ProviderAzure))
+
+		// if err != nil {
+		// 	return err
+		// }
```
pducolin, line 29: **❓ question**
Is this commented section a leftover?
```suggestion
```

scenarios/azure/fakeintake/fakeintake.go
```diff
@@ -0,0 +1,84 @@
...
+
+		containerApp, err := app.NewContainerApp(e.Ctx(), "fakeintake-2", &app.ContainerAppArgs{
+			ResourceGroupName: pulumi.String(e.DefaultResourceGroup()),
+			EnvironmentId:     pulumi.String("/subscriptions/9972cab2-9e99-419b-a683-86bfa77b3df1/resourceGroups/dd-agent-sandbox/providers/Microsoft.App/managedEnvironments/fakeintake-274b947f2"),
```
pducolin, line 33: **❓ question**
Should this be a const? Is it shared across `agent-qa` and `agent-sandbox`?

pducolin (approved):

scenarios/azure/fakeintake/fakeintake.go
KevinFairise2: Oh yeah it was supposed to be deleted. Removing it now

scenarios/azure/fakeintake/fakeintake.go
KevinFairise2: Just pushed the changes I had locally that I forgot to push. This part should no longer be there

PR #1039, opened 2024-08-21
Add needed configuration to create provisioner on datadog-agent

KevinFairise2:
What does this PR do?
---------------------

Add some components that can be used in a provisioner on datadog-agent side. It also configure variables for Azure

-------------------

Motivation
----------

Additional Notes
----------------


resources/azure/aks/cluster.go
```diff
@@ -70,6 +71,17 @@ func NewCluster(e azure.Environment, name string, nodePool containerservice.Mana
 	if err != nil {
 		return nil, pulumi.StringOutput{}, err
 	}
+	_, err = authorization.NewRoleAssignment(e.Ctx(), e.Namer.ResourceName(name), &authorization.RoleAssignmentArgs{
+		PrincipalId: cluster.IdentityProfile.ApplyT(func(identity map[string]containerservice.UserAssignedIdentityResponse) string {
+			return *identity["kubeletidentity"].ObjectId
+		}).(pulumi.StringOutput),
+		PrincipalType:    pulumi.String("ServicePrincipal"),
+		Scope:            pulumi.String(e.DefaultContainerRegistry()),
+		RoleDefinitionId: pulumi.Sprintf("/subscriptions/%s/providers/Microsoft.Authorization/roleDefinitions/7f951dda-4ed3-4680-a7ca-43fe172d538d", e.DefaultSubscriptionID()),
```
FlorentClarret, line 80: **❓ question:** ‏

Should `7f951dda-4ed3-4680-a7ca-43fe172d538d` be in the environment and not hardcoded?

FlorentClarret (approved):
LGTM, left one question

resources/azure/aks/cluster.go
KevinFairise2: Actually I think we do not need to put that in the environment. Because this is the id of the `AcrPull` role that is the same in all Azure subscriptions.
https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles/containers#acrpull

pducolin (approved):
Very nice!

PR #1045, opened 2024-08-22
Add AWS key create/import to setup task

clarkb7:
What does this PR do?
---------------------
Add ability to create or import an AWS key during `inv setup`, or through new invoke tests `inv setup.aws-create-keypair` and `inv setup.aws-import-keypair`, and update the config.

-------------------
setup

Motivation
----------
easier setup and key rotation

Additional Notes
----------------

example new keypair
```
$ inv setup --config-path /tmp/test-config
Pulumi is up to date: v3.129.0
Logged in to LAPTOP-J4KNN9ON as brandenclark (file://~)
🤖 Let's configure your environment for e2e tests! Press ctrl+c to stop me
Which aws account do you want to create instances on? Default [agent-sandbox], available [agent-sandbox|sandbox|tse-playground]:
Do you want to create a new key pair? [Y/N] Default [N]: Y
🔑 Key pair name (default: brandenclark): bc-test-keypair
Creating Windows VMs requires "rsa" key type
🔑 Key type [rsa|ed25519] (default: rsa):
🔑 Private key path (default: /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pem):
🔑 Public key path (default: /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pub):
...
```

example new keypair (confirms overwrite)
```
$ inv setup --config-path /tmp/test-config
Pulumi is up to date: v3.129.0
Logged in to LAPTOP-J4KNN9ON as brandenclark (file://~)
🤖 Let's configure your environment for e2e tests! Press ctrl+c to stop me
Which aws account do you want to create instances on? Default [agent-sandbox], available [agent-sandbox|sandbox|tse-playground]:
Do you want to create a new key pair? [Y/N] Default [N]: Y
🔑 Key pair name (default: brandenclark): bc-test-keypair
Creating Windows VMs requires "rsa" key type
🔑 Key type [rsa|ed25519] (default: rsa):
🔑 Private key path (default: /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pem):
🔑 Public key path (default: /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pub):
Key pair bc-test-keypair already exists.
Do you want to overwrite it? [Y/N] Default [N]: Y
Private key /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pem already exists.
Do you want to overwrite it? [Y/N] Default [N]: Y
...
```

example import existing
```
$ inv setup --config-path /tmp/test-config
Pulumi is up to date: v3.129.0
Logged in to LAPTOP-J4KNN9ON as brandenclark (file://~)
🤖 Let's configure your environment for e2e tests! Press ctrl+c to stop me
Which aws account do you want to create instances on? Default [agent-sandbox], available [agent-sandbox|sandbox|tse-playground]:
Do you want to create a new key pair? [Y/N] Default [N]: N
Do you want to import an existing key pair? [Y/N] Default [N]: Y
🔑 Key pair name (default: brandenclark): bc-test-keypair
🔑 Private key path (default: /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pem): /home/brandenclark/.ssh/mykey.pem
🔑 Public key path (default: /home/brandenclark/.ssh/id_rsa_agent_sandbox_bc-test-keypair.pub): /home/brandenclark/.ssh/mykey.pub
...
```

### DIFF
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -333,7 +333,7 @@ def resolve_keypair_opts(
             account_part = f"{awsConf.account}_" if awsConf.account else ""
             account_part = account_part.replace("-", "_")
             default_private_key_path = Path.home().joinpath(
-                ".ssh", f'id_{key_type or "rsa"}_e2e_{keypair_name}.{key_format}'
+                ".ssh", f'id_{key_type or "rsa"}_e2e_{account_part}{keypair_name}.{key_format}'
             )
         while True:
             private_key_path = ask(f"🔑 Private key path (default: {default_private_key_path}): ")
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -329,8 +329,11 @@ def resolve_keypair_opts(
         key_format = "pem"
     if not private_key_path:
         if default_private_key_path is None or keypair_name != default_keypair_name:
+            # Example: ~/.ssh/id_rsa_e2e_agent_sandbox_mykeyname.pem
+            account_part = f"{awsConf.account}_" if awsConf.account else ""
+            account_part = account_part.replace("-", "_")
             default_private_key_path = Path.home().joinpath(
-                ".ssh", f'id_{key_type or "rsa"}_{keypair_name}.{key_format}'
+                ".ssh", f'id_{key_type or "rsa"}_e2e_{keypair_name}.{key_format}'
             )
         while True:
             private_key_path = ask(f"🔑 Private key path (default: {default_private_key_path}): ")
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -14,9 +14,10 @@ from invoke.tasks import task
 
 from . import doc
 from .config import Config, get_full_profile_path, get_local_config
-from .tool import ask, debug, error, info, is_linux, is_windows, warn
+from .tool import ask, ask_yesno, debug, error, get_aws_cmd, info, is_linux, is_windows, warn
 
 available_aws_accounts = ["agent-sandbox", "sandbox", "agent-qa", "tse-playground"]
+supported_key_types = ["rsa", "ed25519"]
 
 
 @task(help={"config_path": doc.config_path, "interactive": doc.interactive, "debug": doc.debug}, default=True)
@@ -54,7 +55,7 @@ def setup(
     if interactive:
         info("🤖 Let's configure your environment for e2e tests! Press ctrl+c to stop me")
         # AWS config
-        setupAWSConfig(config)
+        setupAWSConfig(ctx, config)
         # Azure config
         setup_azure_config(config)
         # Agent config
@@ -113,7 +114,7 @@ def _check_config(config: Config):
         warn("WARNING: privateKeyPassword is set. Please ensure privateKeyPath is used ONLY for E2E tests.")
 
 
-def setupAWSConfig(config: Config):
+def setupAWSConfig(ctx: Context, config: Config):
     if config.configParams is None:
         config.configParams = Config.Params(aws=None, agent=None, pulumi=None, azure=None, devMode=False)
     if config.configParams.aws is None:
@@ -134,35 +135,23 @@ def setupAWSConfig(config: Config):
             break
         warn(f"{config.configParams.aws.account} is not a valid aws account")
 
-    # aws keypair name
-    if config.configParams.aws.keyPairName is None:
-        config.configParams.aws.keyPairName = getpass.getuser()
-    keyPairName = ask(f"🔑 Key pair name - stored in AWS Sandbox, default [{config.configParams.aws.keyPairName}]: ")
-    if len(keyPairName) > 0:
-        config.configParams.aws.keyPairName = keyPairName
+    # ask user if they want to create a new key or import an existing key
+    if ask_yesno("Do you want to create a new key pair?"):
+        _aws_create_keypair(ctx, config, use_aws_vault=True, aws_account_name=config.configParams.aws.account)
+    elif ask_yesno("Do you want to import an existing key pair?"):
+        _aws_import_keypair(ctx, config, use_aws_vault=True, aws_account_name=config.configParams.aws.account)
+    else:
+        warn("No key pair configured, you will need to manually configure a key pair")
 
     # check keypair name
     if config.options is None:
         config.options = Config.Options(checkKeyPair=False)
     default_check_key_pair = "Y" if config.options.checkKeyPair else "N"
     checkKeyPair = ask(
-        f"Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: "
+        f"Do you want check that the keypair is loaded in ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: "
     )
     if len(checkKeyPair) > 0:
         config.options.checkKeyPair = checkKeyPair.lower() == "y" or checkKeyPair.lower() == "yes"
-
-    # aws public key path
-    if config.configParams.aws.publicKeyPath is None:
-        config.configParams.aws.publicKeyPath = str(Path.home().joinpath(".ssh", "id_ed25519.pub").absolute())
-    default_public_key_path = config.configParams.aws.publicKeyPath
-    while True:
-        config.configParams.aws.publicKeyPath = default_public_key_path
-        publicKeyPath = ask(f"🔑 Path to your public ssh key, default [{config.configParams.aws.publicKeyPath}]: ")
-        if len(publicKeyPath) > 0:
-            config.configParams.aws.publicKeyPath = publicKeyPath
-        if os.path.isfile(config.configParams.aws.publicKeyPath):
-            break
-        warn(f"{config.configParams.aws.publicKeyPath} is not a valid ssh key")
 
     # team tag
     if config.configParams.aws.teamTag is None:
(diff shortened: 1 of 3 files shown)


tasks/setup.py
```diff
@@ -134,36 +135,24 @@ def setupAWSConfig(config: Config):
...
-    keyPairName = ask(f"🔑 Key pair name - stored in AWS Sandbox, default [{config.configParams.aws.keyPairName}]: ")
-    if len(keyPairName) > 0:
-        config.configParams.aws.keyPairName = keyPairName
+    # ask user if they want to create a new key or import an existing key
+    if ask_yesno("Do you want to create a new key pair?"):
+        _aws_create_keypair(ctx, config, use_aws_vault=True, aws_account_name=config.configParams.aws.account)
+    elif ask_yesno("Do you want to import an existing key pair?"):
+        _aws_import_keypair(ctx, config, use_aws_vault=True, aws_account_name=config.configParams.aws.account)
+    else:
+        warn("No key pair configured, you will need to manually configure a key pair")
```
pducolin, line 144: **🔨 warning**
You are overriding the stored key pair name, you should keep the previous code and integrate it in this logic. A possible solution would be:
```suggestion
   if config.configParams.aws.keyPairName is None:
   	config.configParams.aws.keyPairName = getpass.getuser()
    # ask user if they want to create a new key or import an existing key
    if ask_yesno("Do you want to create a new key pair?"):
        _aws_create_keypair(ctx, config, use_aws_vault=True, aws_account_name=config.configParams.aws.account)
    elif ask_yesno("Do you want to import an existing key pair?"):
        _aws_import_keypair(ctx, config, use_aws_vault=True, aws_account_name=config.configParams.aws.account)
    else:
        info(f"Using key pair {config.configParams.aws.keyPairName}")
```

And then use `config.configParams.aws.keyPairName` as the default name. I removed the warning as most of the time the key pair is stored in aws agent sandbox

tasks/setup.py
```diff
@@ -134,36 +135,24 @@ def setupAWSConfig(config: Config):
...
 
     # check keypair name
     if config.options is None:
         config.options = Config.Options(checkKeyPair=False)
     default_check_key_pair = "Y" if config.options.checkKeyPair else "N"
     checkKeyPair = ask(
-        f"Did you create your SSH key on AWS and want me to check it is loaded on your ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: "
+        f"Do you want check that the keypair is loaded in ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: "
```
pducolin, line 151:
```suggestion
        f"Do you want to check if the keypair is loaded in ssh agent when creating manual environments or running e2e tests [Y/N]? Default [{default_check_key_pair}]: "
```

tasks/setup.py
```diff
@@ -134,36 +135,24 @@ def setupAWSConfig(config: Config):
...
-    default_public_key_path = config.configParams.aws.publicKeyPath
-    while True:
-        config.configParams.aws.publicKeyPath = default_public_key_path
-        publicKeyPath = ask(f"🔑 Path to your public ssh key, default [{config.configParams.aws.publicKeyPath}]: ")
-        if len(publicKeyPath) > 0:
-            config.configParams.aws.publicKeyPath = publicKeyPath
-        if os.path.isfile(config.configParams.aws.publicKeyPath):
-            break
-        warn(f"{config.configParams.aws.publicKeyPath} is not a valid ssh key")
-
```
pducolin, line 166: **🔨 warning**
Removing this code might break the experience for folks with an already configured environment, we should try to be backward compatible

tasks/setup.py
```diff
@@ -277,6 +266,353 @@ def setupPulumiConfig(config):
...
+
+    # defaults
+    if awsConf.keyPairName:
+        default_keypair_name = awsConf.keyPairName
+    else:
+        default_keypair_name = getpass.getuser()
+    if awsConf.privateKeyPath:
+        default_private_key_path = awsConf.privateKeyPath
+    else:
+        default_private_key_path = None
```
pducolin, line 299: **💬 suggestion**
I would be more opinionated here and use the value we suggest in the documentation, `id_rsa`

tasks/setup.py
```diff
@@ -277,6 +266,353 @@ def setupPulumiConfig(config):
...
+    else:
+        default_public_key_path = None
+
+    # ask for missing values
+    if not keypair_name:
+        keypair_name = ask(f"🔑 Key pair name (default: {default_keypair_name}): ")
+        if not keypair_name:
+            keypair_name = default_keypair_name
+    if not key_type and require_key_type:
+        warn('Creating Windows VMs requires "rsa" key type')
```
pducolin, line 311: **💬 suggestion**
I would create only `rsa` keys as Windows VMs support only `rsa` keys and other VMs support it
I would avoid warning if possible, except on issues

tasks/setup.py
```diff
@@ -277,6 +266,353 @@ def setupPulumiConfig(config):
...
+            supported = "|".join(supported_key_types)
+            key_type = ask(f"🔑 Key type [{supported}] (default: {keytype_default}): ")
+            if not key_type:
+                key_type = keytype_default
+            if key_type in supported_key_types:
+                break
+            warn(f"{key_type} is not a supported key type")
+    if not key_format:
+        # Just use pem for now
+        key_format = "pem"
```
pducolin, line 323: **❓ question**
What other formats are supported?

tasks/setup.py
```diff
@@ -277,6 +266,353 @@ def setupPulumiConfig(config):
...
+    # generate private key
+    cmd = f'ec2 create-key-pair --key-name "{keypair_name}" --key-type {key_type} --query KeyMaterial --output text'
+    out = ctx.run(_get_aws_cmd(cmd), hide=True)
+    if out is None:
+        raise Exit(f"Failed to create key pair {keypair_name}")
+    key_material = out.stdout.strip()
+    # write private key to disk
+    os.makedirs(Path(private_key_path).parent, exist_ok=True)
+    with open(private_key_path, "w") as f:
+        f.write(key_material)
```
pducolin, line 502: **💭 thought**
There are a couple of missing steps:
- adding a passkey to protect the private key, or ssh scanner will complain
- adding the key to ssh-agent, at list on unix environments

tasks/tool.py
```diff
@@ -33,6 +33,18 @@ def ask(question: str) -> str:
...
 
 
+def ask_yesno(question: str, default='N') -> bool:
+    res = ""
+    yes_opts = ["y", "yes"]
+    no_opts = ["n", "no"]
+    while res.lower() not in (yes_opts + no_opts):
+        res = ask(question + f" [Y/N] Default [{default}]: ")
+        if res == "":
+            res = default
```
pducolin, line 43: **💬 suggestion**
No need to iterate one more time
```suggestion
            res = default
            break
```

pducolin (approved):
Nice add, few comments to prevent breaking the experience for already configured environments

tasks/setup.py
clarkb7: It's not being overridden, it's used as the default in [resolve_keypair_opts](https://github.com/DataDog/test-infra-definitions/pull/1045/files#diff-c17ad71d631afbbe74eea29844a37a1501410870263f1ea09d58f4a786e91772R279-R283).

I'll update the warning to only show if no values are set.

tasks/setup.py
clarkb7: it wasn't removed, it was moved to [resolve_keypair_opts](https://github.com/DataDog/test-infra-definitions/pull/1045/files#diff-c17ad71d631afbbe74eea29844a37a1501410870263f1ea09d58f4a786e91772R279-R283).

tasks/setup.py
clarkb7: this is done further down, and takes the key type, keypair name, and key format into account
https://github.com/DataDog/test-infra-definitions/pull/1045/files#diff-c17ad71d631afbbe74eea29844a37a1501410870263f1ea09d58f4a786e91772R326-R328

tasks/setup.py
clarkb7: putty's ppk format

tasks/setup.py
clarkb7: > adding a passkey to protect the private key, or ssh scanner will complain

There are some weird synergies between us, pulumi, and AWS, here., i've previously documented them [here](https://github.com/DataDog/test-infra-definitions/blob/b7042cd9a532218fb9bb6f787cd020948b4b9786/tasks/setup.py#L506-L525).
- if you have an encrypted `privateKeyPath` configured, but don't provide `privateKeyPassword`, then pulumi will fail and won't check ssh-agent.
- our `host_ssh.go` code looks similar, but doesn't append the auth methods so will skip ssh-agent if `privateKeyPath` is configured.
- pulumi doesn't support windows ssh-agent so we need to `privateKeyPath` and if encrypted `privateKeyPassword`
- if you don't have `privateKeyPath` configured, or if it is encrypted, you can't RDP to Windows hosts / decrypt their password

I think they're all surmountable, but will require changes in other places. For example, our RDP helper task could prompt the user for the passphrase and temporarily decrypt the key when requesting the password from AWS.

tasks/setup.py
pducolin: Oh I see, so by default now the key is `username`.pem. This might conflict with other users keys, it changes from the current default of using `id_rsa.pem` for the path and `<username>` for the keypair name. Could we keep the current behaviour?

tasks/setup.py
pducolin: We could also create the key on the fly as @vboulineau suggested a long time ago, at VM creation, and then remove the ssh key both locally and remote. This way DD ssh key monitor doesn't bother us, as they are ephemeral.

pducolin: It would remove the need for this setup. We would keep only the key for the CI, stored in 1Password and in `AWS`

clarkb7: that would be neat!

PR #1049, opened 2024-08-22
Improvements to `setup.debug` task

clarkb7:
What does this PR do?
---------------------
fix fingerprint parsing bug (was chopping off first byte)

Match imported RSA keys by calculating public key MD5 from RSA private key 

Allow `AWS_PROFILE` as an alternative to `AWS_VAULT`

Skip `ssh-agent` assertions on Windows, since pulumi doesn't support it

-------------------
troubleshooting

Motivation
----------
https://datadoghq.atlassian.net/browse/WINA-980

Additional Notes
----------------
run with
```
inv setup.debug
```


pducolin (approved):
This is awesome, thank you!

PR #1089, opened 2024-09-09
Add the tasks to run VMs on GCP

FlorentClarret:
What does this PR do?
---------------------

Add the tasks to run VMs on GCP

- `inv gcp.create-vm`
- `inv gcp.destroy-vm`

They work the same way Azure and AWS commands work

-------------------

Motivation
----------

GCP is now available

Additional Notes
----------------

To make it work, you need to log in manually for now:

- Run `gcloud auth application-default login`
- or with a service account `export GOOGLE_CREDENTIALS=<path-to-credentials.json>`

### DIFF
--- a/tasks/gcp/common.py
+++ b/tasks/gcp/common.py
@@ -1,0 +1,57 @@
+from typing import List, Union
+
+
+def get_default_os_family() -> str:
+    return "ubuntu"
+
+
+def get_os_families() -> List[str]:
+    return [
+        get_default_os_family(),
+    ]
+
+
+def get_package_for_os(os: str) -> str:
+    package_map = {
+        get_default_os_family(): "deb",
+    }
+
+    return package_map[os]
+
+
+def get_deploy_job(os: str, arch: Union[str, None], agent_version: Union[str, None] = None) -> str:
+    """
+    Returns the deploy job name within the datadog agent repo that creates
+    images used in create-vm
+    """
+    pkg = get_package_for_os(os)
+    if agent_version is None:
+        v = 'a7'
+    else:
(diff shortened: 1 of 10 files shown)


tasks/gcp/common.py
```diff
@@ -0,0 +1,57 @@
+from typing import List, Union
+
```
KevinFairise2, line 2: Can probably be done in a follow up PRs. But I feel like most of these functions could defined in a `tasks/common.py` file so we do not need to have it duplicated for each cloud provider

tasks/gcp/vm.py
```diff
@@ -0,0 +1,160 @@
...
+    agent_version: Optional[str] = None,
+    debug: Optional[bool] = False,
+    interactive: Optional[bool] = True,
+    ssh_user: Optional[str] = None,
+    account: Optional[str] = None,
+    os_family: Optional[str] = None,
+    os_version: Optional[str] = None,
+    architecture: Optional[str] = None,
+    instance_type: Optional[str] = None,
+    deploy_job: Optional[str] = None,
```
KevinFairise2, line 58: Is that useful? It is currently not used

KevinFairise2 (approved):

tasks/gcp/common.py
FlorentClarret: Some of them only in fact because the default architecture, os, etc can be different. But some of them could, I'll have a look!

tasks/gcp/vm.py
FlorentClarret: I'll drop it!

PR #1111, opened 2024-09-18
Add architecture and Descriptor to HostOutput component

pgimalac:
What does this PR do?
---------------------
Add `Architecture` field and `Descriptor` function to HostOutput component.

-------------------

Motivation
----------
Being able to update the environment easily no matter the host (in particular we have similar tests on Linux and Windows):

```go
v.UpdateEnv(awshost.Provisioner(
	awshost.WithEC2InstanceOptions(
		ec2.WithOS(v.Env().RemoteHost.Descriptor()),
		... // <- make your actual change, eg. agent config
	),
))
```

Additional Notes
----------------
I think keeping the OS should actually be done by `UpdateEnv` by default without having to explicitly re-declare your OS/Arch, but as it's not currently the case this is a convenient way to make it simpler.

### DIFF
--- a/components/remote/component.go
+++ b/components/remote/component.go
@@ -11,12 +11,17 @@ import (
 type HostOutput struct {
 	components.JSONImporter
 
-	Address   string    `json:"address"`
-	Username  string    `json:"username"`
-	Password  string    `json:"password,omitempty"`
-	OSFamily  os.Family `json:"osFamily"`
-	OSFlavor  os.Flavor `json:"osFlavor"`
-	OSVersion string    `json:"osVersion"`
+	Address      string          `json:"address"`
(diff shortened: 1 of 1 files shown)


vboulineau:
I understand the use case, but in your test code, either you use the default value or you have access to the original value as well as, no?

pgimalac:
@vboulineau but the VM creation is done when you create the suite with `e2e.Run`, then when you're in a test and you call `UpdateEnv` you lost that information.
I could just store it in the suite struct I guess, but I feel like just adding the arch to the host fields makes it much easier.

You can see https://github.com/DataDog/datadog-agent/pull/29429 for an example building on top of this PR.

PR #1114, opened 2024-09-19
Add support for Autopilot

FlorentClarret:
What does this PR do?
---------------------

This PR adds a flag and a configmap parameter to use Autopilot on GKE. As a lot of features are not authorized by GKE the default Helm values enable fewer things by default.
It is likely to not work with any custom image as Datadog is granted some exception on Autopilot for its stable images. We'll probably need to request an exception to be able to deploy images built in the pipeline. 

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/resources/gcp/gke/cluster.go
+++ b/resources/gcp/gke/cluster.go
@@ -11,6 +11,7 @@ func NewCluster(e gcp.Environment, name string, autopilot bool, opts ...pulumi.R
 	opts = append(opts, e.WithProviders(config.ProviderGCP))
 
 	cluster, err := container.NewCluster(e.Ctx(), e.Namer.ResourceName(name), &container.ClusterArgs{
+		InitialNodeCount:   pulumi.Int(1),
 		MinMasterVersion:   pulumi.String(e.KubernetesVersion()),
 		NodeVersion:        pulumi.String(e.KubernetesVersion()),
 		DeletionProtection: pulumi.Bool(false),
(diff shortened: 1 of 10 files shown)


resources/gcp/gke/cluster.go
```diff
@@ -7,14 +7,15 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-func NewCluster(e gcp.Environment, name string, opts ...pulumi.ResourceOption) (*container.Cluster, pulumi.StringOutput, error) {
+func NewCluster(e gcp.Environment, name string, autopilot bool, opts ...pulumi.ResourceOption) (*container.Cluster, pulumi.StringOutput, error) {
 	opts = append(opts, e.WithProviders(config.ProviderGCP))
 
 	cluster, err := container.NewCluster(e.Ctx(), e.Namer.ResourceName(name), &container.ClusterArgs{
-		InitialNodeCount:   pulumi.Int(2),
```
pducolin, line 14: **❓ question**
Is it expected to remove `InitialNodeCount`?

pducolin (approved):

resources/gcp/gke/cluster.go
KevinFairise2: Nice catch. Actually it breaks autopilot if we set it to 2. But it break classic GKE if we do not set it all. Updating the value to 1 which should be ok for both the deployments

PR #1117, opened 2024-09-20
kmt: validate downloaded images

gjulianm:
What does this PR do?
---------------------

Modifies the download process of VM images by adding a Python script that performs the download based on a JSON file with instructions:

1. It first checks the checksum before downloading the image. This allows us to avoid downloads of existing images, but also force a re-download if the image is corrupt.
2. Once the script checks which images need to be downloaded, it will do so with `curl`.
3. After the download, the 	checksum is checked to ensure the image is not corrupted.

-------------------

microVMs

Motivation
----------

The purpose of these changes is to make retries more consistent and correct.

Additional Notes
----------------

Tested in https://github.com/DataDog/datadog-agent/pull/29472 - pipeline https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/45694168

### DIFF
--- a/scenarios/aws/microVMs/microvms/files/download-vm-images.py
+++ b/scenarios/aws/microVMs/microvms/files/download-vm-images.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+
+def usage():
+    print(f"Usage: {sys.argv[0]} <data-file>")
+
+
+def main(data_file):
+    with open(data_file) as f:
+        download_data = json.load(f)
+
+    images_to_download = []
+
+    for image in download_data:
+        checksum_file = Path(image["checksum_path"])
+        if not check_integrity(checksum_file):
+            print(f"Integrity check failed for checksum file: {checksum_file}, downloading image")
+            images_to_download.append(image)
+
+    curl_args = ["curl", "--no-progress-meter", "--fail", "--show-error", "--retry", "3", "--parallel"]
+    for image in images_to_download:
+        curl_args += [image["image_source"], "-o", image["image_path"]]
+        curl_args += [image["checksum_source"], "-o", image["checksum_path"]]
+
+    try:
+        subprocess.run(curl_args, check=True)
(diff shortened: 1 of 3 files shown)


usamasaqib:
We have had issues in the past with having an individual task for downloading each VM image. This approach leads to ssh connection instability. The theory here is that having a lot of parallel ssh connections open while loading the network can cause the keep alives to fail, making pulumi think the process on the server exited without a status code.

From what I remember pulumi does not provide a way to change the ssh keep alive interval, which could have helped us to solve this problem.

scenarios/aws/microVMs/microvms/files/download-vm-images.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
```
brycekahle, line 1: This is adding python as a dependency on the metal instance, correct? I don't think we currently require python.

gjulianm: We don't require it, but AFAIK Python is always available, a lot of system tools already rewuure it. In this case I didn't have to add any dependency to the metal instance AMI, I also used only standard libraries to avoid extra dependencies. 

I tried to use Bash at first, but parsing the images to download wasn't trivial and the code would have ended up more complex and fragile.

scenarios/aws/microVMs/microvms/files/download-vm-images.py
```diff
@@ -0,0 +1,81 @@
...
+            images_to_download.append(image)
+
+    curl_args = [
+        "curl",
+        "--no-progress-meter",
+        "--fail",
+        "--show-error",
+        "--retry",
+        "3",
+        "--parallel",
```
brycekahle, line 32: I think we are missing `--parallel-max` to prevent the issue @usamasaqib was talking about.

gjulianm: Fixed, also added a comment about it (I didn't know that `curl` [defaults to 50](https://everything.curl.dev/cmdline/urls/parallel.html) for `--parallel-max`)

PR #1125, opened 2024-09-26
[ADXT-514] Setup for GCP

KevinFairise2:
What does this PR do?
---------------------

Add a step in the setup to make sure `gke-gcloud-auth-plugin` is installed. This is required to run creation of GKE cluster
It also fixes default account for GKE that should be `agent-sandbox`

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/tasks/setup.py
+++ b/tasks/setup.py
@@ -40,6 +40,9 @@ def setup(
         error("Gcloud CLI not found, please install it: https://cloud.google.com/sdk/docs/install")
         raise Exit(code=1)
 
+    # Ensure gke-gcloud-auth-plugin is installed
+    _install_gcloud_auth_plugin(ctx)
+
     pulumi_version, pulumi_up_to_date = _pulumi_version(ctx)
     if pulumi_up_to_date:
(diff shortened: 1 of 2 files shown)


tasks/setup.py
```diff
@@ -114,6 +117,22 @@ def _install_pulumi(ctx: Context):
             os.environ["PATH"] = f"{os.environ['PATH']}:{path}"
 
 
+# Check if gke-gcloud-auth-plugin is installed and install it if not
+def _install_gcloud_auth_plugin(ctx):
+    res = ctx.run("gcloud components list --format=json --filter 'name: gke-gcloud-auth-plugin'", hide=True)
```
pducolin, line 122: **💬 suggestion**
You might need to skip this when on Windows, not sure it works there

pducolin (approved):

tasks/setup.py
KevinFairise2: I think it should work on Windows as well because the `gcloud` cli is available on Windows. And we make sure the CLI is installed even on Windows

PR #1141, opened 2024-10-01
Do not put special characters in Winwdows random password

KevinFairise2:
What does this PR do?
---------------------

#incident-31096

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/ec2/vm.go
+++ b/scenarios/aws/ec2/vm.go
@@ -64,7 +64,7 @@ func NewVM(e aws.Environment, name string, params ...VMOption) (*remote.Host, er
 		if vmArgs.osInfo.Family() == os.WindowsFamily {
 			randomPassword, err := random.NewRandomString(e.Ctx(), e.Namer.ResourceName(name, "win-admin-password"), &random.RandomStringArgs{
 				Length:  pulumi.Int(20),
-				Special: pulumi.Bool(true),
+				Special: pulumi.Bool(false),
 			}, pulumi.Parent(c), e.WithProviders(config.ProviderRandom))
 			if err != nil {
 				return err


scenarios/aws/ec2/vm.go
```diff
@@ -64,7 +64,7 @@ func NewVM(e aws.Environment, name string, params ...VMOption) (*remote.Host, er
 		if vmArgs.osInfo.Family() == os.WindowsFamily {
 			randomPassword, err := random.NewRandomString(e.Ctx(), e.Namer.ResourceName(name, "win-admin-password"), &random.RandomStringArgs{
 				Length:  pulumi.Int(20),
-				Special: pulumi.Bool(true),
+				Special: pulumi.Bool(false),
```
pgimalac, line 67: You could use [OverrideSpecial](https://www.pulumi.com/registry/packages/random/api-docs/randomstring/#overridespecial_go) if the goal is just not to have dollar signs, but I guess removing special characters entirely works too !

pgimalac (approved):

scenarios/aws/ec2/vm.go
KevinFairise2: Yeah I do not have the full list of invalid characters, but `{ }` are an issue as well. So it is safer to just not put any special character

julien-lebot: For the record, `$`, `{`, and `}` are valid password special characters, the actual issue here is that the password was enclosed in double-quotes instead of single-quotes: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.4

PR #1147, opened 2024-10-03
Patch docker daemon to not use ddbuild runner IP range

KevinFairise2:
What does this PR do?
---------------------

Patch `daemon.json` config to not use the default `172.18.0.0/16` IP range that is conflicting with mudkip-f cluster ip range https://github.com/DataDog/k8s-platform-resources/blob/943d780e5db9eb0b6dae5d547732fe8622126df0/values/us1.ddbuild.io/mudkip-f-k8s/values.yaml
And that is now used to run Gitlab runners: https://github.com/DataDog/dd-source/pull/135523
 More context #incident-31160

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/docker/component.go
+++ b/components/docker/component.go
@@ -148,13 +148,28 @@ func (d *Manager) install() (*remote.Command, error) {
 		return nil, err
 	}
 
+	// Patch ip range that docker uses to create its bridge networks
+	// This is to avoid conflicts with other IP ranges used internally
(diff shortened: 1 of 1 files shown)


components/docker/component.go
```diff
@@ -148,13 +148,28 @@ func (d *Manager) install() (*remote.Command, error) {
 		return nil, err
 	}
 
+	// #incident-31160: Patch ip range that docker uses to create is bridge networks
```
vboulineau, line 151: I think you can just state it conflicts with other private ip usage, it's still open source code.

pducolin (approved):

PR #1167, opened 2024-10-08
[task] default yes in destroy task

pducolin:
What does this PR do?
---------------------

Default destroy without asking for confirmation on destroy task

-------------------

All

Motivation
----------

Personally I always want to destroy the stack I have in my state, no need to confirm

Additional Notes
----------------


tasks/destroy.py
```diff
@@ -17,7 +17,7 @@ def destroy(
     scenario_name: str,
     config_path: Optional[str] = None,
     stack: Optional[str] = None,
-    force_yes: Optional[bool] = False,
+    force_yes: Optional[bool] = True,
```
KevinFairise2, line 20: What about we just remove that params then? I would be surprised that someone ever use the `--no-force-yes` flag

pducolin: Done!

PR #1173, opened 2024-10-09
Update EKS config map to allow account-admin and read-only SSO roles

KevinFairise2:
What does this PR do?
---------------------

It allows our roles to access clusters that are created in the CI. By default only the role that created the cluster can access it. This is a bit blocking when you want to inspect clusters created by the CI. With that change `account-admin` role will always get full access to the EKS cluster and `read-only` will be able to view resources inside the cluster.
It also "repair" the console so that we can see the resources directly in the `Resources` panel on the EKS page

-------------------

Motivation
----------

Additional Notes
----------------


pducolin (approved):

vboulineau:
FYI, we should move to the IAM way, ConfigMap way is legacy and is probably going to be deprecated at some point in the future.

KevinFairise2:
Good to know, was not sure if we chose the Configmap authentication on purpose.
I will create a card for us to migrate to authentication with EKS API

PR #979, opened 2024-07-25
enable ecs collection in all e2e tests

kangyili:
What does this PR do?
---------------------

We’ve added ECS collection to the orchestration check, and we want to enable it for agents in the e2e tests. 

- https://github.com/DataDog/datadog-agent/pull/22060
- https://github.com/DataDog/datadog-agent/pull/21984

We want to broadly enable the collection internally before releasing it to the public.

Which scenarios this will impact?
-------------------

Motivation
----------

Additional Notes
----------------

PR #1181, opened 2024-10-10
[ecs] Enable process collection on Fargate

kkhor-datadog:
What does this PR do?
---------------------
Adds an environment variable config to agents running on Fargate to enable process collection. Tasks are required to have `pidMode: task` set for process collection to work, which was done in https://github.com/DataDog/test-infra-definitions/pull/1063

Also adds a fix to make the ECS cluster connection string display correctly after creating an ECS cluster with the invoke task.

Reference: https://github.com/DataDog/test-infra-definitions/pull/979

-------------------
ECS Fargate

Motivation
----------
Enable process collection on ECS Fargate

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/ecsFargate.go
+++ b/components/datadog/agent/ecsFargate.go
@@ -44,6 +44,10 @@ func ECSFargateLinuxContainerDefinition(e config.Env, image string, apiKeySSMPar
 				Name:  pulumi.StringPtr("DD_ECS_TASK_COLLECTION_ENABLED"),
 				Value: pulumi.StringPtr("true"),
 			},
+			ecs.TaskDefinitionKeyValuePairArgs{
+				Name:  pulumi.StringPtr("DD_PROCESS_CONFIG_PROCESS_COLLECTION_ENABLED"),
+				Value: pulumi.StringPtr("true"),
+			},
 		}, ecsFakeintakeAdditionalEndpointsEnv(fakeintake)...), ecsAgentAdditionalEnvFromConfig(e)...),
 		Secrets: ecs.TaskDefinitionSecretArray{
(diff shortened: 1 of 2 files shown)


pducolin:
Thank you for the fix to ecs.py!

pducolin (approved):

components/datadog/agent/ecsFargate.go
```diff
@@ -44,6 +44,10 @@ func ECSFargateLinuxContainerDefinition(e config.Env, image string, apiKeySSMPar
 				Name:  pulumi.StringPtr("DD_ECS_TASK_COLLECTION_ENABLED"),
 				Value: pulumi.StringPtr("true"),
 			},
+			ecs.TaskDefinitionKeyValuePairArgs{
+				Name:  pulumi.StringPtr("DD_PROCESS_CONFIG_PROCESS_COLLECTION_ENABLED"),
```
pducolin, line 48: **❓ question**
Should this be enabled by default? Should it be allowed to be disabled? I guess we can do it through the agent env vars

pducolin (approved):

components/datadog/agent/ecsFargate.go
kkhor-datadog: We'd want this to be enabled by default, for consistency with the ECS EC2 agent:
https://github.com/DataDog/test-infra-definitions/blob/7e55b9e3279af35d88e4965820114d00e659942b/components/datadog/agent/ecs.go#L134-L137

For some context, this came up when launching an ECS cluster with the invoke task and noticing that the ECS EC2 tasks in the cluster were having their processes collected, but the Fargate tasks in the cluster weren't

PR #1182, opened 2024-10-11
Create an EKS cluster ready to use component

KevinFairise2:
What does this PR do?
---------------------

Create a component for creating an EKS cluster. With a simple API to begin with

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/eks/cluster.go
+++ b/scenarios/aws/eks/cluster.go
@@ -1,0 +1,279 @@
+package eks
+
+import (
+	"github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+	"github.com/DataDog/test-infra-definitions/components"
+	kubecomp "github.com/DataDog/test-infra-definitions/components/kubernetes"
+	"github.com/DataDog/test-infra-definitions/resources/aws"
+	localEks "github.com/DataDog/test-infra-definitions/resources/aws/eks"
+	"github.com/pulumi/pulumi-aws/sdk/v6/go/aws/ec2"
+	awsEks "github.com/pulumi/pulumi-aws/sdk/v6/go/aws/eks"
+	awsIam "github.com/pulumi/pulumi-aws/sdk/v6/go/aws/iam"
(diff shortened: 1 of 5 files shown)


scenarios/aws/eks/cluster.go
```diff
@@ -0,0 +1,279 @@
+package eks
+
+import (
+	"github.com/DataDog/test-infra-definitions/common/config"
+	"github.com/DataDog/test-infra-definitions/common/utils"
+	"github.com/DataDog/test-infra-definitions/components"
+	kubecomp "github.com/DataDog/test-infra-definitions/components/kubernetes"
+	"github.com/DataDog/test-infra-definitions/resources/aws"
+	localEks "github.com/DataDog/test-infra-definitions/resources/aws/eks"
```
pducolin, line 9: **🥜 nitpick**
Would rename this to `awseks`, at first I thought it was referencing local kind

tasks/aws/eks.py
```diff
@@ -90,7 +90,7 @@ def create_eks(
 
 def _show_connection_message(ctx: Context, full_stack_name: str, config_path: Optional[str]):
     outputs = tool.get_stack_json_outputs(ctx, full_stack_name)
-    kubeconfig_output = json.loads(outputs["dd-Cluster-aws-eks"]["kubeConfig"])
+    kubeconfig_output = json.loads(outputs["dd-Cluster-eks"]["kubeConfig"])
```
pducolin, line 93: **👏 praise**

pducolin (approved):
Approving, a double check by the containers team would be better

scenarios/aws/eks/cluster.go
KevinFairise2: Agree, I just copy pasted what existed, but that looks better to change it to `awseks`

KevinFairise2: Well actually we already have another package named `awseks` refering to `github.com/pulumi/pulumi-aws/sdk/v6/go/aws/eks` . Here `localeks` refer to resource defined in the local repository (`test-infra-definitions`)

vboulineau (approved):

PR #1189, opened 2024-10-15
Fix GKE authentication in CI

KevinFairise2:
What does this PR do?
---------------------

Activate service account in the CI and prompt for gcloud auth login when running locally. Otherwise the connection to GKE cluster cannot be made

-------------------

Motivation
----------

Additional Notes
----------------


FlorentClarret (approved):
Sorry I thought I had approved during the first review. 

Thanks!

PR #1204, opened 2024-10-18
Fix unit tests for yaml merge functions

KevinFairise2:
What does this PR do?
---------------------

Fix bug in unit tests. Tests were always passing because we compared the result of `Unmarshall` function that return an error and not the result of the Unmarshalled object

Modified one of the test fixture to show that it works with two format of YAML list

-------------------

Motivation
----------

Additional Notes
----------------


common/utils/yaml_test.go
```diff
@@ -80,10 +79,42 @@ func TestMergeYAMLWithSlices(t *testing.T) {
...
 }
+
+func assertMapsEqual(t *testing.T, expected, actual map[string]interface{}) {
+	assert.Equal(t, len(expected), len(actual), "expected %v, got %v", expected, actual)
+	for k, v := range expected {
+		switch v.(type) {
+		case map[string]interface{}:
+			assertMapsEqual(t, v.(map[string]interface{}), actual[k].(map[string]interface{}))
+		case []interface{}:
+			assertSlicesEqual(t, v.([]interface{}), actual[k].([]interface{}))
```
pducolin, line 98: **💬 suggestion**
```suggestion
	assert.ElementsMatch(t, v.([]interface{}), actual[k].([]interface{}))
```
should work https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.ElementsMatch

common/utils/yaml_test.go
```diff
@@ -80,10 +79,42 @@ func TestMergeYAMLWithSlices(t *testing.T) {
 
 			var gotYAML map[string]interface{}
 			var expectedYAML map[string]interface{}
-
-			gotMap := yaml.Unmarshal([]byte(got), &gotYAML)
-			expectedMap := yaml.Unmarshal([]byte(tc.expectedResult), &expectedYAML)
-			assert.Equal(t, gotMap, expectedMap, "expected %v, got %v", expectedMap, gotMap)
+			err = yaml.Unmarshal([]byte(got), &gotYAML)
```
pducolin, line 82: **❓ question**
Have you tried https://pkg.go.dev/github.com/stretchr/testify/assert#Assertions.YAMLEq?
It would be:
```suggestion
			assert.YAMLEq(t, tc.expectedResult, got)
```

common/utils/yaml_test.go
KevinFairise2: I'll take that one!

common/utils/yaml_test.go
KevinFairise2: I just tried, and that does not work for our use case. Because in YAML
```
tags:
  - a
  - b
```
and
```
tags:
  - b
  - a
 ```
are not considered equal. Because they are list and the order is important. In our case we do not care about the order, so we cannot really use `assert.YAMLEq` directly

But using `assert.ElementsMatch` already simplify the code! Thanks!

pducolin (approved):

PR #1207, opened 2024-10-18
[vm] disable unattended upgrades on debian/ubuntu on EC2

pducolin:
What does this PR do?
---------------------

Disable apt unattended upgrades at startup through userdata

-------------------

aws debian and ubuntu

Motivation
----------

[ADXT-686](https://datadoghq.atlassian.net/browse/ADXT-686)

Additional Notes
----------------

[ADXT-686]: https://datadoghq.atlassian.net/browse/ADXT-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scenarios/aws/ec2/vm.go
+++ b/scenarios/aws/ec2/vm.go
@@ -142,8 +142,9 @@ func defaultVMArgs(e aws.Environment, vmArgs *vmArgs) error {
 		if err != nil {
 			return err
 		}
-
 		vmArgs.userData = vmArgs.userData + sshUserData
+	} else if vmArgs.osInfo.Flavor == os.Ubuntu || vmArgs.osInfo.Flavor == os.Debian {
+		vmArgs.userData = vmArgs.userData + os.DebianDisableUnattendedUpgradesScriptContent
 	}
 
 	return nil
--- a/components/os/debian_disable_upgrades.go
+++ b/components/os/debian_disable_upgrades.go
@@ -1,0 +1,8 @@
+package os
+
+import (
+	_ "embed"
+)
+
+//go:embed scripts/debian-disable-unattended-upgrades.sh
+var DebianDisableUnattendedUpgradesScriptContent string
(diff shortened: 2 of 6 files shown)


scenarios/aws/ec2/vm.go
```diff
@@ -142,8 +142,9 @@ func defaultVMArgs(e aws.Environment, vmArgs *vmArgs) error {
 		if err != nil {
 			return err
 		}
-
 		vmArgs.userData = vmArgs.userData + sshUserData
+	} else if vmArgs.osInfo.Flavor == os.Ubuntu || vmArgs.osInfo.Flavor == os.Debian {
+		vmArgs.userData = vmArgs.userData + os.DebianDisableUnattendedUpgradesScriptContent
```
KevinFairise2, line 147: Does it really work if `vmArgs.userData` already contains something?
I feel like we could end with a script that looks like:
```
#!/bin/bash
echo "First userdata"
#!/bin/bash
apt-get -y remove unattended-upgrades
```
And I am not sure it would work

pducolin: TIL `#!` is called [shebang](https://en.wikipedia.org/wiki/Shebang_%28Unix%29). A script starting with shebang that is used as if it was an executable will use the interpreter defined after the shebang to interpret commands.

Following shebang are discarded as comments.

Tried running the following on zsh 

```
#!/bin/bash
echo "First userdata"
#!/bin/bash
echo "something else"
```

And I get 

```
First userdata
something else
```

KevinFairise2: As long as we provide user data with same language it should work. If we start to pass bash script and python script it will break but that is unlikely we do that

KevinFairise2 (approved):

scenarios/aws/ec2/vm.go
KevinFairise2: We could also enforce bash scripting and automatically set the shebang to bash.

pducolin: We cannot run python scripts from userdata, it supports shell scripts only

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/user-data.html

But we could remove any shebang and enforce ours, removing any line starting with `#!`

PR #1239, opened 2024-11-07
Handle naming for kind the same way as for EKS

KevinFairise2:
What does this PR do?
---------------------

Naming was a mess on kind. Furthermore there is a trick on kind because the agent cannot determine alone the name of the cluster. A new flag allows us to pass the cluster name to the agent from the Kubernetes cluster output so it remains coherent

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/kubernetesagentparams/params.go
+++ b/components/datadog/kubernetesagentparams/params.go
@@ -68,6 +68,8 @@ func NewParams(options ...Option) (*Params, error) {
 }
 
 // WithClusterName sets the name of the cluster. Should only be used if you know what you are doing. Must no be necessary in most cases.
+// Mainly used to set the clusterName when the agent is installed on Kind clusters. Because the agent is not able to detect the cluster name.
+// It takes a pulumi.StringOutput as input to be able to use the pulumi output of the cluster name.
 func WithClusterName(clusterName pulumi.StringOutput) func(*Params) error {
 	return func(p *Params) error {
 		values := pulumi.Sprintf(`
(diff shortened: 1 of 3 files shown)


components/datadog/kubernetesagentparams/params.go
```diff
@@ -67,6 +67,21 @@ func NewParams(options ...Option) (*Params, error) {
 	return common.ApplyOption(version, options)
 }
 
+// WithClusterName sets the name of the cluster. Should only be used if you know what you are doing. Must no be necessary in most cases.
```
pducolin, line 70: **💬 suggestion**
Could help adding more context, for example with a use case where this is used

pducolin (approved):

components/kubernetes/kind.go
```diff
@@ -27,10 +27,10 @@ const (
 var kindClusterConfig string
 
 // Install Kind on a Linux virtual machine.
-func NewKindCluster(env config.Env, vm *remote.Host, resourceName, kindClusterName string, kubeVersion string, opts ...pulumi.ResourceOption) (*Cluster, error) {
-	return components.NewComponent(env, resourceName, func(clusterComp *Cluster) error {
+func NewKindCluster(env config.Env, vm *remote.Host, name string, kubeVersion string, opts ...pulumi.ResourceOption) (*Cluster, error) {
+	return components.NewComponent(env, name, func(clusterComp *Cluster) error {
+		kindClusterName := env.CommonNamer().DisplayName(49) // We can have some issues if the name is longer than 50 characters
```
pducolin, line 32: **❓ question**
What is the resulting name for the cluster, using common namer?

KevinFairise2: The resulting name will be the name of the stack, with a limit of 49 characters here. We have something similar for the naming of the EKS cluster: https://github.com/DataDog/test-infra-definitions/blob/main/scenarios/aws/eks/cluster.go#L88

PR #1241, opened 2024-11-08
[aws][ec2] add  support for imds2 in ec2 resource

pducolin:
What does this PR do?
---------------------

This PR adds the possibility to enable/disable IMDSv1 when creating an ec2 instance.
It adds the HttpToken setting for metadata option for ec2 instance. If http tokens are required then every call to IMDSv1 will fail since only IMDSv2 requires http token to perform.

-------------------

This parameter force the usage of IMDSv2 or the usage of the OS Hostname instead of the instance ID

Motivation
----------

This will help IMDSv2 tests before the transition between IMDSv1 and IMDSv2

Additional Notes
----------------

### DIFF
--- a/resources/aws/ec2/vm.go
+++ b/resources/aws/ec2/vm.go
@@ -21,11 +21,19 @@ type InstanceArgs struct {
 	InstanceProfile string
 
 	// Optional
-	UserData string
+	UserData           string
+	HTTPTokensRequired bool
 }
 
 func NewInstance(e aws.Environment, name string, args InstanceArgs, opts ...pulumi.ResourceOption) (*ec2.Instance, error) {
 	defaultInstanceArgs(e, &args)
+
+	var metadataOptions *ec2.InstanceMetadataOptionsArgs
+	if args.HTTPTokensRequired {
+		metadataOptions = &ec2.InstanceMetadataOptionsArgs{
+			HttpTokens: pulumi.String("required"),
+		}
+	}
 
 	instance, err := ec2.NewInstance(e.Ctx(), e.Namer.ResourceName(name), &ec2.InstanceArgs{
 		Ami:                     pulumi.StringPtr(args.AMI),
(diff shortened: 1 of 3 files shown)


resources/aws/ec2/vm.go
```diff
@@ -44,7 +52,9 @@ func NewInstance(e aws.Environment, name string, args InstanceArgs, opts ...pulu
 			"Name": e.Namer.DisplayName(255, pulumi.String(name)),
 		},
 		InstanceInitiatedShutdownBehavior: pulumi.String(e.DefaultShutdownBehavior()),
+		MetadataOptions:                   ec2.InstanceMetadataOptionsPtr(metadataOptions),
```
ogaca-dd, line 55: Does this line behave as expected when metadataOptions is `nil`?

scenarios/aws/ec2/vmargs.go
```diff
@@ -81,3 +83,10 @@ func WithInstanceProfile(instanceProfile string) VMOption {
 		return nil
 	}
 }
+
```
ogaca-dd, line 86: I think it makes sense to add some documentation for this function as the name of the field in ec2.MetadataOptions is not obvious.

ogaca-dd (approved):

resources/aws/ec2/vm.go
louis-cqrl: if the arg is nil then the function return nil

pducolin: By default `metadataOptions` is `nil`

scenarios/aws/ec2/vmargs.go
pgimalac: I agree something like `WithEnableIMDS2` would be clearer

resources/aws/ec2/vm.go
```diff
@@ -21,12 +21,20 @@ type InstanceArgs struct {
 	InstanceProfile string
 
 	// Optional
-	UserData string
+	UserData           string
+	HTTPTokensRequired bool
```
pgimalac, line 25: By default this is false, but I think it makes sense to have it enabled by default since it's the case on AWS (or it will be very soon)

pgimalac: https://aws.amazon.com/fr/blogs/aws/amazon-ec2-instance-metadata-service-imdsv2-by-default/
> Mid-2024 – Newly released Amazon EC2 instance types will use IMDSv2 only by default.

pgimalac (approved):

pgimalac (approved):

pducolin (commented):
All good for me!

PR #1246, opened 2024-11-14
[CONTP-437] Add check_cardinality to nginx

zhuminyi:
What does this PR do?
---------------------
A follow up for recent feature in 7.60: https://github.com/DataDog/datadog-agent/pull/29984
which allows user to configure the cardinality of check instance via annotation check config.

This PR set nginx check cardinality to high.

-------------------
New test metric "nginx.net.connections" will be added https://github.com/DataDog/datadog-agent/pull/35995
for testing additional tags like container id, name etc. 

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/apps/nginx/k8s.go
+++ b/components/datadog/apps/nginx/k8s.go
@@ -93,7 +93,7 @@ func K8sAppDefinition(e config.Env, kubeProvider *kubernetes.Provider, namespace
 							map[string]interface{}{
 								"nginx": map[string]interface{}{
 									"init_config":           map[string]interface{}{},
-									"check_tag_cardinality": "low",
+									"check_tag_cardinality": "high",
 									"instances": []map[string]interface{}{
 										{
 											"nginx_status_url": "http://%%host%%/nginx_status",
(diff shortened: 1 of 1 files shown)


chouetz (approved):

components/datadog/apps/nginx/k8s.go
```diff
@@ -92,7 +92,8 @@ func K8sAppDefinition(e config.Env, kubeProvider *kubernetes.Provider, namespace
 						"ad.datadoghq.com/nginx.checks": pulumi.String(utils.JSONMustMarshal(
 							map[string]interface{}{
 								"nginx": map[string]interface{}{
-									"init_config": map[string]interface{}{},
+									"init_config":           map[string]interface{}{},
+									"check_tag_cardinality": "high",
```
L3n41c, line 96: This will have no effect because the checks cardinality is already set to `high` globally in agent’s config: https://github.com/DataDog/test-infra-definitions/blob/d610d4c9aa4be9f5ef0b06b2ec8dd56aa508c055/components/datadog/agent/kubernetes_helm.go#L252

In order to validate the new `check_tag_cardinality` feature, we would need another value, different from the “global” one.

PR #1283, opened 2024-12-09
[ADXT-512] Disable dual shipping by default

CelianR:
What does this PR do?
---------------------

Disables dual shipping by default and renames the option.

See https://github.com/DataDog/datadog-agent/pull/31873 for implementation

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/ec2/vm_run.go
+++ b/scenarios/aws/ec2/vm_run.go
@@ -41,6 +41,10 @@ func VMRun(ctx *pulumi.Context) error {
 				fakeIntakeOptions = append(fakeIntakeOptions, fakeintake.WithLoadBalancer())
 			}
 
+			if env.AgentUseDualShipping() {
+				fakeIntakeOptions = append(fakeIntakeOptions, fakeintake.WithoutDDDevForwarding())
+			}
+
 			fakeintake, err := fakeintake.NewECSFargateInstance(env, vm.Name(), fakeIntakeOptions...)
 			if err != nil {
 				return err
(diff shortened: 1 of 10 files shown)


pducolin (approved):

scenarios/aws/ec2/vm_run.go
```diff
@@ -112,6 +116,10 @@ func VMRunWithDocker(ctx *pulumi.Context) error {
 				fakeIntakeOptions = append(fakeIntakeOptions, fakeintake.WithLoadBalancer())
 			}
 
+			if env.AgentUseDualShipping() {
```
KevinFairise2, line 119: Agent Dual shipping was an option available only on Kubernetes. If we add that, passing the flag `ddagent:dualshipping` would only deactivate the fakeintake payload forwarding because agent dual shipping option is not offered on non-kubernetes agent. Which would be misleading because we end up in a situation where nothing is sent to dddev anymore

scenarios/aws/ec2/vm_run.go
```diff
@@ -41,6 +41,10 @@ func VMRun(ctx *pulumi.Context) error {
...
+				fakeIntakeOptions = append(fakeIntakeOptions, fakeintake.WithoutDDDevForwarding())
+			}
```
KevinFairise2, line 46: We can remove that one as well, we cannot configure dual shipping for the agent on VM

KevinFairise2 (approved):

PR #1290, opened 2024-12-12
group aws and azure native dependencies

pducolin:
What does this PR do?
---------------------

Group together aws-sdk and pulumi-azure-native-sdk dependencies

-------------------

None

Motivation
----------

Prevent pulumi plugins mismatch

Additional Notes
----------------

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -85,3 +85,16 @@ updates:
     directory: /components/datadog/apps/cws/images/cws-centos7
     schedule:
       interval: daily
+
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -85,3 +85,16 @@ updates:
...
       interval: daily
+
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      aws-sdk-go-v2:
+        patterns:
+          - '^github\.com\/aws\/aws-sdk-go-v2'
```
ofek, line 96: I think patterns must be globs rather than regular expressions

pducolin: Thank you, fixed now

KevinFairise2 (approved):

PR #1294, opened 2024-12-13
Fix podman/docker config calling ddtool

CelianR:
What does this PR do?
---------------------

When executing `inv localpodman.create-vm`, it was possible to obtain a timeout on vault with ddtool hook.
Replaced the config by a custom empty config.

> [!NOTE]
> Do we want to be able to customize the config? Maybe it is better to provide an argument to replace by a custom config

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/resources/local/podman/vm.go
+++ b/resources/local/podman/vm.go
@@ -20,7 +20,7 @@ type VMArgs struct {
 
 //go:embed data/Dockerfile
 var dockerfileContent string
-var customDockerConfig string = "{}"
+var customDockerConfig = "{}"
 
 func NewInstance(e resourceslocal.Environment, args VMArgs, opts ...pulumi.ResourceOption) (address pulumi.StringOutput, user string, port int, err error) {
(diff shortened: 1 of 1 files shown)


resources/local/podman/vm.go
```diff
@@ -49,14 +50,22 @@ func NewInstance(e resourceslocal.Environment, args VMArgs, opts ...pulumi.Resou
 		return pulumi.StringOutput{}, "", -1, err
 	}
 
+	// Use a config to avoid docker hooks that can call vault or other services (credHelpers)
+	err = os.WriteFile(path.Join(dataPath, "config.json"), []byte(customDockerConfig), 0600)
+	if err != nil {
+		return pulumi.StringOutput{}, "", -1, err
+	}
+
+	podmanCommand := "podman --config " + dataPath
```
pducolin, line 59: **👏 praise**
Oh nice!!

pducolin (approved):

PR #1306, opened 2024-12-16
[gitlab] add manual trigger for bump in datadog-agent

pducolin:
What does this PR do?
---------------------

* Add manual job to trigger a bump of test-infra-definitions in `datadog-agent` from dev branches
* Sync and tidy all go modules in `datadog-agent` 

-------------------

None

Motivation
----------

We often want to check that changes here do not break or fix tests in `datadog-agent`, this should reduce the manual operations to do this

Syncing and tidying all go modules as now `datadog-agent` uses go workspace, thus a bump in `test/new-e2e` might bump deps in other `datadog-agent` modules

Additional Notes
----------------

Requires https://github.com/DataDog/datadog-agent/pull/32245

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -100,7 +100,21 @@ release-runner-image:
     - if: $CI_COMMIT_BRANCH == "main"
       when: on_success
 
-.bump-version-on-datadog-agent:
+bump-version-on-datadog-agent:
+  stage: post-release
+  image: registry.ddbuild.io/ci/datadog-agent-buildimages/deb_x64:v48815877-9bfad02c
+  tags: ["arch:amd64"]
+  rules:
+    - if: $CI_COMMIT_BRANCH == "main"
+      needs: ["release-runner-image"]
+      when: on_success
+    - if: $CI_COMMIT_BRANCH =~ /^mq-working-branch-/
+      when: never
+    - when: manual
+      needs: ["build-runner-image"]
+      allow_failure: true
+      variables:
+        EXTRA_UPDATE_ARGS: "--is-dev-image"
   variables:
     EXTRA_UPDATE_ARGS: ""
   before_script:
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -126,10 +121,36 @@ bump-version-on-datadog-agent:
...
+
+bump-version-on-datadog-agent:
+  extends: .bump-version-on-datadog-agent
+  stage: post-release
+  image: ${CI_REGISTRY_IMAGE_TEST}:${CI_COMMIT_SHORT_SHA}
+  tags: ["arch:amd64"]
+  needs: ["release-runner-image"]
+  rules:
+    - if: $CI_COMMIT_BRANCH == "main"
+      when: on_success
```
KevinFairise2, line 155: What about we keep a single job that is automatic on main and manual on dev branch?
```suggestion
bump-version-on-datadog-agent:
  extends: .bump-version-on-datadog-agent
  stage: post-release
  image: ${CI_REGISTRY_IMAGE_TEST}:${CI_COMMIT_SHORT_SHA}
  tags: ["arch:amd64"]
  needs: ["release-runner-image"]
  rules:
    - if: $CI_COMMIT_BRANCH == "main"
      when: on_success
    - when: manual
       allow_failure: true
       variables:
         EXTRA_UPDATE_ARGS: "--is-dev-image"
```

pducolin: Thought about this, but they would have different `needs`, as the `main` one needs the release image, while the dev one needs the test image. Could not find how to set needs based on rules

KevinFairise2: Updated the suggestion! But that should work as well with needs in rules
Per https://docs.gitlab.com/ee/ci/yaml/#rulesneeds

pducolin: I'm also removing the private job definition, no need for it anymore

pducolin: Done!

KevinFairise2 (approved):

pducolin:
Failing as I changed the image and it does not have `jq` anymore, now I know why we use the test-infra image

PR #1310, opened 2024-12-18
AMLII-2213 Add FIPS agent image option

vickenty:
What does this PR do?
---------------------

Add option to request FIPS-flavored agent image

-------------------

New e2e JMXFetch tests (TBD) using FIPS JMX agent images

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/docker.go
+++ b/components/datadog/agent/docker.go
@@ -129,13 +129,11 @@ func dockerAgentComposeManifest(agentImagePath string, apiKey pulumi.StringInput
 func defaultAgentParams(params *dockeragentparams.Params) {
 	// After setting params.FullImagePath check if you need to use JMX Docker image
 	defer func(p *dockeragentparams.Params) {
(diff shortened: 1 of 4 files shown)


components/datadog/agent/docker.go
```diff
@@ -133,6 +133,11 @@ func defaultAgentParams(params *dockeragentparams.Params) {
 			p.FullImagePath = fmt.Sprintf("%s-jmx", p.FullImagePath)
 		}
 	}(params)
+	defer func(p *dockeragentparams.Params) {
+		if p.FIPS {
+			p.FullImagePath += "-fips"
+		}
+	}(params)
```
pducolin, line 140: **❓ question**
How does this work if both jmx and fips are set?

pducolin: What about having them in a single defer function? Might help with readability

pducolin (approved):

PR #1342, opened 2025-01-06
[task][vm] add remote host to ssh known hosts

pducolin:
What does this PR do?
---------------------

Add newly created host to `known_hosts` file at create-vm time

-------------------

VMs

Motivation
----------

Required for ADXT-690, as we want less interactions as possible to quickly set up a remote VM for windows development

Additional Notes
----------------

Tested it locally:

1. Create a VM on aws
```bash
inv aws.create-vm --no-install-agent
```
2. Verified that I could connect without being prompted to add the remote host to known hosts, and that `~/.ssh/known_hosts` contained my new host
```
# known_hosts content
...
10.1.57.39 ssh-rsa <public key>
10.1.57.39 ecdsa-sha2-nistp256 <public key>
10.1.57.39 ssh-ed25519 <public key>
```
3. Destroyed the VM
```bash
inv aws.destroy-vm
```
4. Verified that `~/.ssh/known_hosts` did not contain my remote host anymore
```
# known_hosts content
...
```


tasks/tool.py
```diff
@@ -262,15 +262,27 @@ def show_connection_message(
...
 
 
-def clean_known_hosts(host: str) -> None:
+def add_known_host(ctx: Context, host: str) -> None:
+    """
+    Add the host to the known_hosts file.
+    """
+    # remove the host if it already exists
+    clean_known_hosts(ctx, host)
+    home = os.environ.get("HOME", f"/Users/{getpass.getuser()}")
```
KevinFairise2, line 271: Maybe we can use 
```
from pathlib import Path
home = Path.home()
```
instead? So we do not need to default on Windows path when the env variable is not set

tasks/tool.py
```diff
@@ -262,15 +262,27 @@ def show_connection_message(
...
+    ctx.run(f"ssh-keyscan -H {host} >> {home}/.ssh/known_hosts")
+
+
+def clean_known_hosts(ctx: Context, host: str) -> None:
     """
     Remove the host from the known_hosts file.
     """
+    ctx.run(f"ssh-keygen -R {host}")
```
KevinFairise2, line 279: Isn't it enough to remove the key from known_host? When I run that locally it removes the host

pducolin: Here is an example of what I get when running the `create-vm` task, `ssh-keyscan` adds the following lines to `known_hosts`

```
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
10.1.59.88 ssh-rsa <public key>
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
10.1.59.88 ecdsa-sha2-nistp256 <public key>
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
10.1.59.88 ssh-ed25519 <public key>
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
```

Note the headers. They are sent to stderr.
Running `ssh-keygen -R 10.1.58.184` leaves the comments in `known_hosts`

```
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
# 10.1.59.88:22 SSH-2.0-OpenSSH_8.9p1 Ubuntu-3ubuntu0.10
```

Thinking I could avoid that by getting the output from `ctx`, then writing stdout only to `known_files`. This would avoid calling the extra step to clean up comment headers from `known_hosts` at clean up time

KevinFairise2: Oh yeah I do not have any comment in my current `known_hosts` file so I missed them.
But I think that indeed it would make things easier if we do not put the comment in the `known_hosts` file. And I feel that the fact it is sent to stderr means that we should not keep it

KevinFairise2 (approved):

PR #1351, opened 2025-01-10
[ADXT-815] [e2e] filepath.Join should be agnostic of the runner OS

KevinFairise2:
What does this PR do?
---------------------

Some operations currently uses the runner OS information to determine the separator. It does not make sense as the commands are excuted on the target OS. This PR updates that so the commands are executed depending on the target OS

-------------------

Motivation
----------

Additional Notes
----------------


clarkb7 (approved):

components/command/unixOSCommand.go
```diff
@@ -121,7 +125,7 @@ func (fs unixOSCommand) copyLocalFile(runner *LocalRunner, name string, src, dst
 
 func (fs unixOSCommand) copyRemoteFile(runner *RemoteRunner, name string, src, dst pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error) {
 	tempRemotePath := src.ToStringOutput().ApplyT(func(path string) string {
-		return filepath.Join(runner.OsCommand().GetTemporaryDirectory(), filepath.Base(path))
```
chouetz, line 124: Is it the only occurrence we needed to change?

chouetz (approved):

components/command/unixOSCommand.go
KevinFairise2: Yep in our code. There are other occurences but in KMT specific which I think there is no chance it can be ran from a Windows laptop

PR #1364, opened 2025-01-16
[ADXT-84] Add create-vm option to set agent config

CelianR:
What does this PR do?
---------------------

Adds an option that takes a path to set the agent config.
Note that this won't override the default config with the api key but both will be merged together.

-------------------

Motivation
----------

Additional Notes
----------------


tasks/aws/ecs.py
```diff
@@ -68,6 +70,7 @@ def create_ecs(
         extra_flags=extra_flags,
         full_image_path=full_image_path,
         agent_flavor=agent_flavor,
+        agent_config_path=agent_config_path,
```
KevinFairise2, line 73: I think we do not want it for ECS, ECS should rather be configured with environment variables. And AFAIU this parameter does not do anything in the current code?

KevinFairise2 (approved):

PR #1365, opened 2025-01-16
[kind] mount host /sys into kind node

YoannGh:
What does this PR do?
---------------------

Similarly to #774, the system-probe container needs access to the host /sys filesystem to retrieve cgroup data of processes. This PR hence updates the kind node configuration to mount the host /sys directory into the control-plane container so that it can be propagated to the system-probe container. 

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/kubernetes/kind-cluster.yaml
+++ b/components/kubernetes/kind-cluster.yaml
@@ -1,10 +1,12 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
-- role: control-plane
-  extraMounts:
-  - hostPath: /proc
-    containerPath: /host/proc
+  - role: control-plane
+    extraMounts:
+      - hostPath: /proc
+        containerPath: /host/proc
+      - hostPath: /sys
+        containerPath: /host/sys
 containerdConfigPatches:
   - |-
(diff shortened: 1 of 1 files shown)


components/kubernetes/kind-cluster.yaml
```diff
@@ -1,10 +1,12 @@
...
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
-- role: control-plane
-  extraMounts:
-  - hostPath: /proc
-    containerPath: /host/proc
+  - role: control-plane
+    extraMounts:
+      - hostPath: /proc
+        containerPath: /host/proc
```
vboulineau, line 7: I will put again my comment from https://github.com/DataDog/test-infra-definitions/pull/774#discussion_r1570246532

With Kind we're not testing with real host proc or host sys as we are not interested in the PIDs and containers not managed by Kind, we never had any issue so far, so I am not sure the use case is valid.

vboulineau: I see, I'm not 100% sure if container monitoring will be impacted or not. If not that's fine and can be merged, although it should be explained around the extra mounts.

PR #1371, opened 2025-01-20
[ADXT-84] Add custom helm config argument for create-vm

CelianR:
What does this PR do?
---------------------

This adds an argument to create-vm to specify an helm config to merge with the default one. This is useful to configure the agent environment variables for instance.

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/common/config/environment.go
+++ b/common/config/environment.go
@@ -56,7 +56,7 @@ const (
 	DDAgentDualShipping                  = "dualshipping"
 	DDAgentSite                          = "site"
 	DDAgentMajorVersion                  = "majorVersion"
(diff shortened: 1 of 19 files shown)


common/config/environment.go
```diff
@@ -33,6 +33,7 @@ const (
 	DDInfraExtraResourcesTags               = "extraResourcesTags"
 	DDInfraSSHUser                          = "sshUser"
 	DDInfraInitOnly                         = "initOnly"
+	DDInfraHelmConfig                       = "helmConfig"
```
KevinFairise2, line 36: Can we put that in the Agent namespace instead? Because this is related to the config of the agent and not the test infrastructure

KevinFairise2 (approved):

PR #1384, opened 2025-01-28
Refactor docker image url creation to allow passing FIPS and JMX flavor

KevinFairise2:
What does this PR do?
---------------------

- Refactor a bit the way we compute Docker image path so we use the same method in every environment that need the agent Docker image.
- Add a way to provide a FIPS flag through config map that will force the framework to use the FIPS version of the agent. This is useful because we want to be able to run all the existing e2e test with the FIPS version of the agent.

Pipeline on datadog-agent with the change: https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/54231255

-------------------

Motivation
----------

Additional Notes
----------------


components/datadog/agent/docker.go
```diff
@@ -127,15 +130,15 @@ func dockerAgentComposeManifest(agentImagePath string, apiKey pulumi.StringInput
...
-	}(params)
+	// // After setting params.FullImagePath check if you need to use JMX Docker image
+	// defer func(p *dockeragentparams.Params) {
+	// 	if p.FIPS {
+	// 		p.FullImagePath += "-fips"
+	// 	}
+	// 	if p.JMX {
+	// 		p.FullImagePath = fmt.Sprintf("%s-jmx", p.FullImagePath)
+	// 	}
+	// }(params)
```
CelianR, line 141: This should be removed I think

CelianR (approved):

PR #1386, opened 2025-01-28
microvms: Use ARP to resolve DHCP leases in macOS

gjulianm:
What does this PR do?
---------------------

This PR adds ARP table querying as an additional source to listen for DHCP leases. In some cases we've seen that the `dhcpd_leases` file contains hardware identifiers that are not MAC addresses  (see [RFC 2132](https://www.rfc-editor.org/rfc/rfc2132#section-9), section 9.14, where it's specified that DHCP client IDs might not always be Ethernet addresses) which means we cannot extract the MAC address for the new VM from there. The ARP table should have information in those cases, as we do have a DHCP lease and the IP-MAC assignment should be present in the table.

-------------------

microvms

Motivation
----------

Additional Notes
----------------

Tested locally in macOS by launching a KMT local instance.

### DIFF
--- a/scenarios/aws/microVMs/microvms/network.go
+++ b/scenarios/aws/microVMs/microvms/network.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -216,7 +218,11 @@ func parseBootpDHCPLeases() ([]dhcpLease, error) {
 				return nil, fmt.Errorf("parseBootpDHCPLeases: invalid hw_address format: %s", hwaddr)
 			}
 
-			parsingLease.mac = parts[1]
+			mac, err := normalizeMAC(parts[1])
+			if err != nil {
+				return nil, fmt.Errorf("parseBootpDHCPLeases: error normalizing MAC address: %w", err)
+			}
+			parsingLease.mac = mac
 		}
 		if line == "}" {
 			leases = append(leases, parsingLease)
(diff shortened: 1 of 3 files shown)


scenarios/aws/microVMs/microvms/network.go
```diff
@@ -227,25 +238,116 @@ func parseBootpDHCPLeases() ([]dhcpLease, error) {
...
+	}
+
+	return dhcpLease{
+		ip:  ip,
+		mac: mac,
+	}, nil
+}
+
+// normalizeMAC normalizes a MAC address to the format XX:XX:XX:XX:XX:XX, in lowercase and with leading zeros
+func normalizeMAC(mac string) (string, error) {
```
brycekahle, line 290: please use `net.ParseMAC()` and `HardwareAddr.String()`

brycekahle (changes requested):

scenarios/aws/microVMs/microvms/network.go
gjulianm: I initially used that, apparently `net.ParseMAC` doesn't like addresses where the digits have no zero padding (e.g., 1:2:3:4:5:6). Technically those are invalid MACs ([section 3.1.8 in the spec](https://www.ieee802.org/secmail/pdfocSP2xXA6d.pdf)) but for some reason it's what BootP writes in that file. I can add the clarification to a comment.

brycekahle: Could you still parse them into a `net.HardwareAddr`? That way you can reuse the `String()` method.

gjulianm: Changed

PR #1444, opened 2025-02-25
Add cilium helm and kind cluster installation

hmahmood:
What does this PR do?
---------------------
Adds a new variant to a remote kind cluster that supports cilium. There is also a new helm installation component for cilium added.

-------------------

Motivation
----------

Additional Notes
----------------


components/kubernetes/kind.go
```diff
@@ -26,8 +29,25 @@ const (
...
+
+var kindCiliumClusterTemplate *template.Template
+
+func init() {
+	var err error
+	kindCiliumClusterTemplate, err = template.ParseFS(kindCilumClusterFS, "kind-cilium-cluster.yaml")
+	if err != nil {
+		panic(err)
+	}
+}
```
KevinFairise2, line 43: Is there a real advantage to put it there instead of in  `NewCiliumKindCluster` function. So that it is not executed when not useful.

components/kubernetes/cilium/cilium.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (p *Params) HasKubeProxyReplacement() bool {
+	if v, ok := p.HelmValues["kubeProxyReplacement"]; ok {
+		return boolValue(v)
+	}
+
+	return false
+}
+
+func NewHelmInstallation(e config.Env, params *Params, opts ...pulumi.ResourceOption) (*HelmComponent, error) {
```
KevinFairise2, line 60: I think it can make sense that `NewHelmInstallation` take a kubernetes cluster as an argument, because it needs information about the cluster to work properly
`func NewHelmInstallation(e config.Env, cluster *componentskube.Cluster, params *Params, opts ...pulumi.ResourceOption)`
That way it is explicit that you need to give the cluster for it to work. And you can have that: https://github.com/DataDog/datadog-agent/pull/34398/files#diff-9dbae0537bf500d77990a2ef3697b49f572cd967cca5f890c84058eea75885b6R133-R134 in this function.

components/kubernetes/kind.go
hmahmood: Moved to `NewCiliumKindCluster`

components/kubernetes/cilium/cilium.go
hmahmood: Added the parameter

PR #1445, opened 2025-02-26
Use patched AMI for Ubuntu 22

KevinFairise2:
What does this PR do?
---------------------

The curl version in Ubuntu seems to be buggy: https://stackoverflow.com/questions/72627218/openssl-error-messages-error0a000126ssl-routinesunexpected-eof-while-readin
It should reduce flakiness in our e2e tests

Images generated with: https://gitlab.ddbuild.io/DataDog/ami-builder/-/pipelines/56995600

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/scenarios/aws/ec2/os_resolver.go
+++ b/scenarios/aws/ec2/os_resolver.go
@@ -141,6 +141,14 @@ func resolveUbuntuAMI(e aws.Environment, osInfo *os.Descriptor) (string, error) 
 		volumeType = "ebs-gp3"
 	}
 
+	// PATCH: Ubuntu 22.04 has a bugged version of curl. Use our own patched AMI instead
+	if osInfo.Version == "22.04" {
+		if paramArch == os.ARM64Arch {
+			return "ami-07e8e5f83dd25f929", nil
+		}
+		return "ami-0c6d6caaaa95d1d0c", nil
+	}
+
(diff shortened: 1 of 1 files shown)


CelianR (approved):

scenarios/aws/ec2/os_resolver.go
```diff
@@ -141,6 +141,14 @@ func resolveUbuntuAMI(e aws.Environment, osInfo *os.Descriptor) (string, error)
 		volumeType = "ebs-gp3"
 	}
 
+	// PATCH: Ubuntu 22.04 has a bugged version of curl. Use our own patched AMI instead
+	if osInfo.Version == "22.04" {
```
vboulineau, line 145: I understand that this is a one-shot image, which is not great to maintain. I also saw the PRs to upgrade to 24.04 reverted, what was the reason, can we work on fixing whatever prevents the upgrade?

KevinFairise2: Yeah that's not really perfect.
However updating to Ubuntu 24.04 as a default seems to break a lot of test on datadog-agent. And we have some interest from Fleet team to keep some test running on Ubuntu 22.04 as well, and in the current situation the upstream Ubuntu 22.04 image brings flakiness in the CI

vboulineau: Can we patch it at runtime perhaps, with a curl tarball stored in our S3, adding to the init script?

PR #1455, opened 2025-02-28
Disabling automatic update should not fail provisioning if it has no effect

KevinFairise2:
What does this PR do?
---------------------

Make sure user script does not fail provisioning if automatic updates are already disabled or the script is not relevant for the platform

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/os/scripts/apt-disable-unattended-upgrades.sh
+++ b/components/os/scripts/apt-disable-unattended-upgrades.sh
@@ -6,7 +6,7 @@ sudo systemctl disable unattented-upgrades.service || true
 sudo systemctl stop unattented-upgrades.service || true
 
 sudo systemctl disable apt-daily.service || true
-sudo systemctl disable apt-daily.time || truer
+sudo systemctl disable apt-daily.time || true
 sudo systemctl stop apt-daily.service || true
 sudo systemctl stop apt-daily.timer || true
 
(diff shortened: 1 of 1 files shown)


components/os/scripts/apt-disable-unattended-upgrades.sh
```diff
@@ -1,2 +1,16 @@
...
-apt-get -y remove unattended-upgrades
\ No newline at end of file
+apt-get -y remove unattended-upgrades
+
+# Try to disable unattended upgrades and apt automatic updates, should not fail if it is not installed
+sudo systemctl disable unattented-upgrades.service || true
+sudo systemctl stop unattented-upgrades.service || true
+
+sudo systemctl disable apt-daily.service || true
+sudo systemctl disable apt-daily.time || truer
```
CelianR, line 9:
```suggestion
sudo systemctl disable apt-daily.time || true
```

components/os/scripts/apt-disable-unattended-upgrades.sh
```diff
@@ -1,2 +1,16 @@
 #!/bin/bash
-apt-get -y remove unattended-upgrades
\ No newline at end of file
+apt-get -y remove unattended-upgrades
+
+# Try to disable unattended upgrades and apt automatic updates, should not fail if it is not installed
```
CelianR, line 4: NIT: You could use `set +e`

CelianR (approved):

components/os/scripts/apt-disable-unattended-upgrades.sh
KevinFairise2: Unfortunately it does not work. Otherwise it should have failed because `set +e` should be the default behavior of a bash script

PR #1456, opened 2025-02-28
Allow installing package from the filesystem

KevinFairise2:
What does this PR do?
---------------------

Add an alternative install method that allows to install a package that is present on the file system. Goal is to give the ability to use locally built packages in the E2E tests.

The new `--local-package` can take a single file or a folder. Giving a folder will be useful if you want to run tests with different OS with a single command.

Note: I updated `CopyFile` to `CopyToRemote` because the former is deprecated and is configured to not use concurrentWrite on the underlying SFTP client, which make the upload about 30 times longer

-------------------

Motivation
----------

Additional Notes
----------------


components/datadog/agent/host.go
```diff
@@ -81,7 +85,78 @@ func (h *HostAgent) installAgent(env config.Env, params *agentparams.Params, bas
...
 		}, baseOpts...)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	return installCmd, nil
+}
+
+func (h *HostAgent) directInstallInstallation(env config.Env, params *agentparams.Params, baseOpts ...pulumi.ResourceOption) (command.Command, error) {
+
```
CelianR, line 94:
```suggestion
```

components/datadog/agent/host.go
```diff
@@ -81,7 +85,78 @@ func (h *HostAgent) installAgent(env config.Env, params *agentparams.Params, bas
...
+	if pathInfo.IsDir() {
+		matches, err = fs.Glob(os.DirFS(params.Version.LocalPath), fmt.Sprintf("*%s", wantedExt))
+		if err != nil {
+			return nil, err
+		}
+		if len(matches) == 0 {
+			return nil, fmt.Errorf("no package found in %s with extension %s", params.Version.LocalPath, wantedExt)
+		}
+		if len(matches) > 1 {
+			env.Ctx().Log.Warn(fmt.Sprintf("Found multiple packages to install, using the first one: %s", matches[0]), nil)
```
CelianR, line 124: Question: In which case is it possible? If it is likely to happen, I think we can sort the matches (if `fs.Glob` doesn't do so) and use the last one instead (I think that the last one would be the latest one?)

CelianR (approved):
LGTM!

components/datadog/agent/host.go
KevinFairise2: It is possible if you give a folder that contains several packages with the same extension. We can sort them alphabetically and take the last but I do not think it would change a lot. The warning is here to indicate that the package that will be used might not be the one you expected. In a general case we should not give a folder containing several packages with the same extension

PR #1467, opened 2025-03-10
Do not use CopyToRemote everywhere

KevinFairise2:
This reverts commit 9bbbec2

What does this PR do?
---------------------

Partially revert a change that was made to use `CopyToRemote` instead of `CopyFile` because it does not support well copying a file when it is not determined at plan time. Let's use `CopyToRemote` only for uploading package file, we know it works there.

I'll open a discussion with pulumi-command maintainer to check if it is something that they are aware of

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/command/osCommand.go
+++ b/components/command/osCommand.go
@@ -34,6 +34,7 @@ type OSCommand interface {
 	NewCopyToRemoteFile(runner Runner, name string, localPath, remotePath pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
 	copyLocalFile(runner *LocalRunner, name string, src, dst pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
 	copyRemoteFile(runner *RemoteRunner, name string, src, dst pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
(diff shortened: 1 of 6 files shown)


components/command/osCommand.go
```diff
@@ -31,8 +31,10 @@ type OSCommand interface {
 	PathJoin(parts ...string) string
 
 	NewCopyFile(runner Runner, name string, localPath, remotePath pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
+	NewCopyToRemoteFile(runner Runner, name string, localPath, remotePath pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
 	copyLocalFile(runner *LocalRunner, name string, src, dst pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
 	copyRemoteFile(runner *RemoteRunner, name string, src, dst pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
+	copyRemoteFileV2(runner *RemoteRunner, name string, src, dst pulumi.StringInput, opts ...pulumi.ResourceOption) (pulumi.Resource, error)
```
CelianR, line 37: NIT: Would it be possible to have a more explicit name / a comment with the difference between both versions?

CelianR (approved):
NIT: I think that the PR name should not be `Revert ...` since it's more than a revert

PR #1499, opened 2025-04-04
Export architecture in host output and add newline

KevinFairise2:
What does this PR do?
---------------------

Export the architecture in the host output
Add a newline at the end of a printed line, to make sure it does not break go test output

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/remote/component.go
+++ b/components/remote/component.go
@@ -13,13 +13,14 @@ type HostOutput struct {
 
 	CloudProvider components.CloudProviderIdentifier `json:"cloudProvider"`
 
-	Address   string    `json:"address"`
-	Port      int       `json:"port"`
-	Username  string    `json:"username"`
-	Password  string    `json:"password,omitempty"`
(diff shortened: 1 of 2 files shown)


components/remote/component.go
```diff
@@ -13,13 +13,14 @@ type HostOutput struct {
...
-	OSFlavor  os.Flavor `json:"osFlavor"`
-	OSVersion string    `json:"osVersion"`
+	Address      string    `json:"address"`
+	Port         int       `json:"port"`
+	Username     string    `json:"username"`
+	Password     string    `json:"password,omitempty"`
+	OSFamily     os.Family `json:"osFamily"`
+	OSFlavor     os.Flavor `json:"osFlavor"`
+	OSVersion    string    `json:"osVersion"`
+	Architecture string    `json:"architecture"`
```
pgimalac, line 23: Why not use os.Architecture ? It's the string type, but more explicit 😄

pgimalac (approved):

PR #1521, opened 2025-05-07
Use pinned version of the test apps docker images

L3n41c:
What does this PR do?
---------------------

Use pinned version of the test apps docker images.

-------------------

All.

Motivation
----------

We need to control when a change done in a test app docker image is used by the `datadog-agent` CI.
With the floating `main` tag, a change merged in `test-infra-definitions` was immediately used by all the branches and PRs of `datadog-agent` without any “bump” PR.

Additional Notes
----------------

### DIFF
--- a/components/datadog/apps/dogstatsd/docker-compose.yaml
+++ b/components/datadog/apps/dogstatsd/docker-compose.yaml
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   dogstatsd-udp:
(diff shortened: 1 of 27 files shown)


ichinaski (approved):

KevinFairise2 (commented):
Great! That's definitely something we want. 
Don't we need to update our jobs that push the images so that they are tagged with the correct App version? Otherwise we will still push `main` tags, don't we?

KevinFairise2 (approved):

PR #1525, opened 2025-05-14
Avoid picking artifact from other pipelines on Windows runner

KevinFairise2:
What does this PR do?
---------------------

Check for the pipeline id in the artifact name for windows. Workaround for CIEXE-143

-------------------

Motivation
----------

Additional Notes
----------------

### DIFF
--- a/components/datadog/agent/package.go
+++ b/components/datadog/agent/package.go
@@ -14,7 +14,7 @@ import (
 // GetPackagePath retrieve the name of the package that should be installed.
 // It will return the path to the package that should be installed for the given flavor and agent flavor.
 // If the package is not found, it will return an error.
-// If multiple packages are found, it will return the first one and print a warning.
+// If multiple packages are found, it will fail.
 // Args:
 //   - localPath: either a path to a folder or a path to a file
 //     if a folder is provided, it should have a structure similar to what the agent CI exposes.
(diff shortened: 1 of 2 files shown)


components/datadog/agent/package.go
```diff
@@ -93,6 +94,14 @@ func GetPackagePath(localPath string, flavor tifos.Flavor, agentFlavor string, a
 				}
 			}
 
+			// If we're on Windows, we need to check if the pipeline ID is in the package name, this is a workaround to avoid an issue with the Windows Runners
+			// where other pipelines packages can pollute the artifacts
+			if flavor == tifos.WindowsServer && pipelineID != "" {
+				if !strings.Contains(entry.Name(), pipelineID) {
```
KSerrania, line 100: Does this work for actual Agent releases and release candidates? We don't put the pipeline id in the name of the package in this case.

components/datadog/agent/package.go
```diff
@@ -93,6 +94,14 @@ func GetPackagePath(localPath string, flavor tifos.Flavor, agentFlavor string, a
 				}
 			}
 
+			// If we're on Windows, we need to check if the pipeline ID is in the package name, this is a workaround to avoid an issue with the Windows Runners
+			// where other pipelines packages can pollute the artifacts
+			if flavor == tifos.WindowsServer && pipelineID != "" {
```
clarkb7, line 99: not all artifacts have the pipeline in their name, see https://github.com/DataDog/datadog-agent/pull/23981

components/datadog/agent/package.go
KevinFairise2: Yep you're right. I'll make the condition a bit more flexible by keeping the artifacts that does not contain `pipeline.`

chouetz (approved):

PR #1543, opened 2025-05-16
Use io_uring to handle I/O to VM backing store

usamasaqib:
What does this PR do?
---------------------

This PR changes the way qemu writes I/O from within the VM to the backing store. The new mechanism instructs QEMU to use `io_uring` instead of creating multiple processes to handle incoming requests. This approach reduces the number of threads launched to handle VM I/O.

More details can be found in this PR https://github.com/DataDog/single-machine-performance/pull/2947

-------------------

Motivation
----------

Additional Notes
----------------


gjulianm:
What's the downside (if any) of enabling this? (or, put another way, why is it not the default for QEMU?)

usamasaqib:
> What's the downside (if any) of enabling this? (or, put another way, why is it not the default for QEMU?)

I am not aware of any downsides. 

> why is it not the default for QEMU?

probably because io_uring is not available on older kernel versions.

usamasaqib:
This change has certainly helped reduce the load average across all the instances launched as part of a single pipeline
https://app.datadoghq.com/dashboard/zs9-uia-gsg?fromUser=false&refresh_mode=sliding&tpl_var_ci.pipeline.id%5B0%5D=65450327&tpl_var_git-branch%5B0%5D=usama.saqib%2Fio-uring-opt&from_ts=1747646212243&to_ts=1747649812243&live=true

gjulianm (approved):

PR #1562, opened 2025-05-23
ECS Fargate Readonly filesystem [CONTP-785]

gabedos:
What does this PR do?
---------------------

Creates an ECS Fargate nginx service that is monitored by the Agent. Gives the Agent ReadOnlyFilesystem permissions. Implements the existing solution for readOnly file systems which involve having an init container copy the files the Agent needs to write into a volume which it maintains the write permissions for.

-------------------

AWS ECS

Motivation
----------

Cover tests on Agent behavior in ECS Fargate with restricted permissions on the Agent container to write.

Additional Notes
----------------

QA: Deploy the aws/ecs scenario via `pulumi up -c scenario=aws/ecs -c ddinfra:env=aws/agent-sandbox -c ddinfra:aws/defaultKeyPairName=xxxxxxxx -c ddagent:apiKey=$DD_API_KEY -s gabedos-ecs-test`

image


components/datadog/agent/ecsFargate.go
```diff
@@ -9,15 +9,64 @@ import (
...
+	var readonlyFSMountPoints ecs.TaskDefinitionMountPointArray
+	var readonlyFSDependsOn ecs.TaskDefinitionContainerDependencyArray
+	if readonlyRootFS {
+		initContainer = &ecs.TaskDefinitionContainerDefinitionArgs{
+			Cpu:                    pulumi.IntPtr(0),
+			Name:                   pulumi.String("init-copy-agent-config"),
+			Image:                  pulumi.String(image),
+			Essential:              pulumi.BoolPtr(true),
+			ReadonlyRootFilesystem: pulumi.BoolPtr(false),
+			Command:                pulumi.StringArray{pulumi.String("sh"), pulumi.String("-c"), pulumi.String("cp -R /etc/datadog-agent/* /agent-config/")},
```
clamoriniere, line 27: Do we have a precise list of the files the Agent needs to modify?

I know we’re using a similar workaround in other environments, but I wonder if it might be worth discussing with the agent-configuration team to support this logic natively.

Here’s the flow I envision:
	•	The Agent detects that it’s running in a read-only environment (we could pass a configuration option to indicate which folder is writable).
	•	In that case, the Agent treats /etc/datadog-agent/ as read-only.
	•	Any file that it needs to modify in that directory would be automatically copied to the writable location.

The goal here is twofold:
	1.	Have a clearer understanding and traceability of which files the Agent is creating or modifying on the filesystem.
	2.	Avoid recurring issues related to read-only filesystems when the Agent attempts to write new files.

This complexity could be encapsulated in a component used by any feature that needs to write to the filesystem, making the behavior more robust and transparent.

**To be clear, I’m not asking you to address this before merging this pull request.**
However, it would be valuable to start a discussion with the team responsible for Agent configuration to raise the issue and explore what solutions they might propose.

components/datadog/apps/nginx/ecsFargateReadonly.go
```diff
@@ -0,0 +1,112 @@
...
+	EcsFargateComponent := &ecsComp.Workload{}
+	if err := e.Ctx().RegisterComponentResource("dd:apps", namer.ResourceName("grp"), EcsFargateComponent, opts...); err != nil {
+		return nil, err
+	}
+
+	opts = append(opts, pulumi.Parent(EcsFargateComponent))
+
+	serverContainer := &ecs.TaskDefinitionContainerDefinitionArgs{
+		Name:  pulumi.String("nginx"),
+		Image: pulumi.String("ghcr.io/datadog/apps-nginx-server:main"),
```
L3n41c, line 29: In order to get better test reproducibility, the use of floating `main` tag is now deprecated and should be replaced by “fixed” tags:

```suggestion
		Image: pulumi.String("ghcr.io/datadog/apps-nginx-server:" + apps.Version),
```

components/datadog/apps/nginx/ecsFargateReadonly.go
```diff
@@ -0,0 +1,112 @@
...
+				pulumi.String("CMD-SHELL"),
+				pulumi.String("apk add curl && curl --fail http://localhost || exit 1"),
+			},
+		},
+		LogConfiguration: ecsClient.GetFirelensLogConfiguration(pulumi.String("nginx"), pulumi.String("nginx"), apiKeySSMParamName),
+	}
+
+	queryContainer := &ecs.TaskDefinitionContainerDefinitionArgs{
+		Name:  pulumi.String("query"),
+		Image: pulumi.String("ghcr.io/datadog/apps-http-client:main"),
```
L3n41c, line 72: In order to get better test reproducibility, the use of floating `main` tag is now deprecated and should be replaced by “fixed” tags:

```suggestion
		Image: pulumi.String("ghcr.io/datadog/apps-http-client:" + apps.Version),
```

components/datadog/apps/nginx/ecsFargateReadonly.go
```diff
@@ -0,0 +1,112 @@
...
+		Command: pulumi.StringArray{
+			pulumi.String("-url"),
+			pulumi.String("http://localhost"),
+		},
+		Cpu:       pulumi.IntPtr(50),
+		Memory:    pulumi.IntPtr(32),
+		Essential: pulumi.BoolPtr(true),
+	}
+
+	serverTaskDef, err := ecsClient.FargateTaskDefinitionWithAgent(e, "nginx-fg-readonly", pulumi.String("nginx-fg-readonly"), 1024, 2048, false,
```
L3n41c, line 82: Is this a typo?

```suggestion
	serverTaskDef, err := ecsClient.FargateTaskDefinitionWithAgent(e, "nginx-fg-readonly", pulumi.String("nginx-fg-readonly"), 1024, 2048, true,
```

components/datadog/apps/nginx/ecsFargateReadonly.go
L3n41c, line 1: This file is a copy of `ecsFargate.go`.
Would it be possible to keep only one of the two `ecsFargate.go` and `ecsFargateReadonly.go` files instead of duplicating their content?

L3n41c (commented):
This PR adds a parameter to have the datadog agent container started with read only root filesystem or with the default read write root filesystem.

Would it make sense to simplify the code by keeping *only* the read only root filesystem setup?

Would it reduce the code coverage?
If we validate that the agent only write to a subset of directories (corresponding to the docker volumes) with the readonly root filesystem setup, having another setup where the agent *could* write to anywhere but *won’t* doesn’t bring any extra value.

resources/aws/ecs/fargateService.go
```diff
@@ -74,15 +74,38 @@ func FargateTaskDefinitionWithAgent(
...
 	cpu, memory int,
+	readonlyRootFS bool,
 	containers map[string]ecs.TaskDefinitionContainerDefinitionArgs,
 	apiKeySSMParamName pulumi.StringInput,
 	fakeintake *fakeintake.Fakeintake,
 	image string,
 	opts ...pulumi.ResourceOption,
 ) (*ecs.FargateTaskDefinition, error) {
-	containers["datadog-agent"] = *agent.ECSFargateLinuxContainerDefinition(&e, image, apiKeySSMParamName, fakeintake, GetFirelensLogConfiguration(pulumi.String("datadog-agent"), pulumi.String("datadog-agent"), apiKeySSMParamName))
+	initContainer, agentContainer := agent.ECSFargateLinuxContainerDefinition(&e, "public.ecr.aws/datadog/agent:latest", apiKeySSMParamName, fakeintake, GetFirelensLogConfiguration(pulumi.String("datadog-agent"), pulumi.String("datadog-agent"), apiKeySSMParamName), readonlyRootFS)
```
KevinFairise2, line 84:
```suggestion
	initContainer, agentContainer := agent.ECSFargateLinuxContainerDefinition(&e, image, apiKeySSMParamName, fakeintake, GetFirelensLogConfiguration(pulumi.String("datadog-agent"), pulumi.String("datadog-agent"), apiKeySSMParamName), readonlyRootFS)
```


**🔨 warning:** ‏
FargateTaskDefinitionWithAgent is sometimes used with the version of the agent built in the pipeline, to actually test the agent. If you change that all the tests will start to use the latest agent version

components/datadog/apps/nginx/ecsFargateReadonly.go
gabedos: Ahh thank you for the context! I branched off before that change was in place

PR #1624, opened 2025-07-16
Bump orchestrion to 1.4.0

CelianR:
What does this PR do?
---------------------

This fixes issues we have for the `build` github action.
There is a nil pointer exception which is fixed with orchestrion `v1.4.0`.

-------------------

Motivation
----------

Additional Notes
----------------

Cf this [thread](https://dd.slack.com/archives/C02DW5ZH1EF/p1752666719370799).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install orchestrion
         run: |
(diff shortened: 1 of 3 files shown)


davidor (approved):

Ishirui (approved):
Thanks for looking into this !

PR #1629, opened 2025-07-18
added Windows server 2025 to the support list

guohdd:
What does this PR do?
---------------------
This PR addresses the need of Windows 2025 server support described in [WINA-1309](https://datadoghq.atlassian.net/browse/WINA-1309)

- Fixed issues caused by path containing spaces so that Windows server could be deployed from Windows PowerShell.
- Added specific firewall rule for Windows Server 2025.
- Branden figured out a solution to the issue that root drive on Windows Server 2025 is missing the SE_DACL_AUTO_INHERITED flag (WINA-1694). Microsoft support case ID: 2508040010002067.
- Updated AD function level from 6 to 7 to support Windows Server 2025. Level 7 also supports Windows Server 2012, 2016, 2019 and 2022.

-------------------
The Datadog Windows agent will go through testing on Windows Server 2025.

Motivation
----------
To ensure Datadog agent supports Windows Server 2025.

Additional Notes
----------------
Validation has been done by running full pipelines:
- https://gitlab.ddbuild.io/DataDog/datadog-agent/-/pipelines/72798715

[WINA-1309]: https://datadoghq.atlassian.net/browse/WINA-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/os/scripts/setup-ssh.ps1
```diff
@@ -1,10 +1,69 @@
...
+# own rule that's more permissive for the testing environment.
+function Set-SshFirewallConfiguration {
+  # return if the OS is not Windows server 2025
+  if (-not (Is-WindowsServer2025)) {
+    return
+  } else {
+    # first add the following command: icacls.exe C:\ /inheritance:d
+    # this disables inheritance on the C:\ drive
+    icacls.exe C:\ /inheritance:d
+  }
```
clarkb7, line 37: lets move this out of the firewall config function.

and explain why we're doing this in the comment, and link our card and MS support case id
```
# TODO(WINA-XXX): The root drive on Windows Server 2025 is missing the SE_DACL_AUTO_INHERITED flag.
# This causes new files/directories under the root drive to have incorrect permissions,
# which causes failures in some of our E2E tests.
# Microsoft support case ID: XXX
# https://learn.microsoft.com/en-us/windows-hardware/drivers/ifs/security-descriptor-control
# The following command restores the flag without affecting the DACL
```

guohdd: changes have been made per code review

PR #1700, opened 2025-09-17
Frank.spano/contint 4708

frank-spano:
What does this PR do?
---------------------
Pulls latest release version of kubernetes and kind to use in test suite when "latest" is passed as kubernetes version.

-------------------
Right now only TestKindSuite, invoked by the datadog-agent repo

Motivation
----------
Outage detailed in https://datadoghq.atlassian.net/browse/CONTINT-4708

Additional Notes
----------------

### DIFF
--- a/components/kubernetes/kind.go
+++ b/components/kubernetes/kind.go
@@ -70,11 +70,9 @@ func NewKindClusterWithConfig(env config.Env, vm *remote.Host, name string, kube
 		if kindVersionConfig.UsePublicRegistry {
 			// Use public Docker Hub for latest/dynamic versions not available in internal mirror
 			nodeImage = fmt.Sprintf("docker.io/%s:%s", kindNodeImageName, kindVersionConfig.NodeImageVersion)
-			env.Ctx().Log.Info(fmt.Sprintf("CONTINT-4708: Using PUBLIC registry for Kind node image: %s (KindVersion: %s, KubeVersion: %s)", nodeImage, kindVersionConfig.KindVersion, kindVersionConfig.KubeVersion), nil)
 		} else {
 			// Use internal mirror for cached/static versions
 			nodeImage = fmt.Sprintf("%s/%s:%s", env.InternalDockerhubMirror(), kindNodeImageName, kindVersionConfig.NodeImageVersion)
-			env.Ctx().Log.Info(fmt.Sprintf("CONTINT-4708: Using INTERNAL mirror for Kind node image: %s (KindVersion: %s, KubeVersion: %s)", nodeImage, kindVersionConfig.KindVersion, kindVersionConfig.KubeVersion), nil)
 		}
 		createCluster, err := runner.Command(
 			commonEnvironment.CommonNamer().ResourceName("kind-create-cluster"),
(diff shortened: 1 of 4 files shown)


components/kubernetes/kind_versions.go
```diff
@@ -86,8 +107,156 @@ var kubeToKindVersion = map[string]KindConfig{
...
+	Draft      bool   `json:"draft"`
+	PreRelease bool   `json:"prerelease"`
+}
+
+// getLatestKindVersionDynamic fetches the latest Kind version from GitHub releases API
+func getLatestKindVersionDynamic() (string, error) {
+	client := &http.Client{Timeout: 30 * time.Second}
+
+	// Fetch releases from GitHub API
+	githubURL := "https://api.github.com/repos/kubernetes-sigs/kind/releases"
```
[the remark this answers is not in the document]
frank-spano: hmm it does redirect to the latest version, which I should think probably be a proper release version? On the other hand, I don't think iterating through the list of releases is that big of a deal. Let me think about it, maybe we can the link you provided, and if it doesn't resolve to a proper release version we can call the api for all releases.

lavigne958 (approved):

components/kubernetes/kind.go
```diff
@@ -66,7 +66,14 @@ func NewKindClusterWithConfig(env config.Env, vm *remote.Host, name string, kube
 			return err
 		}
 
-		nodeImage := fmt.Sprintf("%s/%s:%s", env.InternalDockerhubMirror(), kindNodeImageName, kindVersionConfig.NodeImageVersion)
+		var nodeImage string
+		if kindVersionConfig.UsePublicRegistry {
```
chouetz, line 70: Do you confirm this is always `false` when undefined (which is the case for all versions defined in the `KindConfig` map? I would like to prevent an external dependency if it's not necessary

components/kubernetes/kind_versions.go
```diff
@@ -86,8 +107,156 @@ var kubeToKindVersion = map[string]KindConfig{
...
+		KubeVersion:       latestVersion.String(), // Clean version for semantic parsing
+		UsePublicRegistry: true,                   // Latest versions must be pulled from Docker Hub
+	}, nil
+}
+
 // GetKindVersionConfig returns the kind version and the kind node image to use based on kubernetes version
 func GetKindVersionConfig(kubeVersion string) (*KindConfig, error) {
+	// Handle "latest" as a special case
+	if kubeVersion == "latest" {
+		return getLatestKindVersionConfig()
```
chouetz, line 257: Should we catch the error here, and just raise a warning in case of failure in the child methods?

components/kubernetes/kind_versions.go
```diff
@@ -86,8 +107,156 @@ var kubeToKindVersion = map[string]KindConfig{
 	},
 }
 
+// getKindVersionForKubernetes determines the appropriate Kind version for a given Kubernetes version
+// Based on Kind release compatibility: https://github.com/kubernetes-sigs/kind/releases
+// Used as fallback if dynamic resolution fails
+func getKindVersionForKubernetes(kubeVersion *semver.Version) string {
```
KevinFairise2, line 113: Why not using the map just above that was exactly made to do that AFAIU?

PR #1702, opened 2025-09-18
Add launch template to enforce hop limit for AL2023

justin-lesko:
What does this PR do?
---------------------
Overrides the hop limit to 2 on AL2023 nodes by using a launch template.

-------------------
aws/eks

Motivation
----------
Resolve an Agent error due to hop limit being too low.

Additional Notes
----------------

### DIFF
--- a/resources/aws/eks/nodeGroups.go
+++ b/resources/aws/eks/nodeGroups.go
@@ -26,7 +26,7 @@ const (
 
 func NewAL2023LinuxNodeGroup(e aws.Environment, cluster *eks.Cluster, nodeRole *awsIam.Role, opts ...pulumi.ResourceOption) (*eks.ManagedNodeGroup, error) {
 	name := "linux"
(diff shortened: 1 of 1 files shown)


lavigne958 (approved):

L3n41c (commented):
I know that `LaunchTemplate` is strictly needed only when we have to explicitly set the `HttpPutResponseHopLimit` option to `2`, i.e.: only on Amazon Linux 2023 nodes.
But I can’t see any drawback in simplifying the code by having the same homogeneous setup for all node types.

If we move all the code of `newLinuxLaunchTemplate` directly inside `newManagedNodeGroup`,
* `NewAL2023LinuxNodeGroup` and `NewAL2023LinuxARMNodeGroup` can be one-line like they were before,
* `newManagedNodeGroup` don’t need an extra parameter.
* `newManagedNodeGroup` don’t need the `if launchTemplate != nil {` block anymore.

The code should be simpler if we use Launch templates for all node groups.

PR #1709, opened 2025-09-23
CONTINT-4792 - stress-ng app use tmp directory

lavigne958:
What does this PR do?
---------------------

Fixes the deployment of TraceGen, cpustress and dogstatsd on openshift.

## TLDR

- in case of openshift scenario set variable `useNestedVirt` to enable nested on gcp
- use a temporary dir when necessary (read-only root fs on openshift)
- use a serviceAccount privileged access to allow to mount a host socket 

## Tracegen app fixes

the stress-ng app requires it's current working directory to be writable
and a given temporary path to be writable.

currently we provide no temporary path so stress-ng uses its current
working direcotyr path.
see here when option `temp-path` is *not* given:
https://github.com/ColinIanKing/stress-ng/blob/85e2ebc2bb3c1aa2b8db767c17fa4ed829a3f1c6/core-filesystem.c#L356-L358

To solve that:

1. request a temporary directory in the pod spec under the path `/tmp`
2. pass that temporary directory to stress-ng
3. change the working directory too for io-stress to work too.

We need to way to identify that we run an openshit scenario and enable
the nested virtualization.

Use a instance type that allows nested virtualization. google cloud docs
entions that all `e2` instances can't run nested virt.

to conclude, use the regular stress-ng deployment and remove the
openshift dedicated one.

## Tracegen and dogstats fixes

Refactor the function to deploy tracegen application.

On openshift, we must create a new service account that allows us to
have some priviledged access to the host for the app tracegen-uds.

In openshift scenario: create the serviceAccount, create the
clusterRoleBinding to bind that service account to the privileged
SecurityContextConstraint from openshift.

Same goes for Dogstatsd

-------------------

`gcp/openshift`

Motivation
----------

Use the same test/workload app for all deployments

Additional Notes
----------------

### DIFF
--- a/components/datadog/apps/tracegen/k8s.go
+++ b/components/datadog/apps/tracegen/k8s.go
@@ -17,12 +17,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
-func deployDeployment(e config.Env, name string, deploymentSpec *appsv1.DeploymentArgs, opts ...pulumi.ResourceOption) error {
-	_, err := appsv1.NewDeployment(e.Ctx(), name, deploymentSpec, opts...)
-	return err
-}
-
-func generateTracegenTcpSpec(namespace string) *appsv1.DeploymentArgs {
+func generateTracegenTCPSpec(namespace string) *appsv1.DeploymentArgs {
 	return &appsv1.DeploymentArgs{
 		Metadata: &metav1.ObjectMetaArgs{
 			Name:      pulumi.String("tracegen-tcp"),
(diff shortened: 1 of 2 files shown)


bitomaxsp (approved):

CelianR (approved):

lavigne958 (commented):
currently trying the changes, will open a PR once ready.
Branch will be: `crayon/improve_openshift_test_workload_deploy`

lavigne958:
> currently trying the changes, will open a PR once ready. Branch will be: `crayon/improve_openshift_test_workload_deploy`

never mind, this will be a duplicate of: #1727

PR #1712, opened 2025-09-25
Improve cluster name in gcp/gke scenario

lavigne958:
What does this PR do?
---------------------

Improve the cluster name in gcp/gke scenario so we can find the cluster name matching the stack name on GCP console.

Before:
- local object name: gcp-gke
- cluster name on console: gcp-gke-abcd0123

now:
- local object name: gcp-crayon-gke-cp-dashboard
- cluster name on console: gcp-crayon-gke-cp-dashboard-abcd01234

-------------------

`gcp/gke` only

Motivation
----------

Finding my own cluster online is hard, with a few clusters I need to open all of them 1 by 1 to try to identify my cluster.
When deployed using `-c ddinfra:autopilot=true` it's impossible as the cluster hardware is auto managed by gcp. there are no traces of your name anywhere.

With this changes we have the stack trace in the name which helps identify the owner most of the time.

Additional Notes
----------------


resources/gcp/gke/cluster.go
```diff
@@ -10,7 +12,22 @@ import (
 func NewCluster(e gcp.Environment, name string, autopilot bool, opts ...pulumi.ResourceOption) (*container.Cluster, pulumi.StringOutput, error) {
 	opts = append(opts, e.WithProviders(config.ProviderGCP))
 
+	clusterName := e.Namer.DisplayName(40)
```
triviajon, line 15: Why is this 40? Can it be a constant with an explanation?

triviajon (changes requested):

resources/gcp/gke/cluster.go
lavigne958: of course good catch !
it's the max length allowed by gcp for a cluster name.

PR #1879, opened 2026-09-21
Add TCP congestion test tools to apps-npm-tools image

atkinsam:
## What does this PR do?

Adds the tools needed by the TCP congestion e2e suite in datadog-agent
(`test/new-e2e/tests/npm/ec2_1host_tcp_congestion_test.go`) to the
`apps-npm-tools` image: `iperf3`, `iproute2` (tc), `netcat-openbsd`,
`procps` (pkill/pgrep), and `python3`.

## Motivation

The TCP congestion suite currently runs on stock `ubuntu:22.04` containers
with runtime `apt-get install`, flagged as a supply-chain/availability
problem in its PR review (DockerHub rate limits, mutable Ubuntu mirrors):
https://github.com/DataDog/datadog-agent/pull/56683 (P2 comment).

The established pattern for npm e2e tools is the prebuilt, pinned
`apps-npm-tools` image, so this extends it instead of adding a new image.
Purely additive for existing consumers (curl, ab, dig unchanged).

## How I validated

Built the image locally and verified the suite's full in-container
toolchain on the Alpine/busybox userland: `pkill -9 -x`, `nc -z`, `tc qdisc
netem`, `iperf3`, `python3` sockets with TCP_WINDOW_CLAMP, `dd`, `bash`.
End-to-end smoke test: two containers on a bridge network with
`tc netem delay 50ms loss 5%` on the client — iperf3 transfer degraded as
expected (~2.6 Mbit/s under 5% loss), qdisc cleanup OK.

Note: locally validated on arm64 only; CI builds both amd64 and arm64.

Follow-up in datadog-agent once this is released: bump `apps.Version` and
switch the suite's compose file to
`ghcr.io/datadog/apps-npm-tools:{APPS_VERSION}`.


jmw51798 (approved):

apps/version.go
```diff
@@ -1,3 +1,3 @@
 package apps
 
-const Version = "v0.0.5"
+const Version = "v0.0.8"
```
atkinsam: It seems that this version const isn't actually used in the e2e tests. The current image version tag used in the e2e tests separate: https://github.com/DataDog/datadog-agent/blob/main/test/e2e-framework/components/datadog/apps/version.go. This constant is used in some helper code in this repo: https://github.com/DataDog/test-infra-definitions/blob/33a0d15ac90c905ea6a0577915e5595553c23bc3/apps/npm-tools/k8s.go#L64

But that code doesn't appear to be used anywhere. 

I bumped this version to the version I would plan on tagging the merge commit with (v0.0.8) just for posterity but I would be happy to undo this change if anyone feels strongly.

